PR #26, opened 2018-08-02
Added Dockerfile

Thann:
Arch-based image for hsd ~560MB

I can add a docker-compose file too if you want =]


Thann:
@chjj is unbound an run-time dependency, or just compile-time?

PR #34, opened 2018-08-05
Added Dockerfile

hectorm:
Docker image based on node:10-alpine, ~110MB (~40MB gzipped)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+/node_modules/
+/.git/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+FROM node:10-alpine AS build
+
+# Install build dependencies
+RUN apk add --no-cache \
+		bash \
+		g++ \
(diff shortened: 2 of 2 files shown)


Thann:
leveldown has had issues in the past running on alpine because of libmusl, I'll try and see if I can repro those.

bmillemathias:
Thanks for the work. That would be great to have regular update of the docker image also.

Thann:
Hey! we just merged in my Dockerfile PR, but I want to discuss the additions you have, so we can add them if appropriate.

The only substantive difference I see is that you use the `hsd` user. I don't think this is necessary, because there shouldn't be anything secret inside the container that you would want to insulate from the main process by changing the user. (and that process has access to nothing outside the container)

If I'm missing something , please ping me and we can discuss it further =]

tynes:
@hectorm Thank you for your contribution, but I am going to close this now that there is already a `Dockerfile` in the repo

PR #37, opened 2018-08-06
Add a Dockerfile

jmunson:
I noticed there are several other Dockerfile pull requests, but they seem to be missing some things.

Most noticably, this one includes EXPOSE so users know which ports we use by default, VOLUME to make it easier to persist the blockchain, and HEALTHCHECK to leverage the existing CLI tool you wrote to ensure hsd is running and responding.

Additionally, I replaced the bash shebang on the cmd/ scripts with sh, since there were no bashisms used and sh works fine. Minimal distros often ship without bash, so this helps run there.  If you plan to expand these scripts to use bash specific features I could modify the Dockerfile to include bash instead, but they seem fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM node:alpine AS build
+RUN apk add --no-cache g++ gcc bash make unbound-dev python2
+RUN mkdir /hsd
+WORKDIR /hsd
+COPY package.json .
+ARG NODE_ENV=production
+RUN npm install
+
+
+
+FROM node:alpine
+WORKDIR /home/hsd
+
+#Set up non-root user to run as
+RUN adduser -S hsd
+USER hsd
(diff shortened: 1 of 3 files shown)


Thann:
Hey! we just merged in my Dockerfile PR, but I want to discuss the additions you have, so we can add them if appropriate.

One substantive difference I see is that you use the hsd user. I don't think this is necessary, because there shouldn't be anything secret inside the container that you would want to insulate from the main process by changing the user. (and that process has access to nothing outside the container)

You also included a healthcheck which actually requires the API key to work, so I figure that's something better to configure with docker-compose. 

If I'm missing something, please ping me and we can discuss it further =]

tynes:
@jmunson Thank you for your contribution, but I'm going to close this as there is already a `Dockerfile` in the repo

PR #38, opened 2018-08-07
node/rpc: minor syntax error fix

rojii:
I've dug around a bit and came to a realization that the reason for this Syntax Error is simply due to the fact that the for await iterable isn't considered valid syntax in versions of nodejs under 9.2.

Disregard this PR.

The ideal thing to do is bump the nodejs package versions to 9.2+

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2284,7 +2284,7 @@ class RPC extends RPCBase {
     const txn = this.chain.db.txn;
     const items = [];
 
-    for await (const [key, value] of txn) {
+    for (const [key, value] of txn) {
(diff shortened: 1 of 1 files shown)


rojii:
I dug around a bit it's pretty clear that the syntax error isn't an error. It's just simply required to use nodejs versions 9.2+ to be able compile correctly. 

Disregard this pr, I'll simply bump the nodejs version in the package.json.

rojii:
Will resubmit the pr with a minor change to the package.json file, which indicates the version of node necessary for running the daemon without the Syntax issue in the RPC file found in the Node library.

PR #44, opened 2018-08-08
lib/node: use an address object in the coin address endpoint [bugfix]

turbomaze:
The coin address endpoint at `./lib/node/http.js` passes an address string to `this.node.getCoinsByAddress(*)`, which eventually passes that same address to the mempool's `getCoinsByAddress(*)`. Attempting to query an address from the CLI or from CURL via `hsd-cli coin $ADDRESS` results in the error message `object is not an address`.

This PR creates an `Address` object after receiving the address string to fix this issue.


boymanjor:
Sorry it took a while to get to this. LGTM.

PR #45, opened 2018-08-08
FreeBSD portability for bash

chrisbgit:
modified:   bin/hsd
	modified:   bin/hsw


tynes:
Thank you for the PR, this file has actually been updated and rewritten in node.js here:
https://github.com/handshake-org/hsd/blob/7ed29b82017dedf639926c969e113de72ecaa29f/bin/hsd#L1

PR #46, opened 2018-08-08
fix: address type in http api /coin/address and /tx/address

LanfordCai:
For /coin/address and /tx/address, they receive address strings, but doesn't convert those address strings to address object, so when we request these apis, it will return errors with message 'Object is not address'. This PR is try to fix this.

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -18,6 +18,7 @@ const ccmp = require('bcrypto/lib/ccmp');
 const util = require('../utils/util');
 const TX = require('../primitives/tx');
 const Claim = require('../primitives/claim');
+const Address = require('../primitives/address');
 const Network = require('../protocol/network');
 const pkg = require('../pkg');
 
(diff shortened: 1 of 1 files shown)


lib/node/http.js
```diff
@@ -187,12 +189,16 @@ class HTTP extends Server {
     // Bulk read UTXOs
     this.post('/coin/address', async (req, res) => {
       const valid = Validator.fromRequest(req);
-      const address = valid.array('addresses');
+      const addressStrings = valid.array('addresses');
 
-      enforce(address, 'Address is required.');
+      enforce(addressStrings, 'Addresses is required.');
+      addressStrings.forEach(addrString => {
```
boymanjor: Instead of using a `forEach` here, how about using a `for` loop and building the array of `Address` objects iteratively. Then we can get rid of the `map` call on line 200. `hsd` is built for speed and `forEach` and `map` are much slower than traditional `for` loops.

lib/node/http.js
```diff
@@ -243,12 +250,16 @@ class HTTP extends Server {
     // Bulk read TXs
     this.post('/tx/address', async (req, res) => {
       const valid = Validator.fromRequest(req);
-      const address = valid.array('addresses');
+      const addressStrings = valid.array('addresses');
 
-      enforce(address, 'Address is required.');
+      enforce(addressStrings, 'Addresses is required.');
+      addressStrings.forEach(addrString => {
```
boymanjor: Same as above.

boymanjor:
Hey, thanks for the PR. I've added some comments in line. Also, to keep the commit history clear, we tend to use the prefix of the commit message to give a hint as to where the change is happening, e.g. `node/http: [what you did]`.

LanfordCai:
Thank you for your suggestions, I have updated the code :) @boymanjor

boymanjor:
LGTM

PR #68, opened 2018-10-16
remove names and tld from npmignore

bucko13:
fix from @boymanjor 's bundle branch based off use with bpanel. Fixes error where bundling would fail when hsd is installed as a module because it can't find the ignored jsons.


boymanjor:
LGTM

PR #72, opened 2018-10-31
update toMilli exponent to 3

bucko13:
The exponents for `toMilli` and `fromMilli` didn't match. Following bcoin's `Amount.toMBTC` ([here](https://github.com/bcoin-org/bcoin/blob/85ed59c842f2aa4c1a8154d74a9cad7696cdfee3/lib/btc/amount.js#L90)), `toMilli` should encode the amount with exponent `5` instead of `4`.

### DIFF
--- a/lib/ui/amount.js
+++ b/lib/ui/amount.js
@@ -90,7 +90,7 @@ class Amount {
    */
 
   toMilli(num) {
-    return Amount.encode(this.value, 5, num);
+    return Amount.encode(this.value, 3, num);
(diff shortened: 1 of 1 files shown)


bucko13:
After discussion with @boymanjor and realizing that the main `Coins` exponent is less than in bcoin (6 instead of 8 from base), I updated to change both `toMilli` and `fromMilli` to be 3.

boymanjor:
LGTM

PR #73, opened 2018-11-02
WIP: introduce browser example

boymanjor:
The full node can now run in the browser using the in-memory filesystem.
Future work includes fixing stateful execution.

### DIFF
--- a/browser/src/app.js
+++ b/browser/src/app.js
@@ -1,0 +1,287 @@
+'use strict';
+
+const Logger = require('blgr');
+const FullNode = require('../../lib/node/fullnode');
+const Amount = require('../../lib/ui/amount');
+const plugin = require('../../lib/wallet/plugin');
+const util = require('../../lib/utils/util');
+const ProxySocket = require('./proxysocket');
+
+const body = document.getElementsByTagName('body')[0];
+const log = document.getElementById('log');
+const wdiv = document.getElementById('wallet');
+const tdiv = document.getElementById('tx');
+const floating = document.getElementById('floating');
+const send = document.getElementById('send');
+const newaddr = document.getElementById('newaddr');
+const chainState = document.getElementById('state');
+const rpc = document.getElementById('rpc');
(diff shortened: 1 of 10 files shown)


browser/src/app.js
```diff
@@ -0,0 +1,283 @@
+'use strict';
+
+const Logger = require('blgr');
+const SPVNode = require('../../lib/node/spvnode');
```
chikeichan: @boymanjor I am running into an error where `this.timer.unref` is undefined. It appears that `spvnode` uses `setInterval().unref` which is available in node env but not browser.

I have added a quick polyfill and now the SPVNode can run in browser and sync!

```

const _setInterval = setInterval;

function Timer(t) {
  this.t = t;
}

Timer.prototype.valueOf = function() {
  return this.t;
};

Timer.prototype.unref = function() {
  clearInterval(this.t);
};

window.setInterval = function(fn, timeout) {
  return new Timer(_setInterval(fn, timeout));
};
```

tynes:
Bump @boymanjor, this has gotten stale. Do you have an idea of where you would like to go with this?

boymanjor:
@tynes I think once we get some of the testing work out of the way I will get back to this. I want to rebase and update it to use `bpkg` instead of `webpack`. I initially thought we might include a browser backend for the urkel tree, but after talking to JJ it doesn't really make sense. The speedup comes from flat files stored on disk. Also, the more compelling use case for `hsd` in the browser is for resolving names, and that can be done through SPV.

tynes:
If https://github.com/handshake-org/hsd/pull/115 gets merged, then `bpkg` will be a `devDependency` and it'll be easier to move forward with what you are saying

pinheadmz:
I think we should close this for now, bcoin has kinda abandoned the browser stuff as well and we removed webpack and all that as well.

Anyone interested in running an hsd browser node can consult this external project: https://github.com/pinheadmz/mobilehsd

Or try the live example at https://easyhandshake.com/browser.html

PR #80, opened 2018-12-11
add missing bw

kilpatty:
Right now if an Addr record is added without a whitelisted name (handshake, bitcoin, ethereum) then we get an assertion error as bw is not passed into writeNameBW. This just adds it back in.


boymanjor:
LGTM

PR #84, opened 2018-12-24
wallet: do not rebroadcast coinbase transactions

braydonf:
See https://github.com/bcoin-org/bcoin/issues/661 for further details. I found this while testing reorganizations in bcoin.


tynes:
I experienced a large reorg on testnet and think that this bug could have caused me to be banned from the rest of the network (since its so small) because I couldn't add peers and was halted on a different chain

boymanjor:
LGTM

PR #86, opened 2019-01-03
rpc: accept account name in auction methods

wi-ski:
This PR introduces a small set of changes:
1. It expands the RPC interface for `'senbid'` to take an optional `'account'` parameter
1. It expands the RPC interface for `'sendOpen'` to take an optional `'account'` parameter
1. It expands the RPC interface for `'sendUpdate'` to take an optional `'account'` parameter

For all of these it looks like account can be either an integer (for account indices) or an account's name.

Would love any feedback y'all might have!

### DIFF
--- a/lib/primitives/address.js
+++ b/lib/primitives/address.js
@@ -4,15 +4,15 @@
  * https://github.com/handshake-org/hsd
  */
 
-"use strict";
+'use strict';
 
-const assert = require("bsert");
-const bio = require("bufio");
-const { bech32 } = require("bstring");
-const blake2b = require("bcrypto/lib/blake2b");
-const sha3 = require("bcrypto/lib/sha3");
-const Network = require("../protocol/network");
-const consensus = require("../protocol/consensus");
+const assert = require('bsert');
+const bio = require('bufio');
+const {bech32} = require('bstring');
+const blake2b = require('bcrypto/lib/blake2b');
+const sha3 = require('bcrypto/lib/sha3');
+const Network = require('../protocol/network');
+const consensus = require('../protocol/consensus');
 
 /*
  * Constants
@@ -41,7 +41,8 @@ class Address extends bio.Struct {
     this.version = 0;
     this.hash = ZERO_HASH160;
 
-    if (options) this.fromOptions(options, network);
+    if (options)
+      this.fromOptions(options, network);
   }
 
   /**
@@ -51,11 +52,12 @@ class Address extends bio.Struct {
    */
 
   fromOptions(options, network) {
-    if (typeof options === "string") return this.fromString(options, network);
+    if (typeof options === 'string')
+      return this.fromString(options, network);
 
     assert(options);
 
-    const { hash, version } = options;
+    const {hash, version} = options;
 
     return this.fromHash(hash, version);
   }
@@ -68,7 +70,8 @@ class Address extends bio.Struct {
 
   getSigops(witness) {
     if (this.version === 0) {
-      if (this.hash.length === 20) return 1;
+      if (this.hash.length === 20)
+        return 1;
 
       if (this.hash.length === 32 && witness.items.length > 0) {
         const redeem = witness.getRedeem();
@@ -94,12 +97,15 @@ class Address extends bio.Struct {
    */
 
   isNull() {
-    if (this.hash.length === 20) return this.hash.equals(ZERO_HASH160);
+    if (this.hash.length === 20)
+      return this.hash.equals(ZERO_HASH160);
 
-    if (this.hash.length === 32) return this.hash.equals(consensus.ZERO_HASH);
+    if (this.hash.length === 32)
+      return this.hash.equals(consensus.ZERO_HASH);
 
     for (let i = 0; i < this.hash.length; i++) {
-      if (this.hash[i] !== 0) return false;
+      if (this.hash[i] !== 0)
+        return false;
     }
 
     return true;
@@ -123,7 +129,8 @@ class Address extends bio.Struct {
   equals(addr) {
     assert(addr instanceof Address);
 
-    return this.version === addr.version && this.hash.equals(addr.hash);
+    return this.version === addr.version
+      && this.hash.equals(addr.hash);
   }
 
   /**
@@ -137,7 +144,8 @@ class Address extends bio.Struct {
 
     const cmp = this.version - addr.version;
 
-    if (cmp !== 0) return cmp;
+    if (cmp !== 0)
+      return cmp;
 
     return this.hash.compare(addr.hash);
   }
@@ -202,7 +210,7 @@ class Address extends bio.Struct {
    */
 
   fromScript(script) {
-    assert(script && typeof script.encode === "function");
+    assert(script && typeof script.encode === 'function');
     return this.fromHash(sha3.digest(script.encode()), 0);
   }
 
@@ -215,7 +223,7 @@ class Address extends bio.Struct {
    */
 
   fromString(data, network) {
-    assert(typeof data === "string");
+    assert(typeof data === 'string');
 
     const addr = bech32.decode(data);
 
@@ -259,19 +267,18 @@ class Address extends bio.Struct {
    */
 
   fromHash(hash, version) {
-    if (version == null) version = 0;
+    if (version == null)
+      version = 0;
 
     assert(Buffer.isBuffer(hash));
     assert((version & 0xff) === version);
 
-    assert(version >= 0 && version <= 31, "Bad program version.");
-    assert(hash.length >= 2 && hash.length <= 40, "Hash is the wrong size.");
+    assert(version >= 0 && version <= 31, 'Bad program version.');
+    assert(hash.length >= 2 && hash.length <= 40, 'Hash is the wrong size.');
 
     if (version === 0) {
-      assert(
-        hash.length === 20 || hash.length === 32,
-        "Witness program hash is the wrong size."
-      );
+      assert(hash.length === 20 || hash.length === 32,
+        'Witness program hash is the wrong size.');
     }
 
     this.hash = hash;
@@ -288,7 +295,7 @@ class Address extends bio.Struct {
    */
 
   fromPubkeyhash(hash) {
-    assert(hash && hash.length === 20, "P2WPKH must be 20 bytes.");
+    assert(hash && hash.length === 20, 'P2WPKH must be 20 bytes.');
     return this.fromHash(hash, 0);
   }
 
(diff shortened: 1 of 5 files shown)


lib/primitives/address.js
```diff
@@ -4,15 +4,15 @@
...
-const sha3 = require('bcrypto/lib/sha3');
-const Network = require('../protocol/network');
-const consensus = require('../protocol/consensus');
+const assert = require("bsert");
+const bio = require("bufio");
+const { bech32 } = require("bstring");
+const blake2b = require("bcrypto/lib/blake2b");
+const sha3 = require("bcrypto/lib/sha3");
+const Network = require("../protocol/network");
+const consensus = require("../protocol/consensus");
```
tynes: hello wi-ski, thank you for submitting a pull request. If you notice in all of the other files, they use `'` over `"` so it may be a good idea to revert them back to `'`

wi-ski: Oops, totally. Will fix

wi-ski:
Roger roger - before I do that, do we like this approach? Would it be preferable to take an account name?

wi-ski:
#bump Will someone talk to me?

wi-ski:
Im super down to split this into different PRs etc.  I'd like some guidance on:

a. What an ideal interface would be. (account name vs index)
b. Whether this method for implementation is even kosher
    b1. Is there a better way to handle scoping these methods to an account?

tynes:
Hey @wi-ski - sorry for slow responses. Its awesome that you want to contribute
Do you mind joining the bcoin slack and we can chat? http://bcoin.io/slack-signup.html

wi-ski:
Heck yea! No worries at all.  Doing it now.

wi-ski:
Hey hey, heads up - it doesn't look I have received an invite. Possible its borked?

wi-ski:
Made the PR foot print significantly smaller. :pray: Ty for tolerating me.

Thann:
I'm curious if there are any side-effects to swapping the account during these phases. Should definitely add for that =]

wi-ski:
Whadya guys think? eh?

wi-ski:
https://media0.giphy.com/media/gZGlQX3wWAV1u/giphy.gif?cid=3640f6095c50f71176432f56590c6979

terrapanda:
@wi-ski 

I suggest the implementation for this option be with 

`hsw-rpc selectwallet [name] --account=[account]`

such that a new interface doesnt need to be implemented for each command.

lib/wallet/rpc.js
```diff
@@ -1877,31 +1877,44 @@ class RPC extends RPCBase {
   }
 
   async sendOpen(args, help) {
-    if (help || args.length < 1 || args.length > 2)
-      throw new RPCError(errs.MISC_ERROR, 'sendopen "name" ( force )');
+    if (help || args.length < 1 || args.length > 3)
+      throw new RPCError(errs.MISC_ERROR, 'sendopen "name" ( force ) ( account )');
```
pinheadmz: Putting the args in this order means `force` is required if you also want to pass `account`. Frankly I have no idea what `force` is doing: https://github.com/handshake-org/hsd/issues/93 but you may want to swap those

lib/wallet/rpc.js
```diff
@@ -1877,31 +1877,44 @@ class RPC extends RPCBase {
...
 
     const wallet = this.wallet;
     const valid = new Validator(args);
     const name = valid.str(0);
     const force = valid.bool(1, false);
+    let account = valid.str(2, '');
+
+    if (account === '')
+      account = 'default';
```
pinheadmz: pretty sure you can combine these right? `valid.str(2, 'default')` ?

lib/wallet/rpc.js
```diff
@@ -1912,18 +1925,22 @@ class RPC extends RPCBase {
...
+    const tx = await wallet.sendBid(name, bid, value, { account });
 
     return tx.getJSON(this.network);
   }
 
   async sendReveal(args, help) {
-    if (help || args.length > 1)
-      throw new RPCError(errs.MISC_ERROR, 'sendreveal "name"');
+    if (help || args.length > 2)
+      throw new RPCError(errs.MISC_ERROR, 'sendreveal "name" ( account )');
```
pinheadmz: Looks like this method has a `sendRevealAll` fallback if `name` is missing. That will be hard to do since SOME name (param#0) will be required to make `account` param#1. I wonder if we should also add a `*` wildcard to trigger the `sendRevealAll` the same way `!name` would. Then you could still also pass an account.

wi-ski:
@pinheadmz - Thank you for the feedback! Seriously, good catch(es)! Ill make those updates after work.

lib/wallet/rpc.js
Thann: You always add new arguments to the end, and you can totally call a function without defining intermediate arguments

lib/wallet/rpc.js
Thann: The difference is when an empty string is passed. And this is exactly how other rpc functions are written

lib/wallet/rpc.js
pinheadmz: If you run this from CLI:
`hsw-rpc sendopen <name> <account>`
... won't it interpret the `<account>` argument as `<force>`?  I encountered similar issue with Braydon's wallet rpc updates. If you don't identify args specifically with `--account=`, etc, the CLI will interpret the args in order from left to right

lib/wallet/rpc.js
pinheadmz: not always? https://github.com/bcoin-org/bclient/blob/master/bin/bwallet-cli#L39

pinheadmz: but I see, its always this way in rpc.js

Thann: Yeah, I mean empty strings become `'default'` for the other rpc functions

lib/wallet/rpc.js
Thann: I guess we could make it a named param, but I was just thinking you could just pass `false` as the second arg

lib/wallet/rpc.js
pinheadmz: Frankly, based on somewhat shallow review, I think `force` is actually unused, not sure what it was ever for, or if it's still supposed to even be there...?

lib/wallet/rpc.js
pinheadmz: Frankly, based on somewhat shallow review, I think `force` is actually unused, not sure what it was ever for, or if it's still supposed to even be there...?

Thann: Idk what the intention is/was. Should probably ask jj

pinheadmz:
Sorry I might've misled you - removing the `force` argument should be its own PR, as a solution to https://github.com/handshake-org/hsd/issues/93 @chjj will probably want to evaluate that on its own

wi-ski:
You saved me just now, great timing.

lib/wallet/rpc.js
wi-ski: Is your thinking like..
# Current
`hsw-rpc sendreveal` # Triggers wallet to send reveal`
`hsw-rpc sendreveal namestring` # Triggers wallet to sendreveal using from default account
`hsw-rpc sendreveal namestring accountname` # Triggers wallet to sendreveal using addresses from accountname
# Proposed
`hsw-rpc sendreveal` * # Triggers wallet to sendrevealall using default account
`hsw-rpc sendreveal` # Missing wild-card/name/accountname throws error?
`hsw-rpc sendreveal * accountname` # Triggers wallet to sendrevealall using addresses from accountname?

lib/wallet/rpc.js
```diff
@@ -1960,13 +1979,20 @@ class RPC extends RPCBase {
   }
 
   async sendUpdate(args, help) {
-    if (help || args.length !== 2)
-      throw new RPCError(errs.MISC_ERROR, 'sendupdate "name" "data"');
+    if (help || args.length !== 3)
```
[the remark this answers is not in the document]
wi-ski: :heavy_check_mark: [Handled here](https://github.com/handshake-org/hsd/pull/86/commits/58e60ac3d43e20a810bb6763901b90db25d13de2)

lib/wallet/rpc.js
pinheadmz: Um more like, I think
### current:
```
# sends reveal for ALL names with wallet.sendRevealAll()
hsw-rpc sendreveal

# sends reveal for one specific name with wallet.sendReveal()
hsw-rpc sendreveal <name>
```

### proposed
```
# sends reveal for ALL names with wallet.sendRevealAll() from default account
hsw-rpc sendreveal

# sends reveal for ALL names with wallet.sendRevealAll() from specific account
hsw-rpc sendreveal * <account>

# sends reveal for one specific name with wallet.sendReveal() from default account
hsw-rpc sendreveal <name>

# sends reveal for one specific name with wallet.sendReveal() from specific account
hsw-rpc sendreveal <name> <account>
```

I'm suggesting using `*` to trigger `sendRevealAll()` but I don't l know if that fits with other rpc conventions... just a thought

wi-ski:
@pinheadmz - I [responded to your comment](https://github.com/handshake-org/hsd/pull/86#discussion_r) for the sendreveal changes - let know your thoughts!

lib/wallet/rpc.js
wi-ski: Im resolving this convo if thats ok :)

lib/wallet/rpc.js
pinheadmz: ...yeah... I dunno. Looks like account-specificity wasn't part of the design here and adding it in is either gunna be a major refactor or a little hacky. You could maybe forget the option of "revealing all" from a single account (for now?) Because param #0 (name) will always be required for you to also have a param #1 (account) you could just forget the whole `*` idea, and that would just mean if you specify an account, you ALSO have to specify a SINGLE name

wi-ski: That seems like its gonna be the plan :)

wi-ski:
This would introduce some interface changes. Would updating documentation be helpful? :) I can dig in and find where one would do that.

wi-ski:
Hey! Revisiting this - could I get a convo going with someone about next steps?

@pinheadmz - Should I get a separate PR going for that force option?

@tynes - I've submitted my email for that slack channel several times, what should I be doing brudah?

wi-ski:
https://media0.giphy.com/media/oadZJB3hwMFjy/giphy.gif

tynes:
@wi-ski - here is a new link to join slack: 
https://join.slack.com/t/bcoin-dev/shared_invite/enQtNTkzODgyMDEzNzQ4LWYwMDVkM2JkMjIwODkwYjY3YmY2MGY2MGM5MmRlMTRmOTI1NjdiYTRmNmM2NWQ5MjE5ZWFiN2Q2NzhkNDY2Nzg

Please be sure to check your spam for the invite

wi-ski:
Roger - looks like im in!

wi-ski:
It looks like these changes made it into master in some form or another, closing this.

PR #88, opened 2019-01-14
Fix type assertion in PGP.fromJSON

chikeichan:
`PGP.fromJSON` should expect a json object. Since the following line is looking for a `json.hash` value, the only other way for this to work with the current code is to change the prototype of `String`


chikeichan:
cc: @boymanjor

PR #94, opened 2019-01-26
wallet: allow duplicate open after auction expires

### LINKED ISSUE #53: Auction closed with no bids
After opening an auction but not bidding on it during the bidding period, the auction is now closed, and  seems unable to be opened again. Do we have to wait the 27 days until it expires to reopen it?

`> hsw-rpc getnameinfo alicepaysbob`
```
{
  "name": "alicepaysbob",
  "nameHash": "3d61e5b16e22c646df9c993a73eba63a9c9c5463feeced11d5915e7d944ea5e8",
  "state": "CLOSED",
  "height": 7192,
  "renewal": 7192,
  "owner": {
    "hash": "0000000000000000000000000000000000000000000000000000000000000000",
    "index": 4294967295
  },
  "value": 0,
  "highest": 0,
  "data": "",
  "transfer": 0,
  "revoked": 0,
  "claimed": false,
  "weak": false,
  "stats": {
    "renewalPeriodStart": 7192,
    "renewalPeriodEnd": 15832,
    "blocksUntilExpire": 7366,
    "daysUntilExpire": 25.58
  }
}
```
`> hsw-rpc getauctioninfo alicepaysbob`
```
{
  "name": "alicepaysbob",
  "nameHash": "3d61e5b16e22c646df9c993a73eba63a9c9c5463feeced11d5915e7d944ea5e8",
  "state": "CLOSED",
  "height": 7192,
  "renewal": 7192,
  "owner": {
    "hash": "0000000000000000000000000000000000000000000000000000000000000000",
    "index": 4294967295
  },
  "value": 0,
  "highest": 0,
  "data": "",
  "transfer": 0,
  "revoked": 0,
  "claimed": false,
  "weak": false,
  "stats": {
    "renewalPeriodStart": 7192,
    "renewalPeriodEnd": 15832,
    "blocksUntilExpire": 7366,
    "daysUntilExpire": 25.58
  },
  "bids": [],
  "reveals": []
}
```

`> hsw-rpc sendopen alicepaysbob`
`Already sent an open for: alicepaysbob.`

`> hsw-rpc sendbid alicepaysbob 0.00001 0.00001`
`Name has not reached the bidding phase yet.`

pinheadmz:
Closes #53 

This PR solves the following issue:

1.  hs-wallet send open `name`
2. generate enough blocks to expire auction without a single bid
3. re-attempt: hs-wallet send open `name`

This currently throws the error `Already sent an open for: name`, but is otherwise allowed by the protocol.

What I did was add one more check to `wallet.makeOpen()` so it won't throw if the name is expired.

Also adds a new test module `wallet-test-auction` which is a sort of integration test, because it needs to test an actual `WalletDB` (not `memwallet`) with a `chain` and `miner`.

### DIFF
--- a/lib/wallet/txdb.js
+++ b/lib/wallet/txdb.js
@@ -827,9 +827,19 @@ class TXDB {
 
       const nameHash = covenant.getHash(0);
       const key = layout.o.encode(nameHash);
+      const hash = await this.bucket.get(key);
 
-      if (await this.bucket.has(key))
+      // Allow a double open if previous auction period has expired
+      // this is not a complete check for name availability or status!
+      if (hash) {
+        const names = this.wdb.network.names;
+        const period = names.biddingPeriod + names.revealPeriod;
+        const oldTX = await this.getTX(hash);
+        if (oldTX.height !== -1 && oldTX.height + period < this.wdb.height)
+          return false;
+
         return true;
+      }
     }
 
     return false;
--- a/test/wallet-auction-test.js
+++ b/test/wallet-auction-test.js
@@ -1,0 +1,187 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+/* eslint no-return-assign: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const Chain = require('../lib/blockchain/chain');
+const {states} = require('../lib/covenants/namestate');
+const WorkerPool = require('../lib/workers/workerpool');
+const Miner = require('../lib/mining/miner');
+const WalletDB = require('../lib/wallet/walletdb');
+const Network = require('../lib/protocol/network');
+const rules = require('../lib/covenants/rules');
+
+const network = Network.get('regtest');
+const NAME1 = rules.grindName(5, 2, network);
+const {
+  treeInterval,
+  biddingPeriod,
+  revealPeriod
+
+} = network.names;
+
+const workers = new WorkerPool({
+  enabled: false
+});
+
+const chain = new Chain({
+  memory: true,
+  network,
+  workers
+});
+
+const miner = new Miner({
+  chain,
+  workers
+});
+
+const cpu = miner.cpu;
+
+const wdb = new WalletDB({
+  network: network,
+  workers: workers
+});
+
+describe('Wallet Auction', function() {
+  let winner, openAuctionMTX, openAuctionMTX2;
+
+  before(async () => {
+    // Open
+    await chain.open();
+    await miner.open();
+    await wdb.open();
+
+    // Set up wallet
+    winner = await wdb.create();
+    chain.on('connect', async (entry, block) => {
+      await wdb.addBlock(entry, block.txs);
+    });
+
+    // Generate blocks to roll out name and fund wallet
+    let winnerAddr = await winner.createReceive();
+    winnerAddr = winnerAddr.getAddress().toString(network);
+    for (let i = 0; i < 4; i++) {
+      const block = await cpu.mineBlock(null, winnerAddr);
+      await chain.add(block);
+    }
+  });
+
+  after(async () => {
+    await wdb.close();
+    await miner.close();
+    await chain.close();
+  });
+
+  it('should open auction', async () => {
+    openAuctionMTX = await winner.createOpen(NAME1, false);
+    await winner.sign(openAuctionMTX);
+    const tx = openAuctionMTX.toTX();
+    await wdb.addTX(tx);
+  });
+
+  it('should fail to create duplicate open', async () => {
+    let err;
+    try {
+      await winner.createOpen(NAME1, false);
+    } catch (e) {
+      err = e;
+    }
+
+    assert(err);
+    assert.strictEqual(err.message, `Already sent an open for: ${NAME1}.`);
+  });
+
+  it('should mine 1 block', async () => {
+    const job = await cpu.createJob();
+    job.addTX(openAuctionMTX.toTX(), openAuctionMTX.view);
+    job.refresh();
+
+    const block = await job.mineAsync();
+
+    assert(await chain.add(block));
+  });
+
+  it('should fail to re-open auction during OPEN phase', async () => {
+    let err;
+    try {
+      await winner.createOpen(NAME1, false);
+    } catch (e) {
+      err = e;
+    }
+
+    assert(err);
+    assert.strictEqual(err.message, 'Name is already opening.');
+  });
+
+  it('should mine enough blocks to enter BIDDING phase', async () => {
+    for (let i = 0; i < treeInterval; i++) {
+      const block = await cpu.mineBlock();
+      assert(block);
+      assert(await chain.add(block));
+    }
+  });
+
+  it('should fail to re-open auction during BIDDING phase', async () => {
+    let err;
+    try {
+      await winner.createOpen(NAME1, false);
+    } catch (e) {
+      err = e;
+    }
+
+    assert(err);
+    assert.strictEqual(err.message, 'Name is not available.');
+  });
+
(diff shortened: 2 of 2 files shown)


pinheadmz:
Pulled out my attempt at a bug fix here after expanding the test a whole bunch. Gonna play with this some more...

pinheadmz:
OK, new approach here is to modify `txdb.isDoubleOpen(tx)` to ignore double opens if enough blocks have passed. This means that `txdb.add(tx, block)` and the `wallet.makeOpen()` check will allow a double open after enough blocks have passed.

That amount of time is `names.biddingPeriod + names.revealPeriod`, but reviewers should double-check me on that! If I pull out all the wallet checks and the txdb checks and the "will fail" tests, that value is enough blocks to pass the tests in `chain.verifyCovenants()`. Subtracting one block from that value will return a `bad-open-state` error from the chain.

test/wallet-test-auctions.js
```diff
@@ -0,0 +1,184 @@
+/* eslint-env mocha */
```
boymanjor: Do you mind moving this to the `auctions-test.js` file?

test/wallet-test-auctions.js
```diff
@@ -0,0 +1,184 @@
...
+});
+
+const cpu = miner.cpu;
+
+const wdb = new WalletDB({
+  network: network,
+  workers: workers
+});
+
+describe('Wallet Auction', function() {
```
boymanjor: I think that once we move this test to `auctions-test.js`, we should call it something like 'Test OPEN' and have it run after the 'Vickrey Auction' describe. What do you think?

test/wallet-test-auctions.js
pinheadmz: The reason I didn't do this at first is because the current `auctions-test` seems to be testing consensus rules against `miner` and `chain`. That test uses `memwallet` which doesn't even [test for double opens](https://github.com/handshake-org/hsd/blob/master/lib/wallet/wallet.js#L1542-L1543).

Since this PR is focused on wallet behavior I thought it was best to start a new test file (other wallet auction functions can be added in here later). I could add `acutions-test` if you want, just wanted to throw out my motivation there...

boymanjor: I see! Ok, can we call it `wallet-auctions-test.js` to stick to the current naming convention?

test/wallet-test-auctions.js
boymanjor: Resolved out of band.

pinheadmz:
@boymanjor I renamed the test file, then rebased on master. Turned out my test was interfering with other tests added since I first posted. So I added one more commit to play nice with others :-)
If that works for you, I'll squash.

test/wallet-test-auctions.js
pinheadmz: 👍

turbomaze:
Seems to be working on testnet4! https://hnscan.com/name/chain

pinheadmz:
Rebased on master, just in case - didn't break any tests locally.

rsolari:
👍 Thanks! This change is an improvement on the current behavior. 

One note is that the new behavior will no longer block anyone from sending an `OPEN` for a `CLOSED` auction, even though a closed auction might have a winner. That seems ok, as long as there's some consensus rule that prevents that open from landing on the chain. The user just won't get feedback that their `OPEN` is invalid.

pinheadmz:
@rsolari were you able to OPEN a closed auction with a winner? That check should be happening already, out of the scope of this PR. If it can be done, we should address that I might've missed that possibility

rsolari:
I didn’t run the test myself, but I’m afk for the weekend. I’ll check when
I get back to a computer on Monday.

test/wallet-auction-test.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    assert(err);
+    assert.strictEqual(err.message, 'Name is not available.');
+  });
+
+  it('should mine enough blocks to expire auction', async () => {
+    for (
+      let i = 0;
+      i < network.names.biddingPeriod + network.names.revealPeriod;
```
tynes: Nit: precalculate `network.names.biddingPeriod + network.names.revealPeriod` into a variable as to not need to use multiple lines for the `for` statement

test/wallet-auction-test.js
```diff
@@ -0,0 +1,191 @@
...
+      await wdb.addBlock(entry, block.txs);
+    });
+  });
+
+  it('should add addrs to miner', async () => {
+    const addr = await winner.createReceive();
+    miner.addresses = [addr.getAddress().toString(network)];
+  });
+
+  it('should mine 20 blocks', async () => {
```
tynes: Is it possible to mine less blocks as to keep the test suite lighter? Could also move this out of an `it` and into `before`

test/wallet-auction-test.js
```diff
@@ -0,0 +1,191 @@
...
+    job.addTX(openAuctionMTX2.toTX(), openAuctionMTX2.view);
+    job.refresh();
+
+    const block = await job.mineAsync();
+
+    assert(await chain.add(block));
+  });
+
+  it('should cleanup', async () => {
+    await miner.close();
```
tynes: This could go in the `after`

test/wallet-auction-test.js
```diff
@@ -0,0 +1,191 @@
...
+
+    winner = await wdb.create();
+
+    chain.on('connect', async (entry, block) => {
+      await wdb.addBlock(entry, block.txs);
+    });
+  });
+
+  it('should add addrs to miner', async () => {
+    const addr = await winner.createReceive();
```
tynes: This could go in the `before`

pinheadmz:
@tynes thanks for the feedback, nits addressed and test is a lot cleaner. PR commits are cleaner. Rebased on master.

@rsolari although this PR doesn't test it explicitly (a fullnode would be required, ie integration test instead of unit test) I did run tests with the CLI using separate wallets and the behavior was as expected: unchanged. Namely, I got the error `Name is not available.` after an auction is closed with a winner, as well as any other state where a name is already `OPEN`.

rsolari:
@pinheadmz Thanks !

I also did some manual testing and correctly got `Name is not available` after an auction closed.

lib/wallet/txdb.js
```diff
@@ -827,9 +827,20 @@ class TXDB {
...
+      const hash = await this.bucket.get(key);
 
-      if (await this.bucket.has(key))
-        return true;
+      // Allow a double open if previous auction period has expired
+      // this is not a complete check for name availability or status!
+      if (hash) {
+        const names = this.wdb.network.names;
+        const period = names.biddingPeriod + names.revealPeriod;
+        const oldTX = await this.getTX(hash);
```
tynes: Nit: remove brackets for 1 line if statement, `oldTX` -> `tx` variable naming for minimal style

lib/wallet/txdb.js
```diff
@@ -827,9 +827,20 @@ class TXDB {
...
+        if (oldTX.height !== -1 && oldTX.height + period < this.wdb.height) {
+          return false;
+        } else {
```
tynes: Nit: Unnecessary `else`

pinheadmz: fixed.

lib/wallet/txdb.js
pinheadmz: np removing brackets, thanks.
Don't wanna rename `oldTX` because variable `tx` is already defined as parameter to the function, its the transaction that _might_ be a double open.

pinheadmz:
@tynes nits addressed, rebased on master.

pinheadmz:
@boymanjor thanks dude, nits addressed, rebased on master to check test, and force pushed.

PR #99, opened 2019-02-08
test: include coin serialization tests

tynes:
This test shows that the coin is properly instantiated
for each network fromJSON and then can be written to
a buffer writer and then recovered from the buffer
writer with all of the same properties.


boymanjor:
@tynes this looks good to merge. Do you mind updating the commit message to match the title?

PR #101, opened 2019-02-14
wallet: prevent sendupdate with bad resource

pinheadmz:
To reproduce: Win an auction, then attempt to `update` with malformed Resource, wasting your miner fees:
```
$ hsw-rpc --network=regtest sendupdate hello '{"MyCatsBreath":["SmellsLikeCatFood"]}'
```
...results in a tx with a covenant with sad empty bytes:
```
"covenant": {
  "type": 7,
  "action": "UPDATE",
  "items": [
    "3338be694f50c5f338814986cdf0686453a888b84f424d792af4b9202398f392",
    "81040000",
    "00000000"
  ]
}
```
This PR checks the resource for at least one value in any of the possible fields (`ns`, `service`, `uri`...) and if all fields are empty arrays, we throw:
```
Resource is empty
```
I didn't write the counter-test yet (successful update with correctly formed resource) because that also requires an auction process. If #94 gets merged, I can write more tests in the `wallet-test-auctions.js` file introduced by that PR

### DIFF
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -1551,6 +1551,30 @@ class Resource extends Struct {
 
     return this;
   }
+
+  isSane() {
+    if (
+      this.canonical ||
+      this.delegate ||
+      this.ns.length ||
+      this.service.length ||
+      this.uri.length ||
+      this.email.length ||
+      this.text.length ||
+      this.location.length ||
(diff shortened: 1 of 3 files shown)


pinheadmz:
Rebased on master, renamed commit.

Issue #102, opened 2019-02-15
Add READMEs for Internal Module Folders


aunyks:
I think there should be READMEs that provide quick (or descriptive) overview's of the contents of internal modules throughout the project, particularly those in `lib/`. Something similar to LND: [one](https://github.com/lightningnetwork/lnd/tree/master/lnrpc), [two](https://github.com/lightningnetwork/lnd/tree/master/tor), and [three](https://github.com/lightningnetwork/lnd/tree/master/zpay32).  

I think this'd lower the barrier to first PR for new developers looking at the project, as the READMEs could speed up understanding the codebase, how it's organized, and how certain components relate to others.

I'm willing to open PRs for this, should the additions be favorable.

boymanjor:
@aunyks I think this is a great idea! If you want to get started, I can help out with review and answer questions about the codebase.

PR #109, opened 2019-03-06
all: add READMEs to bin/ and test/

aunyks:
Adding READMEs to some modules, in reference to #102.

Also, should there exist one of these READMEs for every module in `lib/`?


tynes:
@aunyks Thank you for the contribution!

> Also, should there exist one of these READMEs for every module in lib/?

concept ACK on having a README in every module in `lib` btcd style. The more docs that we have, the easier it will be for contributors to get started and alternative implementations to be built

PR #90, opened 2019-01-18
Add mirror RPC methods to generate transactions

mslipper:
As discussed in #89, this PR adds support for transaction-creating mirror RPC methods.

How are RPC endpoints tested? I don't see any unit tests for the RPC layer right now.

PR #110, opened 2019-03-11
wallet: support the creation of unsigned auction txs

boymanjor:
This PR introduces new RPC calls, for each covenant type, which create unsigned txs to be signed offline. This PR also introduces a paths object to the MTX class. This object is used to indicate the HD paths of the signing keys. The MTX class has also been updated to persist coin information during de/serialization.

This functionality allows hardware wallet signing. End-to-end tests can be found in [hsd-ledger][hsd-ledger] and require a Ledger Nanos S hardware wallet.

This PR supersedes #90.

[hsd-ledger]: https://github.com/boymanjor/hsd-ledger/blob/master/test/e2e/hsd-test.js

### DIFF
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1300,7 +1300,12 @@ class MTX extends TX {
    */
 
   format() {
-    return super.format(this.view);
+    const mtx = super.format(this.view);
+
+    if (this.paths)
+      mtx.paths = this.paths;
+
+    return mtx;
   }
 
   /**
@@ -1309,7 +1314,12 @@ class MTX extends TX {
    */
 
   toJSON() {
-    return super.getJSON(null, this.view);
+    const json = super.getJSON(null, this.view);
+
+    if (this.paths)
+      json.paths = this.paths;
+
+    return json;
   }
 
   /**
@@ -1319,7 +1329,12 @@ class MTX extends TX {
    */
 
   getJSON(network) {
-    return super.getJSON(network, this.view);
+    const json = super.getJSON(network, this.view);
+
+    if (this.paths)
+      json.paths = this.paths;
+
+    return json;
   }
 
   /**
@@ -1329,6 +1344,9 @@ class MTX extends TX {
 
   fromJSON(json) {
     super.fromJSON(json);
+
+    if (json.paths)
+      this.paths = json.paths;
 
     for (let i = 0; i < json.inputs.length; i++) {
       const input = json.inputs[i];
@@ -1343,6 +1361,7 @@ class MTX extends TX {
       coin.index = prevout.index;
 
       this.view.addCoin(coin);
+      this.inputs[i].coin = coin;
     }
 
     return this;
--- a/lib/wallet/layout.js
+++ b/lib/wallet/layout.js
@@ -16,7 +16,7 @@ const bdb = require('bdb');
  *  D -> wallet id depth
  *  p[addr-hash] -> wallet ids
  *  P[wid][addr-hash] -> path data
- *  r[wid][index][hash] -> path account index
+ *  r[wid][index][hash] -> dummy
  *  w[wid] -> wallet
  *  W[wid] -> wallet id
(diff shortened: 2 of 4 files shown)


chjj:
@boymanjor, we need to design this better. We shouldn't be dynamically extending objects. There's also no reason for adding `coin` to an input -- we have coin views for that.

boymanjor:
@chjj if you do not have a problem with the implementation of the `create*` RPCs I can split that out into a separate PR on your `testnet4-next` branch and remove the extra `paths` param from the options hash.

Re: dynamically extending objects -- does that mean we should add a `path` attribute to the `Coin` and `CoinEntry` objects? I originally wanted to avoid updating the db (we will probably need signing paths in `bcoin` too), but I see that we are already passing a `path` object to `txdb.saveCredit`. We could append the `path` attribute to `credit.coin` before storing it in the db, and add it to the `CoinEntry` objects when constructing the coin views.

Re: coin views -- my bad. I noticed we were serializing the inputs with the coin data appended to the inputs and thought we could add them back on. I now realize that we do this because there is no "coin view" on the JSON serialized tx and we needed somewhere to keep this info. There is no need to duplicate it when holding the `MTX` in memory.

chjj:
> if you do not have a problem with the implementation of the create* RPCs I can split that out into a separate PR on your testnet4-next branch and remove the extra paths param from the options hash.

I'd have to look at it again, but for now I reverted it on the testnet branch. I don't have a problem with them, but I think we should eventually start looking to turn all the naming/auction stuff into REST APIs. The RPC was a temporary hack to get things rolling (though, it might end up being one of those "temporary-permanent" solutions which are so popular in software development). 

The RPC kind of sucks because it doesn't support multi-wallet. It was only supposed to be there for bitcoind compat, but here we are (my fault). I'm seriously at a crossroads personally here since I hate REST and I hate the JSON RPC api. I want to rip out _all_ of the rest/json-rpc/websocket api and start from scratch, but it would be a hell of an undertaking.

> does that mean we should add a path attribute to the Coin and CoinEntry objects?

We shouldn't do that as it will use loads more memory when syncing the chain. Every property on an object incurs a cost and consensus > wallet.

If you need the paths, it might be possible to create some kind of WalletCoinView class that inherits from CoinView which also stores the paths. Maybe exposing some method like `getPath(hash, index)`. On the parent CoinView, we can just have this method return null. Would that work?

boymanjor:
>I'd have to look at it again, but for now I reverted it on the testnet branch. I don't have a problem with them, but I think we should eventually start looking to turn all the naming/auction stuff into REST APIs. The RPC was a temporary hack to get things rolling (though, it might end up being one of those "temporary-permanent" solutions which are so popular in software development).

For now, I'll reimplement the RPCs. After we finish testing the wallet and investigating some bugs, @tynes and I plan to focus on the REST APIs for naming/auction txs.

>If you need the paths, it might be possible to create some kind of WalletCoinView class that inherits from CoinView which also stores the paths. Maybe exposing some method like getPath(hash, index). On the parent CoinView, we can just have this method return null. Would that work?

I think this will work. I would suggest `getPath(address)` as the signature because this matches `Wallet.getPath(address)`. The internal map can be a `BufferMap` of address Hash to String. I propose a String as the value because the Path serialization needs to account for the current watch-only wallet bug which allows the autoincrementing account index to _not_ match the actual derivation path of the account key. Bypassing this bug involves an async call to retrieve the account key, and I think we should only do this once, but I am open to alternatives.

`MTX.view` could now be a WalletCoinView and we could add a `Wallet.getWalletCoinView(mtx)` method to retrieve the signing paths. This method could alternatively take an options hash, e.g. `{ view: ... }` or CoinView as a second argument. Passing this argument would force the method to use the provided coins and skip async calls to the db for coins that the mtx is probably already holding.

All that would be left to do is update the `TX` JSON serialization methods to become aware of the paths. We could store them on their corresponding inputs during serialization and return them to the coin view during deserialization a la the coin objects. What do you think about this approach @chjj?

PR #112, opened 2019-03-15
rpc: new wallet rpc 'getaddressinfo'

tynes:
Update: integrated changes done here - https://github.com/bcoin-org/bcoin/pull/731
Update: This PR has been merged into `bcoin`

## getaddressinfo

This ports the `bitcoind` rpc method `getaddressinfo` to the Handshake wallet.
It includes `ismine` and `iswatchonly` values that can use the wallet database
to return the correct values since `validateaddress` hard codes in `ismine: false`.
This can throw off people who want to build on top of handshake, the `getaddressinfo` rpc method can help users who desire an accurate `ismine` property. 
The full list of properties that it returns are `address`, `ismine`, `iswatchonly`, `ischange`, `isspendable`, `isscript`, `isspendable`, `witness_version` and `witness_program`.
`isspendable` is a Handshake specific property derived from `Address.isUnspendable`
which calls this method:

```javascript
  isNulldata() {
    return this.version === 31;
  }
```

https://bitcoincore.org/en/doc/0.17.0/rpc/wallet/getaddressinfo/

## validateaddress

This updates the `validateaddress` RPC method to match `bitcoind` v0.17.0 by adding
`witness_version`, `witness_program`, and `isscript`. It also includes the Handshake
specific property `isspendable`.
https://bitcoincore.org/en/doc/0.17.0/rpc/util/validateaddress/

### DIFF
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -18,7 +18,6 @@ const KeyRing = require('../lib/primitives/keyring');
 const Input = require('../lib/primitives/input');
 const Outpoint = require('../lib/primitives/outpoint');
 const Script = require('../lib/script/script');
-const HD = require('../lib/hd');
 const PrivateKey = require('../lib/hd/private.js');
 
 const KEY1 = 'xprv9s21ZrQH143K3Aj6xQBymM31Zb4BVc7wxqfUhMZrzewdDVCt'
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2093,8 +2093,10 @@ class RPC extends RPCBase {
     return {
       isvalid: true,
       address: addr.toString(this.network),
-      ismine: false,
-      iswatchonly: false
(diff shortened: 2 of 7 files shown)


test/wallet-test.js
```diff
@@ -18,7 +18,6 @@ const KeyRing = require('../lib/primitives/keyring');
 const Input = require('../lib/primitives/input');
 const Outpoint = require('../lib/primitives/outpoint');
 const Script = require('../lib/script/script');
-const HD = require('../lib/hd');
```
tynes: This was not used in the file and was causing the linting to fail

tynes:
I think it would also be useful to include a key `isunspendable`

tynes:
Split into 2 PRs
- https://github.com/handshake-org/hsd/pull/123
- https://github.com/handshake-org/hsd/pull/122

PR #114, opened 2019-03-15
test: add ci

tynes:
This adds CI using Circle CI, copied directly from bcoin
at commit df1e8c06216f919d90c86cc89784f4c65d66eff3

The Circle CI project needs to be registered still
It would be really helpful if a project admin could go here to set it up:
https://circleci.com/add-projects/gh/handshake-org

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,61 @@
+defaults: &defaults
+  working_directory: ~/repo
+  docker:
+    - image: circleci/node:10.2
+
+version: 2.0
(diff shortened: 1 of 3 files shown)


tynes:
Looks like the tests failed because of a timeout, the miner took too long and the next test depended on mining work that was done in the test that timed out

boymanjor:
@tynes increase the timeout and retrigger the job.

tynes:
@boymanjor Everything good to go 👍

PR #115, opened 2019-03-19
test: reorganize test directory

tynes:
Reorganizes the tests to make it easier to contribute and understand the coverage.
At a high level, the `lib` directory is now mirrored in the `test` directory such that
`test/primitives` contains the unit tests for `lib/primitives`
Also adds a `integration` directory for integration tests and a WIP high level `README`
to document the testing setup


package.json
```diff
@@ -65,8 +65,9 @@
   "scripts": {
     "lint": "eslint $(cat .eslintfiles) || exit 0",
     "lint-file": "eslint",
-    "test": "bmocha --reporter spec test/*.js",
-    "test-browser": "NODE_BACKEND=js bmocha --reporter spec test/*.js",
+    "test": "bmocha --reporter spec test/**/*.js",
+    "test-integration": "bmocha --reporter spec test/integration/*.js",
+    "test-browser": "NODE_BACKEND=js bmocha --reporter spec test/**/*.js test/*.js",
```
tynes: delete `test/*.js`

tynes:
@boymanjor Just squashed this into a single commit

boymanjor:
@tynes I updated the readme, rebased with `master`, squashed the commit and added myself as a co-author. If this looks good to you, I think we're ready to merge.

tynes:
LGTM

test/README.md
```diff
@@ -0,0 +1,48 @@
...
+packages found in the `lib` directory. This directory
+structure helps to keep tests scoped well and organized.
+There is also an `integration` directory specifically for
+integration tests. The test files themselves have suffixes
+that correspond to their type of test. The different
+types include `unit`, `integration` and `e2e`. This was
+chosen because the testing framework `bmocha` can `grep`
+on filename. For more documentation on `bmocha`, see
+[here](https://github.com/bcoin-org/bmocha).
+
```
tynes: One of the scripts is off and we do not mention it in the `README`

```
"test-browser": "NODE_BACKEND=js bmocha --reporter spec test/**/*.js",
```

This script is misleading because it doesn't actually run the code in a browser, it just sets the
`NODE_BACKEND` to `js`. There is a way to set the `NODE_BACKEND` in `bmocha` using the
`-B` flag. See https://github.com/bcoin-org/bmocha/blob/fa388102f751f6d2debaea90308d53a67c71d78c/README.md

It is possible to run tests in the browser, see:
https://github.com/bcoin-org/bmocha/blob/fa388102f751f6d2debaea90308d53a67c71d78c/README.md#headless-chrome--browser-support

tynes: The tests are significantly slower

```bash
$ npm install --save-dev browserify
$ ./node_modules/.bin/bmocha -H test/**/*.js
```

tynes: If thought that the tree may not be able to run in browsers, but it looks the there is a Memory File System
that is used if `this.options.treePrefix` is `undefined`.
https://github.com/handshake-org/hsd/blob/ff133262ae7b83dbe86c4d96c55cca5451743bf0/lib/blockchain/chaindb.js#L51

We could set this up to actually run the tests in the browser and differentiate between `browser` and `NODE_BACKEND=js`

tynes: I played around with trying to run tests in the browser. My proposal is to include `bpkg` as a `devDependency` so that it can be installed in the local `node_modules`, as it cannot be found if it is installed globally. Since `bpkg` does not have external dependencies, it should be safe to include as a `devDependency` specifically for testing

See: https://github.com/bcoin-org/bmocha/issues/6

boymanjor: @tynes I think differentiating between `browser` tests and `NODE_BACKEND=js` tests is a good idea. How about we get the tests running in headless chrome, but do not trigger them on every commit? We can run them locally to check for browser compatibility. We can call the other script `test-js-backend`.

boymanjor: > I played around with trying to run tests in the browser. My proposal is to include `bpkg` as a `devDependency` so that it can be installed in the local `node_modules`, as it cannot be found if it is installed globally. Since `bpkg` does not have external dependencies, it should be safe to include as a `devDependency` specifically for testing
> 
> See: [bcoin-org/bmocha#6](https://github.com/bcoin-org/bmocha/issues/6)

SGTM.

tynes: @boymanjor Pushed changes to add `bpkg` as a `devDependency` and update the npm scripts. Tests pass for each of `test`, `test-browser` and `test-js`

Corresponding PR in `bcoin`:
https://github.com/bcoin-org/bcoin/pull/740

test/script/script-unit.js
```diff
@@ -78,7 +78,8 @@ describe('Script', function() {
       input.execute(stack);
       output.execute(stack);
 
-      assert.deepEqual(stack.items, [[1], [3], [5]]);
+      const expected = [Buffer.from([1]), Buffer.from([3]), Buffer.from([5])];
```
tynes: These changes allow `npm test-browser` to pass without affecting `npm test`. Similar changes were bcoin here:
https://github.com/bcoin-org/bcoin/blob/01a71561b1677981a4b355f7deec46f50e68a855/test/script-test.js#L107

test/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Usage
+
+The tests are divided by type to allow isolated runs.
+
+To run all tests:
+```bash
+$ npm run test
+```
+
+To run all tests with the `bcrypto` JavaScript backend:
```
tynes: TODO: update this to match the latest commits

test/integration/mempool-integration.js
```diff
@@ -3,28 +3,28 @@
...
+const CoinView = require('../../lib/coins/coinview');
+const MemWallet = require('../util/memwallet');
 const ALL = Script.hashType.ALL;
 
 const ONE_HASH = Buffer.alloc(32, 0x00);
 ONE_HASH[0] = 0x01;
 
 const workers = new WorkerPool({
-  enabled: true
+  enabled: process.browser ? false : true
```
tynes: I haven't dug in deep enough yet to determine why the workers throw errors when running them in the browser. Turning them off allows the tests to pass. Ideally the workers should work fine in the browser, since they do in `bcoin`. @pinheadmz may have some insight into this

pinheadmz: I'm actually not entirely sure workers ever worked in the browser.
If you go to https://browser.c4yt.io and watch the `network` tab in browser dev tools, you'll see `/worker.js` is never loaded from the web server. [Even though it looks like it should be.](https://github.com/bcoin-org/bcoin/blob/master/browser/src/app.js#L66-L67)

tynes: Resolved based on https://github.com/bcoin-org/bcoin/commit/0b70a940a3f6e6dc8be13b844cda8e1b9d338a60

Issue #41, opened 2018-08-07
Bindings for bcrypto were not built.


popey:
Cloned the repo and tried to build on my Ubuntu 16.04 system.

```
alan@robby:~/src/hsd⟫ npm install --production
audited 930 packages in 5.098s
found 0 vulnerabilities

alan@robby:~/src/hsd⟫ bin/hsd 
Bindings for bcrypto were not built.
Please build them before continuing.
1 alan@robby:~/src/hsd⟫ node --version
v10.8.0
```

chjj:
Could you copy & paste your build log here? If bcrypto wasn't built, there was probably some build errors in there.

dylanbathurst:
same on arch
```
[dylan@archlinux hsd]$ rm package-lock.json
[dylan@archlinux hsd]$ npm install --production

> bcrypto@1.1.0 install /home/dylan/hsd/node_modules/bcrypto
> node-gyp rebuild || echo 'Build failed.'

make: Entering directory '/home/dylan/hsd/node_modules/bcrypto/build'
  CC(target) Release/obj.target/bcrypto/src/aead/aead.o
make: cc: Command not found
make: *** [bcrypto.target.mk:166: Release/obj.target/bcrypto/src/aead/aead.o] Error 127
make: Leaving directory '/home/dylan/hsd/node_modules/bcrypto/build'
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/lib/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:239:12)
gyp ERR! System Linux 4.14.66-1-lts
gyp ERR! command "/usr/bin/node" "/usr/lib/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/dylan/hsd/node_modules/bcrypto
gyp ERR! node -v v10.9.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! not ok
Build failed.

> bsip@0.1.1 install /home/dylan/hsd/node_modules/bsip
> node-gyp rebuild || echo 'Build failed.'

make: Entering directory '/home/dylan/hsd/node_modules/bsip/build'
  CXX(target) Release/obj.target/bsip/src/siphash.o
[...]

Issue #28, opened 2018-08-03
Error during Download and install hsd


pkrasam:
Erroring out at this installation step.

**Command**
$ npm install --production

**Error**
> node-gyp rebuild

  CXX(target) Release/obj.target/unbound/src/node_unbound.o
../src/node_unbound.cc:1:10: fatal error: 'unbound.h' file not found
#include <unbound.h>
         ^~~~~~~~~~~
1 error generated.
make: *** [Release/obj.target/unbound/src/node_unbound.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:258:23)
gyp ERR! stack     at emitTwo (events.js:126:13)
gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
gyp ERR! System Darwin 17.7.0
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/pk/Documents/code/handshake/hsd/node_modules/unbound
gyp ERR! node -v v8.11.3
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok

dylanbathurst:
I had a similar issue when trying to `npm install` using node v8.0.0. I bumped it up v10.1.0 and was able to install successfully.

chjj:
Hmm, unbound is an optional dependency. Is the build failing completely?

grimmy:
i installed `libunbound-dev` on debian and it worked fine then.

kargakis:
> Hmm, unbound is an optional dependency. Is the build failing completely?

It seems that the executable is created but I am unable to run it.

```console
$ ./bin/hsd
Bindings for bcrypto were not built.
Please build them before continuing.
```

kargakis:
```
npm install bcrypto https://github.com/cryptocoinjs/secp256k1-node
```
fixes the above error and gets me through to a different error:
```
/opt/hsd/node_modules/bindings/bindings.js:88
        throw e
        ^

Error: The module '/opt/hsd/node_modules/leveldown/build/Release/leveldown.node'
was compiled against a different Node.js version using
[...]

PR #117, opened 2019-03-22
readme: document minimum node.js version

tynes:
Documents to use Node.js v10 right at the top of the `README.md`

There have been multiple issues that people were having that could have to do with the version of Node.js that they were running.

See:
https://github.com/handshake-org/hsd/issues/41
https://github.com/handshake-org/hsd/issues/28

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 ## Install
 
+`hsd` requires Node.js v10 or higher
+
 ```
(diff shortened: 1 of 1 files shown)


boymanjor:
nit: do you mind changing the commit message header from `docs:` to `readme:`?

tynes:
@boymanjor Updated

PR #118, opened 2019-03-26
walletdb: remove references to witness

tynes:
This pull request removes dead code left over from `bcoin` in the `WalletDB`.
The `walletdb.witness` value was initially set to `false`, which doesn't make sense
in the context of Handshake.  I searched through the code for places where the
`witness` value could be referenced from outside the `WalletDB` class and could
not find any. The tests all still pass

### DIFF
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -2299,7 +2299,6 @@ class WalletOptions {
     this.compression = true;
 
     this.spv = false;
-    this.witness = false;
     this.checkpoints = false;
     this.wipeNoReally = false;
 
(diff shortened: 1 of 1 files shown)


boymanjor:
@tynes after some light digging, I do not think the `wdb.witness`is used in `bcoin` either. I did find a reference to `wdb.options.witness` here: https://github.com/bcoin-org/bcoin/blob/85ed59c842f2aa4c1a8154d74a9cad7696cdfee3/lib/wallet/account.js#L45.

`wdb.options` is a `WalletOptions` object instantiated with the options passed into the `WalletDB` constructor, so the values are the same. We should start a list of unused members on the `wdb` and other major components. As we prepare for breaking changes in `v2.0.0`, it will be useful to know what we can remove.

PR #119, opened 2019-03-26
all: fix linting errors

tynes:
Since we don't currently have any CI, a few linting errors sneaked into the codebase.
This pull request addresses the linting problems and now running `npm run lint` passes
without any errors. The tests all pass with `npm run test`


boymanjor:
nit: do you mind you changing the commit prefix from "lint" to "all." I believe using "all" does a good job signifying it is a change not localized to a specific file or module.

tynes:
@boymanjor Updated the commit message to `all: fix linting errors`

PR #120, opened 2019-03-27
test: wallet constructor

tynes:
Wallets are generally created through `walletDB.create`. This PR uses `walletDB.create` to
target the constructor of the `Wallet`, asserting that wallets are created as they should be.
A few options of the `Wallet` are not exposed through `walletDB.create`, so they are
tested directly via `Wallet.fromOptions`

This branch is based off of https://github.com/handshake-org/hsd/pull/115.
If we merge in that branch first, then I can remove the commits and rebase on the new `master`.


test/wallet/wallet-unit.js
```diff
@@ -32,51 +43,183 @@ describe('Wallet Unit Tests', () => {
 
       {
         // Should instatiate from String.
-        const wallet = await wdb.create({ master: xprv });
+        const walletdb = await wdb.create({ master: xprv });
```
boymanjor: `wdb.create` creates an instance of a `Wallet` not a `WalletDB`.

tynes: Updated this, good catch

nodech:
Expand tests with new options.

PR #122, opened 2019-04-02
rpc: new wallet rpc getaddressinfo

tynes:
Adds a new wallet RPC method getaddressinfo that
corresponds to changes with bitcoind. Returns values:

- address
- ismine
- iswatchonly
- ischange
- isspendable
- isscript
- witness_version
- witness_program

The ismine property previously existed on the node RPC
method validateaddress which was always hardcoded to true.
An additional PR will be added to update validateaddress
to remove the ismine property. isspendable refers to
an address with null data, of version 31.

See PR in bcoin: https://github.com/bcoin-org/bcoin/pull/731
See bitcoind docs:
https://bitcoincore.org/en/doc/0.17.0/rpc/wallet/getaddressinfo/

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -129,6 +129,7 @@ class RPC extends RPCBase {
     this.add('getaccountaddress', this.getAccountAddress);
     this.add('getaccount', this.getAccount);
     this.add('getaddressesbyaccount', this.getAddressesByAccount);
+    this.add('getaddressinfo', this.getAddressInfo);
     this.add('getbalance', this.getBalance);
     this.add('getnewaddress', this.getNewAddress);
     this.add('getrawchangeaddress', this.getRawChangeAddress);
@@ -447,6 +448,29 @@ class RPC extends RPCBase {
     }
 
     return addrs;
+  }
+
+  async getAddressInfo(args, help) {
+    if (help || args.length !== 1)
+      throw new RPCError(errs.MISC_ERROR, 'getaddressinfo "address"');
+
+    const valid = new Validator(args);
+    const addr = valid.str(0, '');
+    const address = parseAddress(addr, this.network);
+
+    const wallet = this.wallet.toJSON();
+    const path = await this.wallet.getPath(address);
+
+    return {
+      address: address.toString(this.network),
+      ismine: path != null,
+      iswatchonly: wallet.watchOnly,
+      ischange: path ? path.branch === 1 : false,
+      isspendable: !address.isUnspendable(),
+      isscript: address.isScripthash(),
+      witness_version: address.version,
+      witness_program: address.hash.toString('hex')
+    };
   }
 
   async getBalance(args, help) {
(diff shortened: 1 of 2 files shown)


test/wallet-rpc-test.js
```diff
@@ -0,0 +1,249 @@
+/* eslint-env mocha */
+
+'use strict';
+
+const {NodeClient,WalletClient} = require('hs-client');
+const assert = require('./util/assert');
```
tynes: TODO: use `bsert` instead

tynes:
Rebased on `master`

tynes:
Double check to make sure matches all updates from review on bcoin PR

tynes:
@boymanjor - Rebased on master, verified that feedback on the same PR to bcoin was applied to this branch. There are additional fields that `bitcoind` returns as part of this RPC, but those can be added in later as needed. A very important field returned is `ismine` for now, for anybody that wants to build an application on top of the wallet.

lib/wallet/rpc.js
```diff
@@ -449,6 +450,31 @@ class RPC extends RPCBase {
...
 
+  async getAddressInfo(args, help) {
+    if (help || args.length !== 1)
+      throw new RPCError(errs.MISC_ERROR, 'getaddressinfo "address"');
+
+    const valid = new Validator(args);
+    const addr = valid.str(0, '');
+    const address = parseAddress(addr, this.network);
+
+    const script = Script.decode(address.hash);
```
boymanjor: I'm not sure how this function call is working correctly. We are passing a hash to`Script.decode`, but it expects to parse a script.

tynes: You are right, good call. Fixed in d933131eb0e9a8f8c85a7aaba2919b5ad5f6554e

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,260 @@
...
+        assert.equal(response.ismine, false);
+      }
+    });
+
+    it('should return the correct program for a p2pkh address', async () => {
+      // m/44'/5355'/0'/0/0
+      const receive = 'rs1q4rvs9pp9496qawp2zyqpz3s90fjfk362q92vq8';
+
+      const address = Address.fromString(receive);
+      const program = Script.decode(address.hash).toHex();
```
boymanjor: We should remove this call to `Script.decode`.

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,260 @@
...
+      const addr = address.toString(network);
+      await wclient.execute('selectwallet', [watchOnlyWalletId]);
+      const response = await wclient.execute('getaddressinfo', [addr]);
+      assert.equal(response.witness_program, program);
+    });
+
+    it('should detect a p2wsh and its witness program', async () => {
+      const script = Script.fromMultisig(2, 2, pubkeys);
+      const address = Address.fromScript(script);
+      const program = Script.decode(address.hash).toHex();
```
boymanjor: We should remove this call to `Script.decode`.

tynes:
Updated to address @boymanjor's feedback

PR #127, opened 2019-04-03
test: fix values from updated block reward

kilpatty:
This commit updates the values expected from the `chain.db.value` in the chain and node tests. The block reward was increased from 1000 to 2000 at commit 601267d1b598617a4eb2cde7b906ab89c45c377c. This commit also updates the expected blocktemplate mintime which was updated in commit be4331fc8d63addb8e5fa6722ad29ac2a922b852.

### DIFF
--- a/test/http-test.js
+++ b/test/http-test.js
@@ -217,7 +217,7 @@ describe('HTTP', function() {
         + '0000000000000000000000000000000000000000000000000000000000000000'
         + 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
       curtime: json.curtime,
-      mintime: 1548148526,
+      mintime: 1554268736,
       maxtime: json.maxtime,
       expires: json.expires,
       sigoplimit: 80000,
--- a/test/node-test.js
+++ b/test/node-test.js
@@ -139,7 +139,7 @@ describe('Node', function() {
   });
 
   it('should have correct chain value', () => {
-    assert.strictEqual(chain.db.state.value, 23002210000);
+    assert.strictEqual(chain.db.state.value, 24002210000);
     assert.strictEqual(chain.db.state.coin, 21);
     assert.strictEqual(chain.db.state.tx, 21);
   });
@@ -176,7 +176,7 @@ describe('Node', function() {
   });
 
   it('should have correct chain value', () => {
-    assert.strictEqual(chain.db.state.value, 25002210000);
+    assert.strictEqual(chain.db.state.value, 26002210000);
     assert.strictEqual(chain.db.state.coin, 22);
     assert.strictEqual(chain.db.state.tx, 22);
   });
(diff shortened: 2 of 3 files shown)


kilpatty:
Not sure what's going on with the linting though. Looks like eslint isn't installing on Circle, but I don't see any changes to the CI script.

tynes:
It could be that some dependency that used to exist no longer exists because they changed their names again. Maybe its time to swap out eslint for bslint

kilpatty:
Yeah strange thing is that the version did change 5 days ago for eslint, but HSD has a handful of builds succeeding with the newest version 15.6 so bslint might be the solution.

tynes:
Could you squash this into a single commit?

kilpatty:
Okay squashed, and changed commit message to standard format

test/http-test.js
```diff
@@ -217,7 +217,7 @@ describe('HTTP', function() {
         + '0000000000000000000000000000000000000000000000000000000000000000'
         + 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
       curtime: json.curtime,
-      mintime: 1548148526,
+      mintime: 1554268736,
```
boymanjor: This is the value in the genesis block: https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/genesis.js#L109

No blocks were mined so we can expect the above + 1 because of this line: https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/node/rpc.js#L1492

test/node-test.js
```diff
@@ -139,7 +139,7 @@ describe('Node', function() {
   });
 
   it('should have correct chain value', () => {
-    assert.strictEqual(chain.db.state.value, 23002210000);
+    assert.strictEqual(chain.db.state.value, 24002210000);
```
boymanjor: The math for this change is correct, i.e.:
```
GENESIS_REWARD = (2000 + 2.21) * 1e6;
BASE_REWARD = 2000 * 1e6;
BLOCKS_MINED = 11;
GENESIS_REWARD + BASE_REWARD * BLOCKS_MINED = 24002210000;
```

test/node-test.js
```diff
@@ -176,7 +176,7 @@ describe('Node', function() {
   });
 
   it('should have correct chain value', () => {
-    assert.strictEqual(chain.db.state.value, 25002210000);
+    assert.strictEqual(chain.db.state.value, 26002210000);
```
boymanjor: Follows from above.

test/node-test.js
```diff
@@ -240,7 +240,7 @@ describe('Node', function() {
   });
 
   it('should have correct chain value', () => {
-    assert.strictEqual(chain.db.state.value, 27002210000);
+    assert.strictEqual(chain.db.state.value, 28002210000);
```
boymanjor: Follows from above.

PR #132, opened 2019-04-04
ci: install babel-eslint

kilpatty:
This commit adds babel-eslint to the circle ci install script. The handshake parser was updated to babel-eslint in commit 726f7317939c679066c3cc27fb768673b7ca7d4a.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
       - run: npm install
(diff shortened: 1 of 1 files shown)


tynes:
Do you mind breaking this PR up such that the only commit is the change to the `config.yaml` file?
We have been trying to enforce a commit message style like: `ci: install babel-eslint` where the prefix is the package name and the message is a declarative description of the change

kilpatty:
Will do!

chjj:
bslint now includes babel-eslint by default. We should switch to that.

PR #134, opened 2019-04-04
node/rpc: peer fullname, netaddress getkey

tynes:
Use the peer method fullname to render the
full address returned in the peer json propery
addr. This matches the hosts.json file, where
addr is pubkey@ip:port. This is useful information
for building automation and helpful when sharing
peers with other people, so that the hosts.json
file doesn't need to be inspected.

Add a getKey method to the NetAddress. The
method accepts an optional string that
determines the serialization of the return
value. The serialization can be base32, hex
or will otherwise return a buffer.

Add fullname and localname methods to the
Peer class. fullname wraps the peer's
address.fullname and localname returns
the local address's fullname. Since
the local address should share the same
key as the address, there is no need to
hold both in memory. localname renders
the the full hostname using the address's
key along with the local hostname and port.

The peer should return its identity key as part of
the API. The getpeerinfo response should now return
the fullname for the addrlocal field.

### DIFF
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -11,6 +11,7 @@ const EventEmitter = require('events');
 const {Lock} = require('bmutex');
 const {format} = require('util');
 const tcp = require('btcp');
+const IP = require('binet');
 const dns = require('bdns');
 const Logger = require('blgr');
 const {RollingFilter} = require('bfilter');
@@ -201,6 +202,28 @@ class Peer extends EventEmitter {
 
   hostname() {
     return this.address.hostname;
+  }
+
+  /**
+   * Getter to retrieve fullname.
+   * @returns {String}
+   */
+
+  fullname() {
+    return this.address.fullname();
+  }
+
+  /**
+   * Getter to retrieve local fullname.
+   * Pull the key from this.address to
+   * prevent storing the key twice
+   * @returns {String}
+   */
+
+  localname() {
+    const {hostname, port} = this.local;
+    const key = this.address.getKey();
+    return IP.toHostname(hostname, port, key);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


kilpatty:
@boymanjor anything more that can be done here to get this merged in?

tynes:
There are some tradeoffs on how this functionality is implemented. The `key` is never added to the local `Address` object.  That is why the `localname` method was added, to pull the key from `address` (that does have the key) and format it correctly with `local.{hostname,port}`.

Storing the key twice would make implementation more simple because then you could call `this.local.fullname();` and have it return the full address including the key. It would take up an extra 32+ bytes of memory for each peer.

I don't think that there are any cases in which the key would be different for the `local` address compared to the `address`.

kilpatty:
So this is just a thought I had last night, but perhaps we should add a test for connecting using just a pool's identity key. I have had trouble being able to get a node to connect to another using just this method, so before deciding on a design to expose it, we should triple check that we have a test to make sure it's functionality works. 

Might not be exactly relevant for this PR, but seems important before we get this merged.

lib/net/peer.js
```diff
@@ -203,6 +204,28 @@ class Peer extends EventEmitter {
...
+   * Getter to retrieve local fullname.
+   * Pull the key from this.address to
+   * prevent storing the key twice
+   * @returns {String}
+   */
+
+  localname() {
+    const {hostname, port} = this.local;
+    const key = this.address.getKey();
+    return IP.toHostname(hostname, port, key);
```
tynes: Look at the implementation of `IP.toHostName` to see if its simple enough to use here so that there is no need to include `binet`

tynes:
Closing: this functionality was merged in here https://github.com/handshake-org/hsd/commit/cc1ef7ab8dfcadbcc1b18b934ccf2d2ec1776042

PR #145, opened 2019-04-11
wallet: support alternative signers

boymanjor:
This PR introduces new RPC calls which create unsigned txs for each covenant type. It also adds a new WalletCoinView class that extends CoinView and stores HD paths. These paths allow alternative, HD wallets to create valid input signatures. The MTX class has also been updated to persist coin information during de/serialization. Finally, this commit adds support for unsigned auction transactions to the HTTP API.

End-to-end tests can be found in https://github.com/boymanjor/hsd-ledger and require a Ledger Nanos S hardware wallet.


lib/wallet/wallet.js
```diff
@@ -2968,6 +2969,10 @@ class Wallet extends EventEmitter {
     assert(mtx.verifyInputs(this.wdb.height + 1, this.network),
       'TX failed context check.');
 
+    // Set the HD paths.
```
tynes: Do you mean to always set the paths unless explicitly saying to not set them?

```javascript
> undefined !== false
true
```

lib/wallet/wallet.js
```diff
@@ -3430,6 +3435,52 @@ class Wallet extends EventEmitter {
...
   }
 
+  /**
+   * Get a wallet coin viewpoint with HD paths.
+   * @param {TX} tx
+   * @returns {Promise} - Returns {@link WalletCoinView}.
+   */
+
+  async getWalletCoinView(tx) {
+    let view = tx.view;
```
tynes: I think it would be better to have an assertion that `view` exists here because a method on it is called down the stack, then it can fail early with a nice error message instead of a `Cannot read property of undefined` error

https://github.com/handshake-org/hsd/blob/91f392ce2696cd550b1fb21668c34bf53b235ac0/lib/primitives/tx.js#L850

tynes:
What do you think about porting over the end to end tests? That would make this PR way larger and would cause a big rebase for https://github.com/handshake-org/hsd/pull/115, but we probably want to move them over here eventually

lib/wallet/path.js
```diff
@@ -212,14 +213,23 @@ class Path extends bio.Struct {
...
+   * @param {String|Network?} network - Network type.
    * @returns {String}
    */
 
-  toPath() {
+  toPath(network) {
     if (this.keyType !== Path.types.HD)
       return null;
 
-    return `m/${this.account}'/${this.branch}/${this.index}`;
```
tynes: I don't love how this function returns a different depth path depending on `network` being defined, but its a low touch change that works. While the only `purpose` right now is `44`, I can think of usecases for different purpose fields like rolling over the keys that control a name in a deterministic and auditable way that leaks no additional information about other transactions sent.  Thats definitely out of scope for this PR, but I think that its good to keep in mind. Also keep in mind the `PSBT_OUT_BIP32_DERIVATION` field in a [PSBT](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki), where `prefix` would be replaced with the fingerprint of `M` (depth 0 public key). These suggestions are more immediately applicable to bcoin, but could also be useful to hsd

lib/wallet/path.js
```diff
@@ -233,18 +243,46 @@ class Path extends bio.Struct {
...
+   */
+
+  fromJSON(json) {
+    assert(json && typeof json === 'object');
+    assert(json.derivation && typeof json.derivation === 'string');
+
+    const path = json.derivation.split('/');
+    const index = parseInt(path.pop(), 10);
+    const branch = parseInt(path.pop(), 10);
+    const account = parseInt(path.pop(), 10);
```
tynes: This will unharden the `account` which I believe is the expected behavior. I think a comment in the jsdoc could be helpful here

```javascript
> parseInt("0'", 10)
0
```

lib/wallet/wallet.js
boymanjor: This is unexpected behavior; great catch!

lib/wallet/wallet.js
boymanjor: I don't want the method to fail if the view is empty. How about something like this?

```
    let view = new CoinView();

    if (tx.view instanceof CoinView)
      view = tx.view;

    if (!tx.hasCoins(view))
      view = await this.txdb.getCoinView(tx);
```

Now we guard against `getWalletCoinView` passing an undefined `view` to `hasCoins`.

lib/wallet/path.js
boymanjor: > I don't love how this function returns a different depth path depending on network being defined, but its a low touch change that works.

I agree. I originally created a new method: `toFullPath` but I did not want to completely change the behavior of the Path object (particular `getJSON`).

Re: support for alternative purposes -- I think we should handle this by either a redesign to the Path object, or specific path objects for each specification of HD. The current object assumes BIP44 style derivation and makes assumptions about the purpose and coin type indices.

lib/wallet/path.js
boymanjor: You are correct; the behavior is expected. I can add a comment about this object assuming hardened derivation above the change level.

boymanjor:
> What do you think about porting over the end to end tests? That would make this PR way larger and would cause a big rebase for #115, but we probably want to move them over here eventually

I think you are right. I'll update the end-to-end tests from `hsd-ledger` to sign using another `hsd` wallet and merge them here.

boymanjor:
> > What do you think about porting over the end to end tests? That would make this PR way larger and would cause a big rebase for #115, but we probably want to move them over here eventually
> 
> I think you are right. I'll update the end-to-end tests from `hsd-ledger` to sign using another `hsd` wallet and merge them here.

On second thought. It is probably a better idea to add sufficient unit and integration tests here.

tynes:
What do you think about passing along `paths` here now that the function signature has been updated for `input.getJSON`? This is in `TX.getJSON`

https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/primitives/tx.js#L1749-L1752

lib/wallet/wallet.js
tynes: Generally instances of `TX` don't have the `CoinView` right? Would it be possible to skip the `tx instanceof MTX` check?

tynes:
@boymanjor Let me know if there is anything that I can do to help get this over the finishline

lib/primitives/mtx.js
```diff
@@ -15,7 +15,9 @@ const Input = require('./input');
 const Output = require('./output');
 const Coin = require('./coin');
 const Outpoint = require('./outpoint');
-const CoinView = require('../coins/coinview');
+const CoinView = require('../coins/coinview.js');
```
nodar-chkuaselidze: no need for `.js`

lib/primitives/mtx.js
```diff
@@ -31,7 +33,7 @@ const {types} = rules;
  * @alias module:primitives.MTX
  * @extends TX
  * @property {Number} changeIndex
- * @property {CoinView} view
+ * @property {WalletCoinView} view
```
nodar-chkuaselidze: In terms of abstractions, I am wondering if it makes sense for General MTX to have WalletCoinView, maybe change the name to something more appropriate ?

lib/wallet/path.js
nodar-chkuaselidze: I believe most of the codebase in bcoin assumes `network=main` if nothing is passed. Wallet always has information about network.

lib/wallet/path.js
nodar-chkuaselidze: maybe move path logic to `fromPath` (because there's `toPath`)

lib/wallet/wallet.js
boymanjor: >Generally instances of TX don't have the CoinView right? Would it be possible to skip the tx instanceof MTX check?

Yes, `TX` objects, generally, do not carry coin views. I included the check for an instance of `MTX` to allow the reader to understand that the caller of this function can pass an `MTX`. I think this function will be cleaner if we only expect `TX` objects as the first argument and allow the caller to pass a `CoinView` as an optional, second argument. What do you think?

lib/wallet/path.js
boymanjor: @nodar-chkuaselidze @tynes the issue here is about backward compatibility. As written, this change tries to introduce the ability to include the `purpose` (defaults to 44) and `coin_type` in the `derivation` attribute without breaking any current usage of `toPath`. Do you think it would be better to update `toPath` to always return a full, BIP44 path? If so, using `network=main` as a default makes a lot of sense. If we care about backward compatibility (more of an issue when we port this change to `bcoin`), then I do not believe we will want to use a default for `network`.

lib/primitives/mtx.js
boymanjor: Oh, good call. This comment is remnants of an older design. The general MTX will begin with a general CoinView object. I'll update the jsdoc.

lib/wallet/path.js
boymanjor: @nodar-chkuaselidze I think the `toPath` method is poorly named. The name suggests the method _creates_ a `Path` object, but it _actually_ creates a String representation of the derivation path.

To me, the name `fromPath` suggests it would take a `Path` object and return an instance of some other object, but given the context, it would need to return an instance of Path (or at least a hash with attributes used to construct a Path object).

This ambiguity makes me lean towards keeping the deserialization logic for the `derivation` attribute here.

boymanjor:
@tynes this is ready for another review.

lib/wallet/rpc.js
```diff
@@ -1402,34 +1413,65 @@ class RPC extends RPCBase {
...
+    };
+
+    const mtx = await wallet.createTX(options);
+
+    return mtx.getJSON(this.network);
+  }
+
+  _validateSendToAddress(args, help, method) {
+    const msg = `${method} "address" amount `
+      + '( "comment" "comment-to" subtractfeefromamount "account" )';
```
tynes: I notice that this has additional arguments (`comment`, `comment-to`) that are meant to keep it compatible with `bitcoind`. These arguments are not used.

https://bitcoincore.org/en/doc/0.17.0/rpc/wallet/sendtoaddress/

Adding `account` breaks the compatibility as `replaceable` is supposed to follow `subtractfeefromamount`.
I think that `account` should replace `comment` so that we can maintain compatibility, and then we have 1 more slot `comment-to` for something `hsd` specific.

lib/wallet/rpc.js
```diff
@@ -1912,61 +1998,149 @@ class RPC extends RPCBase {
...
-    if (help || args.length > 1)
-      throw new RPCError(errs.MISC_ERROR, 'sendreveal "name"');
-
+    const opts = this._validateReveal(args, help, 'sendreveal');
     const wallet = this.wallet;
-    const valid = new Validator(args);
-    const name = valid.str(0);
 
-    if (!name) {
+    if (!opts.name) {
```
tynes: What do you think about having a specific RPC for `sendrevealall` and `createrevealall` so that a bug doesn't cause the wallet to reveal everything when it was not intended to?

lib/wallet/rpc.js
```diff
@@ -1402,34 +1413,65 @@ class RPC extends RPCBase {
...
-    if (help || args.length < 2 || args.length > 5) {
-      throw new RPCError(errs.MISC_ERROR,
-        'sendtoaddress "address" amount'
-        + ' ( "comment" "comment-to" subtractfeefromamount )');
-    }
+    const opts = this._validateSendToAddress(args, help, 'sendtoaddress');
+    const wallet = this.wallet;
 
+    const options = {
+      subtractFee: opts.subtract,
```
tynes: Add `account` here so that it matches `createSendToAddress`?

lib/wallet/path.js
tynes: > I think the toPath method is poorly named.

I agree that it sounds like `toPath` would return an instance of the `Path` class.
What do you think of using `pop` vs `path[path.length - 1]`? I like the latter more personally because its not mutating the `path` array.

I think that its a good idea to assert that `path.length >= 3` because we don't want any of `account`, `branch` or `index` to be `undefined`.

lib/wallet/path.js
```diff
@@ -233,18 +243,46 @@ class Path extends bio.Struct {
...
+      derivation: this.toPath(network)
     };
   }
 
+  /**
+   * Inject properties from a json object.
+   * @param {Object} json
+   */
+
+  fromJSON(json) {
```
tynes: What about adding a `static fromJSON` constructor that wraps this?

test/path-test.js
```diff
@@ -0,0 +1,71 @@
...
+        change: false,
+        derivation: 'm/44\'/5355\'/0\'/0/0'
+      };
+
+      assert.deepStrictEqual(got, want);
+    }
+  });
+
+  it('should deserialize path', () => {
+    const path1 = new Path().fromJSON({
```
tynes: I think using a static fromJSON constructor here would make the style fit the rest of the codebase better.

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,315 @@
...
+      // Mine past OPEN period.
+      await mineBlocks(7, winner);
+    });
+
+    it('should create BID with signing paths', async () => {
+      // Create loser's BID.
+      await util.wrpc('selectwallet', [loser.id]);
+      assert(await util.wrpc('sendbid', [name, 4, 10]));
+
+      // Create, assert, submit and mine wnner's BID.
```
tynes: typo `wnner's`

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,315 @@
...
+    });
+
+    it('should create OPEN with signing paths', async () => {
+      // Create, assert, submit and mine OPEN.
+      const submit = true;
+      const json = await util.wrpc('createopen', [name]);
+      await processJSON(json, submit, winner);
+
+      // Mine past OPEN period.
+      await mineBlocks(7, winner);
```
tynes: The bidding period starts in the next `treeInterval`, its probably better to use that variable rather than a magic number

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,315 @@
...
+
+    it('should create TRANSFER with signing paths', async () => {
+      // Create, assert, submit and mine TRANSFER.
+      const submit = true;
+      const address = (await loser.createAddress('default')).address;
+      const json = await util.wrpc('createtransfer', [name, address]);
+      await processJSON(json, submit);
+    });
+
+    it('Should create TRANSFER cancellatio with signing paths', async () => {
```
tynes: typo `cancellatio`

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,315 @@
...
+      }
+
+      return mtx;
+    };
+
+    before(async () => {
+      util.node.network.coinbaseMaturity = 1;
+      await util.open();
+      winner = await util.wclient.wallet('primary');
+      loser = await util.wclient.createWallet('loser');
```
tynes: You could probably skip assigning  `await util.wclient.createWallet('loser');` to `loser` for safety

test/wallet-rpc-test.js
```diff
@@ -0,0 +1,315 @@
...
+    const name = rules.grindName(2, 0, Network.get('regtest'));
+    let winner, loser;
+
+    const mineBlocks = async (num, wallet, account = 'default') => {
+      const address = (await wallet.createAddress(account)).address;
+      const hashes = await util.nrpc('generatetoaddress', [num, address]);
+      await util.confirmBlock(hashes.pop());
+    };
+
+    const processJSON = async (json, submit, wallet = winner, show) => {
```
tynes: This could use some comments

test/path-test.js
```diff
@@ -0,0 +1,71 @@
...
+    assert.deepStrictEqual(path1, path2);
+
+    const got = path1;
+    const want = new Path();
+    want.name = 'default';
+    want.account = 0;
+    want.branch = 1;
+    want.index = 1;
+
+      assert.deepStrictEqual(got, want);
```
tynes: Nit: spacing off

test/path-test.js
```diff
@@ -0,0 +1,71 @@
...
+      account: 0,
+      change: true,
+      derivation: 'm/0\'/1/1'
+    });
+
+    const path2 = new Path().fromJSON({
+      name: 'default',
+      account: 0,
+      change: true,
+      derivation: 'm/44\'/5355\'/0\'/1/1'
```
tynes: Out of scope for this PR, but we should consider allowing for both `'` and `h` syntax for hardened paths so that we can skip out on having to escape the `'`

test/data/mtx1.json
```diff
@@ -0,0 +1 @@
...7f916ad","witnessHash":"b3f4428bde7742db4c6a65b70a1a3f9b64c32cc1d061ca352db00606f9938bb4","fee":2800,"rate":22764,"mtime":1558044979,"version":0,"inputs":[{"prevout":{"hash":"135a7ab2b078101a4b5c74ccb13ec33aafbfe3cb08d69f79bb6a07183ed4be4d","index":0},"witness":["","03253ea6d6486d1b9cc3ab01a9a321d65c350c6c26a9c536633e2ef36163316bf2"],"sequence":4294967295,"coin":{"version":0,"height":2,"value":2000000000,"address":"rs1q4rvs9pp9496qawp2zyqpz3s90fjfk362q92vq8","covenant":{"type":0,"action":"NONE","items":[]},"coinbase":true},"path":{"name":"default","account":0,"change":false,"derivation":"m/44'/5355'/0'/0/0"}}],"outputs":[{"value":99997200,"address":"rs1qxps2ljf5604tgyz7pvecuq6twwt4k9qsxcd27y","covenant":{"type":0,"action":"NONE","items":[]}},{"value":1900000000,"address":"rs1qsfcpg4w2qwzx6ht5aa30dj5vddwvkdfq82dlae","covenant":{"type":0,"action":"NONE","items":[]}}],"locktime":0,"hex":"0000000001135a7ab2b078101a4b5c74ccb13ec33aafbfe3cb08d69f79bb6a07183ed4be4d00000000ffffffff0210d6f5050000000000143060afc934d3eab4105e0b338e034b73975b1410000000b33f7100000000001482701455ca03846d5d74ef62f6ca8c6b5ccb352000000000000002002103253ea6d6486d1b9cc3ab01a9a321d65c350c6c26a9c536633e2ef36163316bf2"}
```
tynes: Could we prettify this json so its easier to validate against the test cases?

lib/primitives/input.js
```diff
@@ -169,10 +169,11 @@ class Input extends bio.Struct {
    * for JSON serialization.
    * @param {Network} network
    * @param {Coin} coin
+   * @param {Path} path
    * @returns {Object}
    */
 
-  getJSON(network, coin) {
+  getJSON(network, coin, path) {
```
tynes: Looking for additional places this is invoked besides in `lib/primitives/tx.js` and I don't see any that could use `path` being passed along.

It does get invoked in the `txdb` but I don't think that `path` is relevant here.
https://github.com/handshake-org/hsd/blob/325747e25a4843c374e37b96fbd161797052953b/lib/wallet/txdb.js#L3499

lib/wallet/path.js
tynes: >  Do you think it would be better to update toPath to always return a full, BIP44 path?

I think that this is ok for now but we should really figure out the right way to upgrade so that arbitrary bip44 paths are supported

lib/wallet/path.js
boymanjor: @tynes I do not assert the length of `path` explicitly, but `account`, `branch`, and `index` are asserted to match the values passed from the JSON object. That said, I'll add assertions for the proper lengths of 4 and 6 (the "m" will be present in the path array). Any other length will cause an assertion error.

Do you mind elaborating on your concerns regarding mutating the path array?

tynes: > Do you mind elaborating on your concerns regarding mutating the path array?

Its just an additional mental overhead for future people that would like to refactor the code. Not the worst, especially in this case since its a pretty simple codepath.

boymanjor: I'll add a comment that notes the path object is mutated.

lib/wallet/rpc.js
boymanjor: I think this is a great idea, but out of scope for this PR.

lib/primitives/input.js
boymanjor: Agreed. Thanks for the check.

tynes:
The `Details` class in `lib/wallet/txdb.js` may be relevant to this PR

https://github.com/handshake-org/hsd/blob/master/lib/wallet/txdb.js#L3339

test/auction-rpc-test.js
```diff
@@ -0,0 +1,317 @@
...
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const bio = require('bufio');
+const plugin = require('../lib/wallet/plugin');
+const rules = require('../lib/covenants/rules');
+const common = require('./util/common');
+const { ChainEntry, FullNode, KeyRing, MTX, Network, Path } = require('..');
```
tynes: Style nit, prefer no space in `{ }` unless between variable names

tynes:
Looks good to me

lib/wallet/rpc.js
boymanjor: We may want to support "comment" and "comment-to" eventually. Also, `replaceable` is a BIP125 specific argument. Since Handshake does not support BIP125, I prefer to favor backward compatibility with our current codebase, vs. compatibility with the `bitcoind` RPC interface. At least for `hsd`.

lib/wallet/path.js
```diff
@@ -212,14 +213,23 @@ class Path extends bio.Struct {
...
-    return `m/${this.account}'/${this.branch}/${this.index}`;
+    let prefix = 'm';
+
+    if (network) {
+      const purpose = 44;
+      network = Network.get(network);
+      prefix += `/${purpose}'/${network.keyPrefix.coinType}'`;
+    }
+
+    return `${prefix}/${this.account}'/${this.branch}/${this.index}`;
```
nodar-chkuaselidze: This might be an issue whenever `this.account != this.accountKey.accountDepth`..

PR #146, opened 2019-04-11
util: make hex32 function more concise.

kilpatty:
This commit takes the more concise hex32 function from node/rpc and
replaces the current util hex32 function with it. It then replaces all
calls to hex32 in node/rpc with util.hex32 to remain consistent with
all other files in the repo.

### DIFF
--- a/lib/utils/util.js
+++ b/lib/utils/util.js
@@ -100,28 +100,16 @@ util.time = function time(date) {
  */
 
 util.hex32 = function hex32(num) {
-  assert((num >>> 0) === num);
+  assert(num >= 0);
+
   num = num.toString(16);
-  switch (num.length) {
-    case 1:
-      return `0000000${num}`;
-    case 2:
-      return `000000${num}`;
-    case 3:
-      return `00000${num}`;
-    case 4:
-      return `0000${num}`;
-    case 5:
-      return `000${num}`;
-    case 6:
-      return `00${num}`;
-    case 7:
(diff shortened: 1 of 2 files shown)


tynes:
This could use some tests. I also think there was a reason to use the `switch` statement, its only creating one string instead of having to create a new string for each iteration in the `while` loop

kilpatty:
Interesting, that's a good point. I just saw the two different implementations and assumed that this one was better as it was newer (from what I can tell). I will definitely write some tests for this.

I think this also begs the question would we rather have these "helper functions" in each file or collect them in one location? e.g. util

chjj:
Did we benchmark this? I kind of preferred the unrolled version, but it's possible v8 optimizes the loop better (less code complexity).

lib/utils/util.js
```diff
@@ -100,28 +100,16 @@ util.time = function time(date) {
  */
 
 util.hex32 = function hex32(num) {
-  assert((num >>> 0) === num);
+  assert(num >= 0);
```
nodech, line 103: You can use `assert((num >>> 0) === num)` here (make sure num is 32bit) and get rid of `num.length <= 8` assert below.

nodech:
unrolled version is faster:
```
while:    ops=10000000, time=0.000087826, rate=113861498872.77116
unrolled: ops=10000000, time=0.000037609, rate=265893802015.47504
```

PR #147, opened 2019-04-14
docker: add gmp-dev to fix compilation issues on Alpine

kilpatty:
Recent versions of bcrypto have been linking to gmp.h which is causing
build issues inside of the alpine dockerfile.

This commit adds gmp-dev to the dependencies installed in the dockerfile
before npm install is called - allowing bcrypto to be compiled
successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY package.json \
 
 # Install build dependencies and compile
 FROM base AS build
-RUN apk add --no-cache g++ gcc make python2
+RUN apk add --no-cache g++ gcc make python2 gmp-dev
 RUN npm install --production
 
 FROM base


boymanjor:
@kilpatty nice detective work. Do you mind adding "Fixes #148." to the end of your description. GitHub will automatically close the corresponding issue.

For you reference: https://help.github.com/en/articles/closing-issues-using-keywords

kilpatty:
@boymanjor Yeah unfortunately this fixes the error messages associated with #148, but still returns a bcrypto: bindings not compiled. 

I was planning to leave this up in case it was a simple fix to bcrypto, otherwise I think our best bet may be to switch to a non-alpine based docker image for the time being. (Happy to push up a PR for that if need be, have one on standby)

chjj:
Whether it fixes something or not, it's probably good to have gmp installed (users will actually _need_ gmp for the airdrop verifications, so we should probably shout it from the rooftops). Merging.

PR #153, opened 2019-04-22
mempool: reflect spent coins in mempool coinview

tynes:
This updates the Mempool so that `mempool.getSpentView` specifically returns coins in the mempool that have been spent by other coins in the mempool.  Currently `mempool.getSpentView` just wraps `mempool.getCoinView`, so there is no functional distinction between the two methods.

It adds a test to cover the new functionality

This is a port of bcoin commit: 32cba1bf4aa293bcf49c1849587dc1482ce98050
https://github.com/bcoin-org/bcoin/commit/32cba1bf4aa293bcf49c1849587dc1482ce98050

### DIFF
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -756,6 +756,28 @@ class Mempool extends EventEmitter {
   }
 
   /**
+   * Check whether coin is still unspent.
+   * @param {Hash} hash
+   * @param {Number} index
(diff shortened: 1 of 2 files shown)


tynes:
While this is a port of code that already exists in `bcoin`, I think that a deeper investigation needs to happen to be certain that this function isn't used differently throughout the `hsd` codebase before it can be merged

PR #124, opened 2019-04-03
node-rpc: getblockhash returns string

pinheadmz:
In commit add3235d5b0de380257a6d069ca19ba717b60db4, little-endian hashes were removed from hsd (😌 !) an with it, a lot of conversions of hash buffers to reverse hex strings. I think this one just got missed.

#### Before:
```
$ hsd-rpc getblockhash 10
{
  "type": "Buffer",
  "data": [
    207,
    158,
    216,
    73,
    80,
    83,
    4,
    223,
    214,
    43,
    41,
    113,
    173,
    108,
    145,
    0,
    98,
    160,
    4,
    221,
    62,
    132,
    59,
    136,
    119,
    162,
    66,
    8,
    74,
    233,
    1,
    39
  ]
}
```

#### After:
```
$ hsd-rpc getblockhash 10
cf9ed849505304dfd62b2971ad6c910062a004dd3e843b8877a242084ae90127
```

PR #155, opened 2019-04-24
various: convert buffer to string for JSON

kilpatty:
```
This commit adds the toString('hex') function to the following places:
- primitives -> txmeta -> getJSON()
- primitives -> tx -> format()
- node -> rpc -> getblockhash
In addition it adds a small test to txmeta-test to ensure that the getJSON function is working accordingly. 
```

### DIFF
--- a/test/txmeta-test.js
+++ b/test/txmeta-test.js
@@ -10,7 +10,7 @@ const TXMeta = require('../lib/primitives/txmeta');
 const network = Network.get('regtest');
 
 describe('TXMeta', function() {
-  it('should return JSON for txmeta', async () => {
+  it('should return correct confirmations', async () => {
     // unconfirmed at height 100
     const txmeta1 = new TXMeta();
     const txJSON1 = txmeta1.getJSON(network, null, 100);
@@ -21,5 +21,14 @@ describe('TXMeta', function() {
     txmeta2.height = 100;
     const txJSON2 = txmeta2.getJSON(network, null, 100);
     assert.strictEqual(txJSON2.confirmations, 1);
+  });
+
+  it('should return blockhash as string', async () => {
+    // confirmed once at height 100
+    const block = Buffer.from('058f5cf9187d9f60729245956688022474ffc0eda80df6340a2053d5c4d149af');
+    const txmeta2 = TXMeta.fromOptions({ block });
+    const txJSON2 = txmeta2.getJSON(network, null, null);
+
+    assert.strictEqual(txJSON2.block, block.toString('hex'));
   });
 });
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -709,7 +709,7 @@ class RPC extends RPCBase {
     if (!hash)
       throw new RPCError(errs.MISC_ERROR, 'Not found.');
 
-    return hash;
+    return hash.toString('hex');
   }
 
   async getBlockHeader(args, help) {
--- a/lib/primitives/tx.js
+++ b/lib/primitives/tx.js
@@ -1671,7 +1671,7 @@ class TX extends bio.Struct {
 
     if (entry) {
       height = entry.height;
-      block = entry.hash;
+      block = entry.hash.toString('hex');
       time = entry.time;
       date = util.date(time);
     }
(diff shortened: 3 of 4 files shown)


kilpatty:
I am building out the openAPI spec for the hsd fullnode http requests right now, and kept coming across block: { data [23, 24, 200... etc]}, which I realized was likely the node returning a Buffer without first turning it into a string. 

This makes for a lot cleaner JSON in the http response, and the "format" function already returns the block data this way, so I figured it was a good quality of life increase for the code.

tynes:
Good find. This PR is pretty similar to https://github.com/handshake-org/hsd/pull/124, I wonder if there are other locations in the codebase where this happens

pinheadmz:
@kilpatty do you feel like adding a test for this change? It could be a real easy add here:
https://github.com/handshake-org/hsd/blob/master/test/txmeta-test.js

kilpatty:
@tynes I think quite a few, I'm trying to go through the http code right now with a fine-tooth comb so I'll hopefully find any changes in here, but I'm sure there are more scattered throughout the codebase. 

@pinheadmz Forsure, I can do that now! Should I put it in this same commit?

pinheadmz:
Actually if you feel like expanding this PR to cover all the occurrences, you can include the change in #124 (I'll just close my PR). You can add tests in the same or different commits for now, the whole PR is likely to get squashed into one before merge anyway.

kilpatty:
Definitely, more than happy to combine them all into one here. I'll try to finish it up by tonight and see if I can find any more occurrences.

kilpatty:
Okay I think I found all of them, I included your commit @pinheadmz and then I found one more in the TX primitive format() function. In our TXMeta primitive the format function returned a string rather than I buffer, so I figured that was intended in the TX primitive as well. 

I also added a small assert in the test for txmeta-test just to make sure that we feed it a buffer and it returns us a string. I think txmeta needs a lot more tests in general, but in my opinion that's probably out of the scope of this commit. I'll probably push up a more thorough test for txmeta in a separate PR later this week. 

Let me know if there is anything missing here!

test/txmeta-test.js
```diff
@@ -17,9 +17,11 @@ describe('TXMeta', function() {
     assert.strictEqual(txJSON1.confirmations, 0);
 
     // confirmed once at height 100
-    const txmeta2 = TXMeta.fromOptions( {height: 100} );
+    const block = Buffer.from('058f5cf9187d9f60729245956688022474ffc0eda80df6340a2053d5c4d149af');
+    const txmeta2 = TXMeta.fromOptions( {height: 100, block} );
     txmeta2.height = 100;
     const txJSON2 = txmeta2.getJSON(network, null, 100);
     assert.strictEqual(txJSON2.confirmations, 1);
+    assert.strictEqual(txJSON2.block, block.toString('hex'));
```
pinheadmz: Should be in its own test. That way, if the test fails we'll know what caused the regression. You could rename the existing test `should return correct confirmations` (thats my bad :-P) and yours would be like `should get block hash as string`

pinheadmz:
This is great dude, thanks! One last nit about formatting: now that we've expanded the scope a bit, I think the commit message and PR title should reflect that. Maybe something like
`various: convert buffer to string for JSON`, but we can wait on @boymanjor or @chjj to weigh on in style preferences.

test/txmeta-test.js
kilpatty: Okay perfect will do

kilpatty:
Alright I updated the test file to break this into it's own test, as well as updated the title + commit body

test/txmeta-test.js
```diff
@@ -22,4 +22,13 @@ describe('TXMeta', function() {
     const txJSON2 = txmeta2.getJSON(network, null, 100);
     assert.strictEqual(txJSON2.confirmations, 1);
   });
+
+  it('should return blockhash as string', async () => {
+    // confirmed once at height 100
+    const block = Buffer.from('058f5cf9187d9f60729245956688022474ffc0eda80df6340a2053d5c4d149af');
+    const txmeta2 = TXMeta.fromOptions( {block} );
```
tynes: Style nit: `({ block })` over `( {block} )`

kilpatty: I really need to set my eslint back up.... Fixing this now haha

chjj:
Looks like left-overs from the big buffer refactor. I'm glad that's over with. :)

PR #156, opened 2019-04-24
http: add check for height being below tip

kilpatty:
This commit simply adds an enforce line to check that the height being
requested to reset to is below the chain tip. This way we return a bad
request error rather than a internal server error.

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -378,6 +378,8 @@ class HTTP extends Server {
       const height = valid.u32('height');
 
       enforce(height != null, 'Height is required.');
+      enforce(height <= this.chain.height,
+        'Height cannot be greater than chain tip.');
 
       await this.chain.reset(height);
 


kilpatty:
The mentality I approached this with is that the more 500 errors we can remove due to bad inputs and replace them with 400s the better. This is a pretty simple check, but in my opinion returns better feedback to the user if they try to reset to a future block (say in the case that their node is not fully synced already)

lib/node/http.js
```diff
@@ -378,6 +378,8 @@ class HTTP extends Server {
       const height = valid.u32('height');
 
       enforce(height != null, 'Height is required.');
+      enforce(height <= this.chain.tip,
```
chjj: Should be `this.chain.height` or `this.chain.tip.height`.

kilpatty: Ah my fault, I think I accidentally dropped that on a formatting change. Should be fixed now, also switched to this.chain.height as that seemed to be used more than tip.height.

pinheadmz:
@kilpatty feel like porting this to bcoin as well?

kilpatty:
@pinheadmz happy to do so

PR #157, opened 2019-04-28
exports: export Covenant along with other primitives

kilpatty:
This commit simply adds the Covenant to the list of exports. It's a
primtive that I believe is used quite a bit by external developers, so
it would be very useful to have it exported easily rather than having
the developers manually point to the path, especially considering that
every other primitive that makes up a block -> transaction is exported.

### DIFF
--- a/lib/hsd.js
+++ b/lib/hsd.js
@@ -95,6 +95,7 @@ hsd.define('primitives', './primitives');
 hsd.define('Address', './primitives/address');
 hsd.define('Block', './primitives/block');
 hsd.define('Coin', './primitives/coin');
+hsd.define('Covenant', './primitives/covenant');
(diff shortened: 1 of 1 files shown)


lib/hsd.js
```diff
@@ -95,6 +95,7 @@ hsd.define('primitives', './primitives');
 hsd.define('Address', './primitives/address');
 hsd.define('Block', './primitives/block');
 hsd.define('Coin', './primitives/coin');
+hsd.define('Covenant', './primitives/covenant.js');
```
tynes: You can leave off the `.js` suffix, it would better match the rest of the style

kilpatty: Good catch

PR #158, opened 2019-04-28
coinentry: remove unused parameter from JSDoc

kilpatty:
This commit just removes the index parameter from jsdoc as the function
no longer requires that parameter at all.


tynes:
This looks good to me

Issue #143, opened 2019-04-10
Additional Wallet HTTP Endpoints


tynes:
# Overview

An even higher priority TODO than migrating the node RPCs to HTTP is moving the wallet RPCs to HTTP. In its current form, the wallet can only be safely accessed in a single threaded manner since you need to use the wallet RPC `selectwallet` to select which wallet you are interacting with (Thanks to @Thann for pointing that out). There is also no good way to specify specific accounts. This is important for @mslipper, @turbomaze, @BluSyn and anybody else that is building on top of the wallet. Please suggest any changes or updates based on your usage so far.

# Proposal

Additional Endpoints to be added:

| RPC        | HTTP           | Usage          | WalletClient Method |
| ------------- |:-------------:|:------:|:-----------------:|
| `getnames` |`GET /wallet/:id/names` | Get info on all names the wallet has bid/opened | `getNames`
| `getauctions` |`GET /wallet/:id/names` | Alias for above | `getNames`
| `getnameinfo`   | `GET /wallet/:id/name?name=foo` | Get info on a single bid/open  | `getName`
| `getauctioninfo` | `GET /wallet/:id/auction?name=foo` |  Monitor a single auction once the open period ends | `getAuction`
| `sendopen` | `POST /wallet/:id/open`         | Create an open transaction | `createOpen`
| `sendbid` | `POST /wallet/:id/bid`       | Create a bid transaction | `createBid`
| `getbids` | `GET /wallet/:id/bid?name=foo`       | Get bids for a specific name | `getBid`
| `sendreveal`| `POST /wallet/:id/reveal` | Create a reveal transaction | `createReveal`
| `getreveals` | `GET /wallet/:id/reveal?name=foo` | Get reveal for a specific name | `getReveal`
| `sendredeem` | `POST /wallet/:id/redeem` | Create a redeem transaction | `createRedeem`
| `sendupdate` | `POST /wallet/:id/update` | Create an update transaction | `createUpdate`
| `sendrenewal` | `POST /wallet/:id/renewal` | Create a renew transaction | `createRenew`
| `sendtransfer ` | `POST /wallet/:id/transfer` | Create a transfer transaction | `createTransfer`
| `sendfinalize` | `POST /wallet/:id/finalize` | Create a finalize transaction | `createFinalize`
| `sendcancel` | `POST /wallet/:id/cancel` | Cancel the transfer transaction | `createCancel`
[...]

Issue #125, opened 2019-04-03
A referral-only root zone (almost)


chjj:
After a long conversation with @mikedamm, he made a good case for making the root zone referral-only, at least in terms of DNS. It's an idea I had considered... I think right after I decided was fed up with the [complexity] of the resource serialization.

Putting non-referrals directly at the root is _odd_, and will likely confuse a lot of clients.

Note that this doesn't prevent anyone from putting arbitrary data at the root, or putting an A record directly at the apex of their zone: I could still put an A record directly at `chjj.` for example (by running my own nameserver). It may violate some RFCs, but I could do it.

I like this idea for its simplicity, and it mirrors the existing root zone.

It also allows us to rip out a ton of the resource serialization code, much of which is _not_ well thought out.

### Proposal for DNS Handshake Resources (V1)

- `DS` - A committed DS record.
- `NS` - A plain NS record with no glue (cannot point to a name under your own zone).
- `GLUE` - A name/ip pair which generates an NS record in the authority section and A/AAA records in the additional section (_must_ reside under your own zone -- this avoids DNS cache poisoning attacks, which are probably currently possible on testnet).
- `SYNTH` - An ipv4/ipv6 address which acts as a nameserver. This generates a _synthetic_ name for the NS record on the DNS side (similar to what we currently do if no name is provided). The IP itself appears in the additional section as glue.
- Other records up in the air.

This makes the JSON and binary serialization super simple. Example:

``` js
[
  {
    "type": "NS",
    "name": "ns1.chjjdns.example.com." // <-- must point to a another zone
  },
  {
    "type": "GLUE",
    "name": "ns1.chjj.", // <-- must point to our own zone
    "addr": [
      "127.0.0.1",
      "::1"
    ]
  },
  {
    "type": "SYNTH",
    "addr": "127.0.0.1" // <-- generates `_[base32]._hns.` with glue of 127.0.0.1
  },
  {
    "type": "DS",
    "keyTag": 1234,
    "algorithm": 8,
    "digestType": 2,
    "digest": "0000000000000000000000000000000000000000000000000000000000000000"
  }
]
```

Generates:

``` zone
;; AUTHORITY SECTION:
chjj. 86400 IN NS ns1.chjjdns.example.com.
chjj. 86400 IN NS ns1.chjj.
chjj. 86400 IN NS _ab7qaaab._hns.
chjj. 86400 IN DS 1234 8 2 0000000000000000000000000000000000000000000000000000000000000000
[...]

PR #163, opened 2019-05-01
http: add auction endpoints

tynes:
Adds a wallet HTTP endpoint for creating `OPEN` transactions and a more flexible interface compared to the RPC `sendopen`

### Change Summary
- Implements the wallet HTTP endpoints:
  - `GET /wallet/:id/name/:name`
  - `GET /wallet/:id/name`
  - `GET /wallet/:id/auction/:name`
  - `GET /wallet/:id/bid/:name`
  - `GET /wallet/:id/bid`
  - `GET /wallet/:id/reveal/:name`
  - `GET /wallet/:id/reveal`
  - `GET /wallet/:id/resource/:name`
  - `GET /wallet/:id/nonce/:name`
  - `POST /wallet/:id/open`
  - `POST /wallet/:id/bid`
  - `POST /wallet/:id/reveal`
  - `POST /wallet/:id/redeem`
  - `POST /wallet/:id/update`
  - `POST /wallet/:id/renewal`
  - `POST /wallet/:id/transfer`
  - `POST /wallet/:id/cancel`
  - `POST /wallet/:id/finalize`
  - `POST /wallet/:id/revoke`
- Each `POST` endpoint accepts similar options as `POST /wallet/:id/{create,send}`, except for the `outputs` option
- Each of the `wallet.make[covenant type]` (like `wallet.makeOpen`) functions have a type assertion for the `account` parameter updated to be valid for an integer or string, since the functions that consume the `account` parameter work with either the human readable account name (string) or the bip44 account index (integer). I think this solution is better than calling `wallet.getAccountIndex` on the human readable account name passed from the client 
- Tests the `account` parameter, `sign` parameter, `broadcast` parameter

Closes https://github.com/handshake-org/hsd/issues/143


tynes:
This needs an update to `hs-client` to wrap the endpoint `POST /wallet/:id/open`

It should also add a test for the `account` parameter

lib/wallet/wallet.js
```diff
@@ -1478,10 +1478,11 @@ class Wallet extends EventEmitter {
...
    * @param {Boolean} force
+   * @param {Object[]} outputs
    * @returns {MTX}
    */
 
-  async makeOpen(name, force, acct) {
+  async makeOpen(name, force, acct, outputs) {
     assert(typeof name === 'string');
     assert(typeof force === 'boolean');
     assert((acct >>> 0) === acct);
```
tynes: Add assertion here for outputs to match

lib/wallet/wallet.js
```diff
@@ -1478,13 +1478,15 @@ class Wallet extends EventEmitter {
    * @param {String} name
    * @param {Number} acct
```
tynes: Update the jsdoc to `{Number|String}`

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+  });
+
+  it('should mine to the primary/default wallet', async () => {
+    const height = 20;
+
+    for (let i = 0; i < height; i++)
+      await nclient.execute('generatetoaddress', [1, cbAddress]);
+
+    const info = await nclient.getInfo();
+    assert.equal(info.chain.height, height);
```
pinheadmz: Should we also check the generated blocks actually successfully mined to the primary/default wallet? Maybe get wallet history and check for `coinbase: true`. Same with next test. Mine as well check the height again too right?

pinheadmz: Should we also check the generated blocks actually successfully mined to the primary/default wallet? Maybe get wallet history and check for coinbase: true. Same with next test. Mine as well check the height again too right?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+    for (let i = 0; i < 5; i++)
+      await nclient.execute('generatetoaddress', [1, address]);
+
+    assert.ok(true);
+  });
+
+  it('should have no name state indexed', async () => {
+    const names = await wallet.getNames();
+
+    assert.equal(names.length, 0);
```
pinheadmz: How come you use `.equal()` and not `.strictEqual()`? Mine as well seal it up tight for regression tests, right?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+    node.mempool.on('tx', () => entered = true);
+
+    await sleep(100);
+
+    assert.equal(entered, true);
+    const mempool = await nclient.getMempool();
+
+    assert.ok(mempool.includes(json.hash));
+
+    const mtx = MTX.fromJSON(json);
```
pinheadmz: Do you need to convert json into an MTX here? I think the filter and test below will work directly on the JSON. This is also just testing the first line `wallet.createOpen`, maybe it should follow that creation? Because it's not testing a broadcast or mempool entry.

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+
+  it('should create an open and broadcast the transaction', async () => {
+    const json = await wallet.createOpen({
+      name: name
+    });
+
+    let entered = false;
+    node.mempool.on('tx', () => entered = true);
+
+    await sleep(100);
```
pinheadmz: You can try integrating @braydonf 's event wait-er form here: https://github.com/bcoin-org/bcoin/pull/748/files#diff-e169f8c8ca803ee76312e094ba6f17cbR105

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+    const info = await wallet.getAccount(accountTwo);
+    assert.equal(info.balance.tx, 0);
+    assert.equal(info.balance.coin, 0);
+
+    const fn = async () => (await wallet.createOpen({
+      name: name,
+      account: accountTwo
+    }));
+
+    assert.rejects(fn);
```
pinheadmz: Do you want to specify the error here?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+    for (let i = 0; i < treeInterval + 1; i++)
+      await nclient.execute('generatetoaddress', [1, cbAddress]);
+
+    await sleep(100);
+
+    const json = await wallet.createBid({
+      name: name,
+      bid: 1000,
+      lockup: 2000,
+      debug: true
```
pinheadmz: What's this `debug` doing?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+
+    assert(names.length > 0);
+    const [,ns] = names;
+
+    const auction = await wallet.getAuction(ns.name);
+
+    // auction info returns a list of bids
+    // and a list of reveals for the name
+    assert.ok(auction.bids);
+    assert.ok(auction.reveals);
```
pinheadmz: I think `assert.ok([])` passes, might want to tighten this test.

pinheadmz: Actually when I `console.log(auction)` right here I get `reveals: []` -- shouldn't there be a `REVEAL` from the last test?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+    const {revealPeriod} = network.names;
+    for (let i = 0; i < revealPeriod + 1; i++)
+      await nclient.execute('generatetoaddress', [1, cbAddress]);
+
+    await sleep(100);
+
+    const json = await wallet.createRedeem({
+      name: name
+    });
+
```
pinheadmz: I could be reading this wrong, but `wallet` had the LOWER bid, right? So shouldn't this fail? Shouldn't `wallet2` be able to succeed in creating a `REDEEM`? I'm sure this is checked in other tests, but I dunno if you also want to assert that the redeem values are correct (winning bid / 2nd-highest bid / change value)

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+      // update after register or update
+      const updates = json.outputs.filter(({covenant}) => covenant.type === types.UPDATE);
+      assert.equal(updates.length, 1);
+    }
+  });
+
+  it('should get name resource', async () => {
+    const names = await wallet.getNames();
+    // filter out to names that have data
```
pinheadmz: nit: "to" ?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+
+    await wallet.createUpdate({
+      name: name,
+      data: {
+        text: ['foobar']
+      }
+    });
+
+    // mine up to the earliest point in which a renewal
+    // can be submitted, a treeInterval into the future
```
pinheadmz: Will an attempted `REVEAL` fail if it's too early? Maybe that should be tested here first.

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+
+    const {receiveAddress} = await wallet.getAccount(accountTwo);
+
+    const json = await wallet.createTransfer({
+      name,
+      address: receiveAddress
+    });
+
+    const xfer = json.outputs.filter(({covenant}) => covenant.type === types.TRANSFER);
+    assert.equal(xfer.length, 1);
```
pinheadmz: Do you want to check the namestate here and confirm the transfer worked? Might be good just for extra validity check of the transfer tx.

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
...
+      const options = TransactionOptions.fromValidator(valid);
+      const mtx = await req.wallet.createBid(name, bid, lockup, options);
+
+      if (broadcast) {
+        const tx = await req.wallet.sendMTX(mtx, passphrase);
+        return res.json(200, tx.getJSON(this.network));
+      }
+
+      if (sign)
+        await req.wallet.sign(mtx, passphrase);
```
pinheadmz: Should we sign before broadcast? :-P Won't this return before signing?

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
...
+
+      const height = this.wdb.height;
+      const network = this.network;
+      const ns = await req.wallet.getNameStateByName(name);
+
+      res.json(200, ns.getJSON(height, network));
+    });
+
+    // Wallet Names/Auctions
+    this.get('/wallet/:id/names', async (req, res) => {
```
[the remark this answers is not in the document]
boymanjor: It is a special case of `GET /wallet/:id/name/:name`.

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
       await req.wallet.resend();
       res.json(200, { success: true });
     });
+
+    // Wallet Name
+    this.get('/wallet/:id/name', async (req, res) => {
```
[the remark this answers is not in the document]
boymanjor: Ditto for the following URLs:
- `GET /wallet/:id/auction/:name`
- `GET /wallet/:id/bid/:name`
- `GET /wallet/:id/resource/:name`

test/wallet-http.js
tynes: `REDEEM` is for the loser to get their money back. There should be an error if the winner tries to submit a `REDEEM` so this raises a good point around an additional test case

test/wallet-http.js
tynes: Good call, that is a typo

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
...
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
+      const height = this.wdb.height;
+      const network = this.network;
+
+      assert(name, 'Must pass name.');
+      assert(rules.verifyName(name), 'Must pass valid name.');
+
+      const ns = await req.wallet.getNameStateByName(name);
```
pinheadmz: Should this throw if name isn't found in wallet?  I tried breaking your test:
`Error: Cannot read property 'getJSON' of null`

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
...
+      const bids = await req.wallet.getBidsByName(name);
+      const reveals = await req.wallet.getRevealsByName(name);
+
+      const info = ns.getJSON(height, network);
+      info.bids = [];
+      info.reveals = [];
```
pinheadmz: Not including opens or redeems or whatever else here?

lib/wallet/http.js
tynes: `wallet.sendMTX` signs the transaction, which isn't immediately clear just by looking at this file.
https://github.com/handshake-org/hsd/blob/41ff9172ba32ce4b2af78dc11ec4045e0b37168e/lib/wallet/wallet.js#L3255

It was designed so that you could:

- sign and broadcast
- sign and not broadcast
- not sign and not broadcast

I believe this would allow for maximum flexibility when it comes to using `hsd` with watch only wallets and payment channels.

lib/wallet/http.js
```diff
@@ -833,6 +783,405 @@ class HTTP extends Server {
...
+      }
+
+      if (sign)
+        await req.wallet.sign(mtx, passphrase);
+
+      return res.json(200, mtx.getJSON(this.network));
+    });
+
+    // Create Renewal
+    this.post('/wallet/:id/renewal', async (req, res) => {
```
tynes: `renewal` vs `renew`?

test/wallet-http.js
```diff
@@ -0,0 +1,650 @@
...
+  });
+
+  it('should throw error with incompatible broadcast and sign options', async () => {
+    const fn = async () => await (wallet.createOpen({
+      name: name,
+      broadcast: true,
+      sign: false
+    }));
+
+    assert.rejects(fn, 'Must sign when broadcasting');
```
pinheadmz: Not sure why, but `assert.rejects()` doesn't seem to match the exact error message. I can tell because you are missing a period here :-) But any arbitrary string in this assertion will pass the test.

test/wallet-http.js
tynes: Good catch haha, was using that on the server side to pause the execution and introspect the state of things

lib/wallet/http.js
```diff
@@ -1153,6 +1502,77 @@ class HTTPOptions {
...
+    this.depth = valid.u32(['confirmations', 'depth']);
+    this.outputs = [];
+
+    if (valid.has('outputs'))
+      this.setOutputs(valid);
+
+    return this;
+  }
+
+  setOutputs(valid) {
```
boymanjor: Is there a reason we moved this logic into a method? If we do not plan to call this elsewhere, it makes sense to live this inside of `fromValidator`. 

If we leave this as a separate method, we should validate the input, i.e., `valid instanceof Validator`.

lib/wallet/http.js
```diff
@@ -1153,6 +1502,77 @@ class HTTPOptions {
...
+
+  /**
+   * Inject properties from Validator.
+   * @private
+   * @param {Validator} valid
+   * @returns {TransactionOptions}
+   */
+
+  fromValidator(valid) {
+    assert(valid);
```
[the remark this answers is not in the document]
tynes: I tried this approach, its not actually a `Validator`, its a `MultiValidator` which is not exposed from the `bval` library. Since the `Validator` is exported directly (not as a property of an object),
it makes it more difficult to upgrade. One solution is to add `MultiValidator` as a property of the `Validator` so that its possible to `assert(valid instanceof Validator.MultiValidator)`

https://github.com/bcoin-org/bval/blob/master/lib/validator.js#L836

test/wallet-http.js
tynes: Good catch, should use `strictEqual` to avoid errors based on type coercion

lib/wallet/http.js
tynes: Good catch here - there needs to be a test case for this, will add one

lib/wallet/http.js
tynes: This is what is indexed:

https://github.com/handshake-org/hsd/blob/41ff9172ba32ce4b2af78dc11ec4045e0b37168e/lib/wallet/layout.js#L97-L104

test/wallet-http.js
tynes: Good call, no need to convert to an `mtx`

lib/wallet/http.js
tynes: I'm going to skip returning anything else for now since it will dramatically increase the scope of this PR

lib/wallet/http.js
kilpatty: I'm pro -> use nouns in the URL and verbs as the http request type.

I think ```POST wallet renewal```, flows better than ```POST wallet renew``` since renewal is a noun. That being said, I'm not sure how the URLs in the node http calls are designed so I'd have to double check if that's a common thing to do.

lib/wallet/http.js
kilpatty: Just throwing this thought out there, but it would be nice to have a helper function that can group a Bid, Renewal, and Redeem into one object. That way you could just return that object, and it would encompass the 3 potential tx types. Likely out of scope for this PR, but wanted to suggest it anyway. 

Also I think only 1 open is valid per name now, so I think potentially returning that would be a good idea.

lib/wallet/http.js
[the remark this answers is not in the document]
tynes: Each of these has been updated in `hs-client` and here

lib/wallet/http.js
tynes: Added a testcase for this here https://github.com/handshake-org/hsd/pull/163/files#diff-7b92af4a1865adf03377c0804427d57dR535

lib/wallet/http.js
tynes: I was originally thinking that to return the open, a function could be added to the wallet `getOpenByName`. This would fit in with the current way that this handler is architected. I don't think that the open would be indexed unless you opened it using the wallet. Would it be useful to go back and index the open for any name that you bid on? Then a service could know who opened the auction, which is valuable information when participating in an auction.

> it would be nice to have a helper function that can group a Bid, Renewal, and Redeem into one object

Could you elaborate on this more? What is the usecase? Is it to have a global view or a personal view? As in, do you want to be able to query for *all* transactions referring to an auction? Or just transactions that this particular wallet account sent pertaining the auction?

lib/wallet/http.js
```diff
@@ -833,6 +784,441 @@ class HTTP extends Server {
...
+      assert(name, 'Must pass name.');
+      assert(rules.verifyName(name), 'Must pass valid name.');
+
+      const ns = await req.wallet.getNameStateByName(name);
+
+      if (!ns || ns.data.length === 0)
+        return res.json(404);
+
+      try {
+        const resource = Resource.decode(ns.data);
```
tynes: This means that it is not possible to get the data from the wallet API if it is not encoded properly. This relates to https://github.com/handshake-org/hsd/issues/125

lib/wallet/http.js
```diff
@@ -833,6 +784,473 @@ class HTTP extends Server {
...
+        info.bids.push(bid.toJSON());
+
+      for (const reveal of reveals)
+        info.reveals.push(reveal.toJSON());
+
+      return res.json(200, info);
+    });
+
+    // All Wallet Bids
+    this.get('/wallet/:id/bid', async (req, res) => {
```
tynes: Need to add a test for this codepath

lib/wallet/http.js
```diff
@@ -833,6 +784,473 @@ class HTTP extends Server {
...
+      for (const bid of bids) {
+        if (!own || bid.own)
+          items.push(bid.toJSON());
+      }
+
+      res.json(200, items);
+    });
+
+    // All Wallet Reveals
+    this.get('/wallet/:id/reveal', async (req, res) => {
```
tynes: Need to add a test for this codepath

lib/wallet/http.js
```diff
@@ -833,6 +784,473 @@ class HTTP extends Server {
...
+      try {
+        address = Address.fromString(addr, this.network);
+      } catch (e) {
+        return req.json(400);
+      }
+
+      const nameHash = rules.hashName(name);
+      const nonce = await req.wallet.generateNonce(nameHash, address, bid);
+      const blind = rules.blind(bid, nonce);
+
```
tynes: What is a useful JSON API to return in this case? I think that this is returning all of the useful information a client would need to see the blind and the nonce and then be able to recalculate them client side if they are interested in doing so.

test/wallet-http.js
tynes: Yeah you are right that any string here will allow the assertion to pass. I thought that it did a regex under the hood against the error message? I've definitely experienced it doing that before in the past, maybe we have a different version of `bmocha`?

lib/wallet/http.js
```diff
@@ -1153,6 +1502,77 @@ class HTTPOptions {
...
+   */
+
+  fromValidator(valid) {
+    assert(valid);
+
+    this.rate = valid.u64('rate');
+    this.maxFee = valid.u64('maxFee');
+    this.selection = valid.str('selection');
+    this.smart = valid.bool('smart');
+    this.account = valid.str('account');
```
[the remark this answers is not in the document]
tynes: Previous work only handles account names as strings.

https://github.com/handshake-org/hsd/blob/41ff9172ba32ce4b2af78dc11ec4045e0b37168e/lib/wallet/http.js#L299

Making this polymorphic would add some complexity, if there was a desire for bip44 account indexes being passed then I think we could come up with a solution.

lib/wallet/http.js
tynes: I separated the 2 because I thought it looked better aesthetically. Moved it back in though, not very opinionated on it.

lib/wallet/http.js
tynes: Tests for this added in e5bcf52dcb6f1a2cda8ffe32fb3e3e7a719ba2ad

tynes:
The broken CI is really annoying and it seems to be coming from Circle CI failing to install a dependency from a git repo hosted on Github. Looking on Google for a solution, there are many other people experiencing the problem. Circle CI even has a help document on how to solve the problem - https://support.circleci.com/hc/en-us/articles/360009699913-Can-t-read-from-remote-git-repository-during-npm-install

Following their document directly in 1eecc7cf631dac483cc12db83bd1b17ff23060b1, it still does not properly install from a git repo hosted on Github. My repo is public - https://github.com/tynes/hs-client. I tried running their container:

```bash
$ docker run --rm -it circleci/node:10.2 bash
```

Installing from the npm registry works fine, but using npm to install from a git repo fails due to a permission error. Switching to a more standard node image could fix the problem. Or we could merge in `hs-client` https://github.com/handshake-org/hs-client/pull/16 and publish first so we can install from an npm registry and have it just work.

lib/wallet/wallet.js
```diff
@@ -1606,15 +1607,15 @@ class Wallet extends EventEmitter {
...
+   * @param {Number|String} acct
    * @returns {MTX}
    */
 
   async makeBid(name, value, lockup, acct) {
     assert(typeof name === 'string');
     assert(Number.isSafeInteger(value) && value >= 0);
     assert(Number.isSafeInteger(lockup) && lockup >= 0);
-    assert((acct >>> 0) === acct);
+    assert((acct >>> 0) === acct || typeof acct === 'string');
```
tynes: The alternative to making `acct` polymorphic is doing a database lookup earlier to convert the human readable account name to the account index integer that this function is expecting. The line

```javascript
const addr = await this.receiveAddress(acct);
```

works on either an integer or a string, where the integer is expected to be the account index or the string is expected to be the human readable account name.

lib/wallet/http.js
tynes: `GET /wallet/:id/bid` will return all of the bids for the wallet between all of the names. `GET /wallet/:id/bid/:name` will return all of the bids for a given name. Right now, the client has a method for each endpoint

- `getBid` for `/wallet/:id/bid/`
- `getBids` for `/wallet/:id/bid/:name`

Both endpoints can return multiple bids, so the naming is not ideal. I think it could be `getBids` and `getBidsByName` but then its getting really verbose. Thoughts?

The same thing applies to the reveal endpoints.

test/wallet-http.js
tynes: Checking for this right here - https://github.com/handshake-org/hsd/pull/163/files#diff-7b92af4a1865adf03377c0804427d57dR102

lib/wallet/http.js
tynes: Decided to migrate to `getBidsByName` and `getRevealsByName` because those exist as methods on the `wallet` itself.

test/wallet-http.js
pinheadmz: I think the string is the error message `assert` outputs if the first argument is falsey.

lib/wallet/http.js
boymanjor: @tynes this is only a problem if we allow malformed data to be stored through the standard API.

lib/wallet/http.js
boymanjor: Because we are returning the `name` and `bid,` even though the request provides them, why are we not also returning the address? I think we should either return them all or assume the client has the information already.

lib/wallet/http.js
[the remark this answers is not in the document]
boymanjor: I think it is fine to keep the functionality simple for now and expect the account name.

lib/wallet/wallet.js
boymanjor: I think this current approach works fine. The issue we are solving is that validation of the RPC and HTTP endpoints, only accept strings:
- https://github.com/handshake-org/hsd/pull/163/files/96945ed56e8580ca28346850f887261a972a8102#diff-e71441ed6076bfec17d8b3e3e6a9cd8dR1601
- https://github.com/handshake-org/hsd/pull/145/files#diff-fd026b8e2fb746e2c72b16a555745ca6R1990

When the account option is not passed through the API, the `acct` parameter defaults to 0.
- https://github.com/tynes/hsd/blob/wallet-http-create/lib/wallet/wallet.js#L1550

Currently, I believe this is the only way this function would ever receive an integer as its `acct` parameter.

test/wallet-http.js
boymanjor: @tynes was this addressed. I see it is associated with an outdated commit, but the issue has not been resolved.

test/wallet-http.js
boymanjor: Should this be resolved? It is associated with an outdated commit.

test/wallet-http.js
boymanjor: Should this be resolved? It is associated with an outdated commit.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+
+      for (const ns of names)
+        items.push(ns.getJSON(height, network));
+
+      res.json(200, items);
+    });
+
+    // Wallet Name State
+    this.get('/wallet/:id/name/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
```
boymanjor: We do not need this `Validator` object anymore. The name is passed in the resource URL.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+
+        items.push(info);
+      }
+
+      return res.json(200, items);
+    });
+
+    // Wallet Auction
+    this.get('/wallet/:id/auction/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
```
boymanjor: We can get rid of this `Validator` object as well.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+      const height = this.wdb.height;
+      const network = this.network;
+
+      const names = await req.wallet.getNames();
+      const items = [];
+
+      for (const ns of names) {
+        const bids = await req.wallet.getBidsByName(ns.name);
+        const reveals = await req.wallet.getRevealsByName(ns.name);
+
```
boymanjor: nit: how do you feel about removing this new line and adding one after creating `info`.

lib/wallet/http.js
tynes: A `Validator` is still used to parse data from the resource URL in other parts of the codebase, see https://github.com/handshake-org/hsd/blob/master/lib/node/http.js#L159

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+          items.push(bid.toJSON());
+      }
+
+      res.json(200, items);
+    });
+
+    // Wallet Bids by Name
+    this.get('/wallet/:id/bid/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
```
boymanjor: The name is included in the resource URL.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+          items.push(brv.toJSON());
+      }
+
+      res.json(200, items);
+    });
+
+    // Wallet Reveals by Name
+    this.get('/wallet/:id/reveal/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
```
boymanjor: The name is included in the resource URL.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+        if (!own || brv.own)
+          items.push(brv.toJSON());
+      }
+
+      res.json(200, items);
+    });
+
+    // Name Resource
+    this.get('/wallet/:id/resource/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
```
boymanjor: We can get rid of this `Validator` object.

lib/wallet/http.js
```diff
@@ -837,6 +784,501 @@ class HTTP extends Server {
...
+        return res.json(200, resource.toJSON());
+      } catch (e) {
+        return res.json(400);
+      }
+    });
+
+    // Regenerate Nonce
+    this.get('/wallet/:id/nonce/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
```
boymanjor: The name is in the resource URL.

lib/wallet/http.js
tynes: I agree that looks better, pushed up a fix

test/wallet-http.js
tynes: @boymanjor - To test this properly, I think it needs a full node started from scratch so that its not too tightly coupled to previous tests. It currently really only tests that `bids` and `reveals` are properties on the object returned.

lib/wallet/http.js
boymanjor: For sure. I just noticed my error and deleted the rest of these comments.

lib/wallet/http.js
tynes: If this is a problem later we can always return `ns.data.toString('hex')` in the `catch` statement

lib/wallet/http.js
tynes: Added `address` as a property on the returned object

test/util/common.js
```diff
@@ -85,6 +85,33 @@ common.writeTX = function writeTX(name, tx, view) {
...
+  });
+};
+
+common.forValue = async function(obj, key, val, timeout = 30000) {
+  assert(typeof obj === 'object');
+  assert(typeof key === 'string');
+
+  const ms = 10;
+  let interval = null;
+  let count = 0;
```
boymanjor: nit: can we add a newline after this line?

lib/wallet/http.js
```diff
@@ -837,6 +784,502 @@ class HTTP extends Server {
...
+
+    // Wallet Auction
+    this.get('/wallet/:id/auction/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
+      const height = this.wdb.height;
+      const network = this.network;
+
+      assert(name, 'Must pass name.');
```
boymanjor: nit: let's move these assertions up to before we create `height` and `network.

lib/wallet/http.js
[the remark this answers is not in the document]
boymanjor: @tynes this is good for now.

test/wallet-http-test.js
```diff
@@ -0,0 +1,1078 @@
...
+    const height = 5;
+
+    const {address} = await wallet2.createAddress('default');
+    await mineBlocks(height, address);
+
+    const accountInfo = await wallet2.getAccount('default');
+    assert.equal(accountInfo.balance.coin, height);
+  });
+
+  it('should have no name state indexed', async () => {
```
boymanjor: nit: could we add "initially" to the end of this, i.e. 'should have no name state indexed initially'.

test/wallet-http-test.js
```diff
@@ -0,0 +1,1078 @@
...
+    const names = await wallet.getNames();
+
+    assert.strictEqual(names.length, 0);
+  });
+
+  it('should allow covenants with create tx', async () => {
```
boymanjor: What happens if we try to sort the outputs and while including a covenant?

test/wallet-http-test.js
```diff
@@ -0,0 +1,1078 @@
...
+    const output = openOutput(name, address);
+
+    const mtx = new MTX();
+    mtx.outputs.push(output);
+
+    const tx = await wallet.createTX(mtx);
+    assert.equal(tx.outputs[0].covenant.type, types.OPEN);
+  });
+
+  it('should allow covenants with send tx', async () => {
```
boymanjor: Ditto for sorting outputs with a covenant.

test/wallet-http-test.js
tynes: `mtx.sortMembers` has logic specific for covenants and making sure that the input and output stays linked. This codepath is followed every time that any transaction is created through the `send<CovenantType>` codepath. See here:

https://github.com/handshake-org/hsd/blob/57b5b491699d5438c53ddaeacd8a0e0122306853/lib/primitives/mtx.js#L1189

I don't think its a problem allowing the `sort` option, especially because it must explicitly be set to `false` for the sorting to not happen, which none of the current code paths currently do

https://github.com/handshake-org/hsd/blob/7b6ee0b8e973c3c77f52ca70dbef860d1c26a949/lib/wallet/wallet.js#L3198-L3199

test/wallet-http-test.js
tynes: See https://github.com/handshake-org/hsd/pull/163#discussion_r288860220

test/wallet-http.js
tynes: @pinheadmz - tightened up testing here, using state from a previous test and making clear assertions on what the state should be. 👍

tynes:
Will rebase this into a clean git history once it looks good to go. Requires https://github.com/handshake-org/hs-client/pull/16 to be merged and published to `npm` before this can be merged

tynes:
@boymanjor Cleaned up the commits for this, the original branch can be found here:
https://github.com/tynes/hsd/tree/wallet-http-backup
Ran a `git diff` between the branches to make sure that no functionality was left behind.

If we can merge/publish https://github.com/handshake-org/hs-client/pull/16 then I can remove b37d871198d6ae17dffd52a7ecfc4d67864e0666, that is required before this can be merged.

boymanjor:
> @boymanjor Cleaned up the commits for this, the original branch can be found here:
> https://github.com/tynes/hsd/tree/wallet-http-backup
> Ran a `git diff` between the branches to make sure that no functionality was left behind.
> 
> If we can merge/publish [handshake-org/hs-client#16](https://github.com/handshake-org/hs-client/pull/16) then I can remove [b37d871](https://github.com/handshake-org/hsd/commit/b37d871198d6ae17dffd52a7ecfc4d67864e0666), that is required before this can be merged.

We've decided to use the raw HTTP client from `hs-client` to test the new functionality. When @chjj updates the `hs-client` npm package, we can revert this change and rely on the new, client methods.

PR #164, opened 2019-05-01
lint: convert eslint to bslint for ci

kilpatty:
Currently we aren't able to run linting on Handshake without a globally
installed eslint. This will install bslint which should cover all of the
linting needs, as well as removes eslint install from the ci script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
       - run: npm install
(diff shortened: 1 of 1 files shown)


kilpatty:
Per https://github.com/handshake-org/hsd/pull/132

tynes:
@kilpatty poke to remove `bslint` from the `package.json` and install it as part of the CI

chjj:
There's a reason eslint isn't present in any of our projects' package.json: eslint pulls in a _ton_ of garbage. Requiring it for every project is insanity. I think the assumption should be that it is globally installed.

kilpatty:
Sorry about the delay on switching this, thanks for the reminder @tynes

PR #165, opened 2019-05-02
wallet: add locktime and sort params for HTTP send/create

tynes:
This exposes the parameters `locktime` and `sort` through the REST API for the endpoint.

`locktime` will set the nLocktime field on the transaction and `sort` will sort the inputs/outputs using [bip 69](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki)

- `POST /wallet/:id/send`
- `POST /wallet/:id/create`

It also includes tests

### DIFF
--- a/test/wallet-http.js
+++ b/test/wallet-http.js
@@ -1,0 +1,138 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+/* eslint no-return-assign: "off" */
+
+'use strict';
+
+const {NodeClient, WalletClient} = require('hs-client');
+const Network = require('../lib/protocol/network');
+const FullNode = require('../lib/node/fullnode');
+const MTX = require('../lib/primitives/mtx');
+
+const network = Network.get('regtest');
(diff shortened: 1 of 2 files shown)


test/wallet-http.js
```diff
@@ -0,0 +1,138 @@
...
+    for (const [i, output] of outputs.entries()) {
+      assert.equal(tx.outputs[i].value, output.value);
+      assert.equal(tx.outputs[i].address.toString(network), output.address);
+    }
+
+    const mtx = MTX.fromJSON(tx);
+    mtx.sortMembers();
+
+    // the order changes after sorting
+    assert.equal(tx.outputs[0].value === mtx.outputs[0].value, false);
```
tynes: Change this to `assert.ok(!...)`

boymanjor: You could also change this to:
```
assert.ok(tx.outputs[0].value !== mtx.outputs[0].value);

tynes: That is cleaner, just pushed up a fix

PR #168, opened 2019-05-05
Node http nameinfo

wi-ski:


package.json
```diff
@@ -48,7 +48,7 @@
     "bval": "~0.1.6",
     "bweb": "~0.1.8",
     "goosig": "~0.1.0",
-    "hs-client": "~0.0.6",
+    "hs-client": "github:wi-ski/hs-client#node-http-routes",
```
wi-ski: This should never make it to master.

wi-ski: I put this in here to make CI pass, but CI is still failing for some reason. :cry:

wi-ski:
Hey @tynes we spoke about landing these changes are separate PRs. But there were some consistency/non-determinism issues I thought we should chat about first!

package.json
tynes: For now you should use the HTTP methods on the client object, like `client.get`

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+              reserved: false,
+              start: 0,
+              week: 0
+            }
+        });
+        const open = await wclient.execute('sendopen', [NAME0]);
+        assert(open);
+      });
+      it('It should start an auction on the 2nd day', async () => {
+        // Question: This test passes non-deterministically. Why?
```
tynes: There is a race condition with mining blocks, a hacky way to fix the problem is using a `sleep` function, the better way to do it is to listen to events and `await` for the event you are listening for, otherwise timeout after some period

chjj: It could be the overhead of the client interacting with the server, but I think it could also be a race condition between the node and the wallet. The older versions of bcoin's wallet shared the same locks with the node, when we decided to separate everything out, we got more modularity, but we ended up in race condition hell. :(

Like @tynes said, the best thing to do is probably wait a couple hundred ms unless someone wants to write some utils that await wallet/node events.

The other possibility is to switch back to having the wallet share the same locks with the node which might mitigate some of this. Something I've wanted for a while.

chjj:
Good work. This is probably a good start to adding names to the rest API. Will investigate the other race conditions tomorrow.

tynes:
@wi-ski - I have an open PR here for `GET /info/name/:name` https://github.com/handshake-org/hsd/pull/162

I'd happily close it if you got this PR cleaned up

wi-ski:
@tynes - Sitting down to take a pass at this now. Quick question: You ok with these changes landing in a single PR? :angel:

wi-ski:
@tynes @chjj - Hey dudes. I took a stab at getting an event-listening setup working but encountered some errors using the `.on(...)` method of the FullNode. I'm at a loss.

Additionally, I tried to implement various forms of a sleep-_esque_ test util and found that it actually made the test pass **less** consistently. This has been a battle.

I can set aside some to see if there's a better way to listen for events. @tynes - If you feel like any of this usable, please have at it.

:thinking: I really was stumped, listening for events would just mysteriously error _sometimes_ \o.o/.

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+
+const { NodeClient, WalletClient } = require('hs-client');
+
+const nclient = new NodeClient({
+  port: network.rpcPort, // Todo: Make it so this doesnt bind to any port
+  apiKey: 'foo'
+});
+
+const wclient = new WalletClient({
+  port: network.walletPort, // Todo: Make it so this doesnt bind to any port
```
wi-ski:
```suggestion
  port: network.walletPort,
```

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+const assert = require('bsert');
+const Network = require('../lib/protocol/network');
+const FullNode = require('../lib/node/fullnode');
+const rules = require('../lib/covenants/rules');
+const network = Network.get('regtest');
+
+const { NodeClient, WalletClient } = require('hs-client');
+
+const nclient = new NodeClient({
+  port: network.rpcPort, // Todo: Make it so this doesnt bind to any port
```
wi-ski:
```suggestion
  port: network.rpcPort,
```

wi-ski:
Oh oh, I didn't even mention - I cleaned up the PR in that I:
1. Swapped out the "named" methods that were being used on the update hs-client (using `.get(...)` now)
2. Restored package.json

Is there a _recommended_ way of "sleeping" that either of you have had success with btw?

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+});
+
+const wclient = new WalletClient({
+  port: network.walletPort,
+  apiKey: 'foo'
+});
+
+describe('Node http', function() {
+  this.timeout(5000);
+  let NAME0;
```
tynes:
```suggestion
  let NAME0, NAME1;
  let node, miner, chain;
```

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+      it('It should return the name', async () => {
+        const nameHash = rules.hashName(NAME0);
+        const { name } = await nclient.get(`/name/hash/${nameHash.toString('hex')}`);
+        assert.equal(name, NAME0);
+      });
+    });
+  });
+
+  describe('getNameResource', () => {
+    const zonefile = { compat: false, version: 0, ttl: 172800, ns: ['ns1.example.com.@1.2.3.4'] };
```
tynes: I'm not sure if its a good idea to call this data `zonefile`, its decoded namestate data. Maybe something like `records`

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+    it('It should return null when an auction has not been initiated', async () => {
+      const resource = await nclient.get(`/resource/name/${NAME0}`);
+      assert.equal(resource, null);
+    });
+
+    describe('When an auction has been initiated', () => {
+      beforeEach(async () => {
+        await mineBlocks(250);
```
tynes: This section of code is really dense and I don't think it needs to be in a `beforeEach` block, there is only 1 `it` block, would you want another 250 blocks to be mined if another `it` block was added?

test/node-http-test.js
```diff
@@ -0,0 +1,219 @@
...
+      it('It should return the name\'s proof', async () => {
+        const proof = await nclient.get(`/proof/name/${NAME0}`);
+        assert.equal(proof.proof.type, 'TYPE_EXISTS');
+        assert.equal(proof.name, NAME0);
+      });
+    });
+  });
+  describe('grindName', () => {
+    it('It should grind a name', async () => {
+      const { name } = await nclient.get('/grind', { size: 10 });
```
tynes: Assert that the size is correct and that it is a valid name

tynes:
A simple sleep function:

```javascript
async function sleep(time) {
  return new Promise(resolve => setTimeout(resolve, time));
}
```

tynes:
More useful helper functions by @braydonf 

https://github.com/bcoin-org/bcoin/pull/748/files#diff-e169f8c8ca803ee76312e094ba6f17cbR105

```javascript
common.event = async function event(obj, name) {
  return new Promise((resolve) => {
    obj.once(name, resolve);
  });
};
```

https://github.com/bcoin-org/bcoin/pull/758/files#diff-e169f8c8ca803ee76312e094ba6f17cbR105

```javascript
common.forValue = async function(obj, key, val, timeout = 30000) {
  assert(typeof obj === 'object');
  assert(typeof key === 'string');

  const ms = 10;
  let interval = null;
  let count = 0;
  return new Promise((resolve, reject) => {
    interval = setInterval(() => {
      if (obj[key] === val) {
        clearInterval(interval);
        resolve();
      } else if (count * ms >= timeout) {
        clearInterval(interval);
        reject(new Error('Timeout waiting for value.'));
      }
      count += 1;
    }, ms);
  });
};
```

boymanjor:
> @tynes - Sitting down to take a pass at this now. Quick question: You ok with these changes landing in a single PR? 👼

@tynes @wi-ski I think this is fine as one, name related PR.

wi-ski:
Ok - extended the timeout for the test allowed for these to pass. Not ideal.

tynes:
Can you mine less blocks? We don't want to dramatically extend the length of the test suite with just one set of tests

wi-ski:
@tynes - Updates:

1. Using the mentioned event listening util.
    - Test pass seemingly consistently.
2. Mining fewer blocks.

Recent CI passed with longer test times requiring a little under 10 seconds to complete.

Hit me with your feedback when you can.

wi-ski:
#bump. @tynes, anything I can do for this PR to help get it over the line?

lib/node/http.js
```diff
@@ -383,6 +385,108 @@ class HTTP extends Server {
...
+
+      if (!hash)
+        throw new Error('Invalid hash.');
+
+      const ns = await this.chain.db.getNameState(hash);
+
+      if (!ns)
+        return res.json(404);
+
+      return res.json(200, { name: ns.name.toString('binary') });
```
tynes: `ns.getJSON` should be the 2nd argument to `res.json` like this:

See implementation here:
https://github.com/handshake-org/hsd/blob/696dfc4601ef5e8cd954cde727142f39c965fcf6/lib/wallet/rpc.js#L1774

https://github.com/handshake-org/hsd/blob/master/lib/covenants/namestate.js#L651

lib/node/http.js
```diff
@@ -383,6 +385,108 @@ class HTTP extends Server {
...
+
+      const nameHash = rules.hashName(name);
+      const ns = await this.chain.db.getNameState(nameHash);
+
+      if (!ns || ns.data.length === 0)
+        return null;
+
+      const resource = Resource.decode(ns.data);
+
+      return res.json(200, resource.toJSON());
```
tynes: Pass in the name to `getJSON`, see here:
https://github.com/handshake-org/hsd/blob/696dfc4601ef5e8cd954cde727142f39c965fcf6/lib/dns/resource.js#L1309

lib/node/http.js
```diff
@@ -383,6 +385,108 @@ class HTTP extends Server {
...
+        throw new Error('Invalid name.');
+
+      const hash = this.chain.tip.hash;
+      const height = this.chain.tip.height;
+      const root = this.chain.tip.treeRoot;
+      const key = rules.hashName(name);
+      const proof = await this.chain.db.prove(root, key);
+
+      return res.json(200, {
+        hash: hash.toString('hex'),
```
tynes: Its not particularly clear what `hash` is in this context, what about `tip`? It is used this way for node `GET /`
https://github.com/handshake-org/hsd/blob/41ff9172ba32ce4b2af78dc11ec4045e0b37168e/lib/node/http.js#L126

I think `key` should be renamed to `nameHash`

tynes:
@wi-ski thanks for the bump. Just left a few comments for you

test/node-http-test.js
```diff
@@ -0,0 +1,215 @@
...
+        assert(sendBid);
+        const { stats: { blocksUntilReveal } } = await wclient.execute('getauctioninfo', [NAME0]);
+        await mineBlocks(blocksUntilReveal);
+        await wclient.execute('sendreveal', [NAME0]);
+        const { stats: { blocksUntilClose } } = await wclient.execute('getauctioninfo', [NAME0]);
+        await mineBlocks(blocksUntilClose);
+        await wclient.execute('sendupdate', [NAME0, records]);
+        await mineBlocks(1);
+        const resource = await nclient.get(`/resource/name/${NAME0}`);
+        assert.deepEqual(resource, { ...records, name: NAME0, version: 0 });
```
wi-ski: @tynes - :thinking: You ok with this? Or do you have a better way?

wi-ski:
Swinging back on this.  Ill take a look into CI 🤔

wi-ski:
@tynes #bump

lib/node/http.js
```diff
@@ -432,6 +434,131 @@ class HTTP extends Server {
...
+
+      if (!nameHash)
+        throw new Error('Invalid hash.');
+
+      const tip = this.chain.tip.hash;
+      const height = this.chain.tip.height;
+      const root = this.chain.tip.treeRoot;
+      const proof = await this.chain.db.prove(root, nameHash);
+
+      return res.json(200, {
```
wi-ski: @tynes :eyes: We ok with this response shape? Mirrors the behavior of `getnameproof` but omits `.name`

test/util/common.js
```diff
@@ -113,6 +117,20 @@ common.forValue = async function(obj, key, val, timeout = 30000) {
   });
 };
 
+common.constructBlockMiner = function (node, nclient) {
```
wi-ski: I can _feel_ others thinking this is less than optimal - the whole "pass me a full node and api client" signature is wonky 🤔

test/wallet-http-test.js
```diff
@@ -79,6 +79,7 @@ describe('Wallet HTTP', function() {
   });
 
   beforeEach(async () => {
+    mineBlocks = common.constructBlockMiner(node, nclient);
```
tynes: We don't need to do this before every `it` block do we? Its always the same node and the same client

test/util/common.js
tynes: I can't think of a better way besides using a closure like this, can you?

test/util/common.js
```diff
@@ -91,6 +91,10 @@ common.event = async function event(obj, name) {
   });
 };
 
+common.sleep = function sleep(time) {
```
tynes: Good call on moving `sleep` here

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+
+      if (!ns || ns.data.length === 0)
+        return null;
+
+      const resource = Resource.decode(ns.data);
+
+      return res.json(200, resource.getJSON(name));
+    });
+
+    this.get('/proof/name/:name', async (req, res) => {
```
tynes: Now that I think about it, there are different kinds of proofs. There is an Urkle Tree Proof, which authenticates the namestate and then there is a Merkle Proof for transactions in a block. Ideally we want to be able to support both of them and if we use `/proof/name/:name` then we can use `/proof/tx/:txid`, but I think that would only work if transaction indexing is turned on. So using `/proof/name/:name` works.

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+      const nameHash = rules.hashName(name);
+      const proof = await this.chain.db.prove(root, nameHash);
+
+      return res.json(200, {
+        hash: tip.toString('hex'),
+        height: height,
+        root: root.toString('hex'),
+        name: name,
+        key: nameHash.toString('hex'),
+        proof: proof.toJSON()
```
tynes: The shape of `proof.toJSON` looks like:

```
  toJSON() {
    return {
      type: typesByVal[this.type],
      depth: this.depth,
      nodes: this.nodes.map(node => node.toJSON()),
      prefix: this.prefix ? this.prefix.toString() : undefined,
      left: this.left ? this.left.toString('hex') : undefined,
      right: this.right ? this.right.toString('hex') : undefined,
      key: this.key ? this.key.toString('hex') : undefined,
      hash: this.hash ? this.hash.toString('hex') : undefined,
      value: this.value ? this.value.toString('hex') : undefined
    };
  }
```

We should try to deduplicate any fields that correspond between the two. It feels a little bit cleaner to me to follow the style in other endpoints where the only object returned is a `primitive.toJSON`.

We want to be 100% sure that we can easily verify the proof client side.

test/util/common.js
wi-ski: Nope!

test/wallet-http-test.js
wi-ski: Nope!

lib/node/http.js
wi-ski: :thinking: I dont entirely follow.

This shape mirrors the response of the RPC method:
https://github.com/handshake-org/hsd/blob/master/lib/node/rpc.js#L2375

Would you prefer something like:
```js
    return {
      hash: hash.toString('hex'),
      height: height,
      root: root.toString('hex'),
      name: name,
      key: key.toString('hex'),
      ...proof.toJSON() // <- This would achieve your de-duping.
    };
```

It is definitely cleaner to return _something_.toJSON() - but to confirm, we have other handlers in this file that construct their response "in the handler" like: https://github.com/handshake-org/hsd/blob/master/lib/node/http.js#L354

Whichever we decide - Illl do the same for: https://github.com/handshake-org/hsd/pull/168/files/231e30eef2c442d4914c57cddd840a03624abac9#diff-bbe506ab5582252a30d6872d35fd8ce6R540

wi-ski: Current question(s) I _believe_ Im trying to answer:
a. Do we want to return the shape of `proof.toJSON()` with `height` & `root` & `name` included?
or
b. _Something_ else that results in something akin to:
```js
      return res.json(200, primitive.toJSON()); // Assuming just proof.toJSON?
```

tynes: > but to confirm, we have other handlers in this file that construct their response "in the handler" 

This one was done that way because there was no `filter.JSON` and the filter object lives in a different repo and its hard to coordinate PRs in multiple repos

tynes: Between a and b, I think something in between is good, it would be nice to have the height and the root along with the rest of `proof.toJSON`. What we really need to be 100% sure of is that we have everything that we need client side to verify the proof, which I believe `proof.toJSON` gives.

test/node-http-test.js
```diff
@@ -0,0 +1,283 @@
...
+        const { stats: { blocksUntilReveal } } = await wclient.execute('getauctioninfo', [NAME0]);
+
+        await mineBlocks(blocksUntilReveal, cbAddress);
+
+        await wclient.execute('sendreveal', [NAME0]);
+        const { stats: { blocksUntilClose } } = await wclient.execute('getauctioninfo', [NAME0]);
+
+        await mineBlocks(blocksUntilClose, cbAddress);
+
+        await wclient.execute('sendupdate', [NAME0, { compat: false, version: 0, ttl: 172800, ns: ['ns1.example.com.@1.2.3.4'] }]);
```
tynes: This line is a little too long, can you turn the record into its own object? It would also be nice to assert that getting the namestate reflects the record.

test/node-http-test.js
```diff
@@ -0,0 +1,283 @@
...
+        await wclient.execute('sendupdate', [NAME0, records]);
+        await mineBlocks(1, cbAddress);
+
+        const resource = await nclient.get(`/resource/name/${NAME0}`);
+        assert.deepEqual(resource, Object.assign(records, { name: NAME0, version: 0 }));
+      });
+    });
+  });
+
+  describe('getNameProof', () => {
```
tynes: Do you think we could dedupe `getnameproof` and `getnameproofbyhash` with a loop?

wi-ski: :thinking: Yess - we could. I don't think it would improve anything however. Do you feel strongly about this?

wi-ski: There is solid overlap - but there's only "2 things" to abstract, this feels like a situation I'd want to hold off on abstracting into a loop. Your call. But I wanted to voice the thought :ok:

wi-ski:
@tynes:
a. Addressed line length issues.
b. The shape of the "proof" routes' response - I went with the:
```js
      return res.json(200, {
        hash: tip.toString('hex'),
        height: height,
        root: root.toString('hex'),
        key: nameHash.toString('hex'),
        name: ns ? ns.name.toString() : null,
        ...proof.toJSON()
      });
```
approach. I think this achieves what I feel like you described as acceptable. It would be awesome to chat about that in more detail if you have any issues with it.
c. Loops in the test. I left a not about wanting to avoid this, it feels premature. If you feel otherwise after me bringing that up, let me know!

lib/node/http.js
wi-ski: I added the:
```js
    return {
      hash: hash.toString('hex'),
      height: height,
      root: root.toString('hex'),
      name: name,
      key: key.toString('hex'),
      ...proof.toJSON()
    };
```
:point_up: This feels ok - let me know if you feel otherwise!

wi-ski:
@tynes bump <3

wi-ski:
@tynes - Touching base on this. How are we looking?

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
 
       res.json(200, { success: true });
     });
+
+    this.get('/name/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
+      if (!name || !rules.verifyName(name))
+        throw new Error('Invalid parameter.');
```
tynes: Nit: `'Invalid name.'`

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+    this.get('/resource/name/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
+      if (!name || !rules.verifyName(name))
+        throw new Error('Invalid name.');
+
+      const nameHash = rules.hashName(name);
+      const ns = await this.chain.db.getNameState(nameHash);
+
```
tynes: We don't want to return `null` here, we want to return a `res.json(404)`

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+      const hash = valid.bhash('hash');
+
+      if (!hash)
+        throw new Error('Invalid hash.');
+
+      const ns = await this.chain.db.getNameState(hash);
+      const height = this.chain.tip.height;
+      const network = this.network;
+
+      if (!ns)
```
tynes: Also need to check if `ns.data.length === 0`

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+      const nameHash = rules.hashName(name);
+      const proof = await this.chain.db.prove(root, nameHash);
+
+      return res.json(200, {
+        hash: tip.toString('hex'),
+        height: height,
+        root: root.toString('hex'),
+        name: name,
+        key: nameHash.toString('hex'),
+        ...proof.toJSON()
```
tynes: I don't love this solution, I think it would be more ideal to pass some data into `proof.toJSON` but then that would couple the proof to a particular blockchain implementation when it could be agnostic. The proof is defined in the `urkel` library, not `hsd`.

Maybe something like:

```
{
  proof: proof.toJSON(),
  chain: {
    height: height
    root: root.toString('hex')
    name: name
  }
```

We want to be able to support passing in an arbitrary `height` (create proofs at a certain height) eventually, that feels like it could be its own endpoint `GET /proof/name/:name/height/:height` or by adding a query param `GET /proof/name/:name?height=height`

Pinging @boymanjor @pinheadmz for opinions

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
...
+      });
+    });
+
+    this.get('/proof/hash/:nameHash', async (req, res) => {
```
tynes: Nit: `nameHash` -> `hash`

lib/node/http.js
```diff
@@ -432,6 +434,133 @@ class HTTP extends Server {
 
       res.json(200, { success: true });
     });
+
```
tynes: Each of these endpoints could use a comment describing their functionality

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+      memory: true,
+      apiKey: 'foo',
+      network: 'regtest',
+      workers: true,
+      plugins: [require('../lib/wallet/plugin')]
+    });
+
+    node.on('connect', (entry, block) => {
+      const blockHash = block.hash().toString('hex');
+      witnessedBlocks[blockHash] = blockHash;
```
tynes: Why not use a `Set` if you are using an object for membership testing? There is also a `BufferSet` in the `buffer-map` library

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+const wallet = wclient.wallet('primary');
+
+describe('Node http', function() {
+  this.timeout(10000);
+  const witnessedBlocks = {};
+  let NAME0, NAME1;
+  let node, mineBlocks, cbAddress;
+
+  beforeEach(async () => {
+    assert.equal(network.names.auctionStart, 0);
```
tynes: What is the point of this assert?

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+        const nameInfo = await nclient.get(`/name/${NAME0}`);
+        assert.deepEqual(nameInfo, {
+          info: null,
+            start: {
+              reserved: false,
+              start: 0,
+              week: 0
+            }
+        });
+      });
```
tynes: Nit: newline between the `it` blocks

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+          }
+        });
+
+        await mineBlocks(10, cbAddress);
+
+        const open = await wclient.execute('sendopen', [NAME0]);
+        assert(open);
+      });
+      it('It should start an auction on the 2nd day', async () => {
+        await mineBlocks(40, cbAddress);
```
tynes: `40` is a magic number here, it is not immediately obvious how it is tied to the 2nd day

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+        await wclient.execute('sendreveal', [NAME0]);
+        const { stats: { blocksUntilClose } } = await wclient.execute('getauctioninfo', [NAME0]);
+
+        await mineBlocks(blocksUntilClose, cbAddress);
+
+        await wclient.execute('sendupdate', [NAME0, records]);
+        await mineBlocks(1, cbAddress);
+
+        const resource = await nclient.get(`/resource/name/${NAME0}`);
+        assert.deepEqual(resource, Object.assign(records, { name: NAME0, version: 0 }));
```
tynes: I'd vote for not doing a `deepEqual` on the result of an `Object.assign` and instead just comparing the object properties one by one. Then this test will not break if additional information is added.

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+      });
+    });
+  });
+
+  describe('getNameProof', () => {
+    it('It should return a proof type of TYPE_DEADEND when an auction has not been initiated', async () => {
+      const proof = await nclient.get(`/proof/name/${NAME0}`);
+      assert.equal(proof.type, 'TYPE_DEADEND');
```
tynes: Nit: assertion against the enums defined here https://github.com/handshake-org/urkel/blob/fbc93f6fbfeabaf15c1bb36b3f5f7d64ee4c240e/lib/radix/proof.js#L664

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+      assert.equal(proof.name, NAME0);
+    });
+
+    describe('When an auction has been initiated', () => {
```
tynes: Any reason to nest this `describe`?

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+  });
+
+  describe('getNameByHash', () => {
+    it('It should return null when an auction has not been initiated', async () => {
+      const nameHash = rules.hashName(NAME0);
+      const name = await nclient.get(`/resource/hash/${nameHash.toString('hex')}`);
+      assert.equal(name, null);
+    });
+
+    describe('When an auction has been initiated', () => {
```
tynes: Any reason to nest `describe` blocks?

test/node-http-test.js
```diff
@@ -0,0 +1,285 @@
...
+        await mineBlocks(1, cbAddress);
+
+        const proof = await nclient.get(`/proof/name/${NAME0}`);
+        assert.equal(proof.type, 'TYPE_EXISTS');
+        assert.equal(proof.name, NAME0);
+      });
+    });
+  });
+
+  describe('getNameProofByHash', () => {
```
tynes: Nit: `Get Name Proof By Hash`

lib/node/http.js
pinheadmz: I don't want to opinionate too much since I'm new to this thread but... To compare, `rpc gettxoutproof` just returns a raw hex string, and has a matching verifier `rpc verifytxoutproof`. I think getting a proof is useless (?) without a matching endpoint to verify it, and that should perhaps dictate the output format of the `get proof` endpoint? 

Briefly checking this:

https://github.com/handshake-org/urkel/blob/c29466104b25f85ec2ef7a5025af625a6df21eda/lib/radix/proof.js#L256

`  verify(root, key, hash, bits) {...`

... I would expect a proof-getter to return the proof (either JSON or serialized hex) along with the current (by default) tree root and perhaps the `key`, which is just the name hash, but the user might not have the hash if they are using this endpoint.

PR #173, opened 2019-05-14
node/http: return txn hex

tynes:
Add the hash of the database transaction to the
result of `GET /` for the node HTTP server. This
could be useful in conjunction with the proof API,
since `.chain.treeRoot` is lagging (only is committed
to disk every so often).

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -125,13 +125,14 @@ class HTTP extends Server {
           height: this.chain.height,
           tip: this.chain.tip.hash.toString('hex'),
           treeRoot: this.chain.tip.treeRoot.toString('hex'),
-          progress: this.chain.getProgress(),
+          txn: this.chain.db.txn.rootHash().toString('hex'),
           state: {
             tx: this.chain.db.state.tx,
             coin: this.chain.db.state.coin,
             value: this.chain.db.state.value,
             burned: this.chain.db.state.burned
-          }
+          },
+          progress: this.chain.getProgress()
         },
         pool: {
           host: addr.host,


lib/node/http.js
```diff
@@ -125,6 +125,7 @@ class HTTP extends Server {
           height: this.chain.height,
           tip: this.chain.tip.hash.toString('hex'),
           treeRoot: this.chain.tip.treeRoot.toString('hex'),
```
tynes: This could influence people towards using the chain in an unsafe way. It could be useful for block explorers though.
It could also be useful in conjunction with the `getnameproof` RPC

lib/node/http.js
```diff
@@ -125,6 +125,7 @@ class HTTP extends Server {
...
+          txn: this.chain.db.txn.rootHash().toString('hex'),
```
tynes: What is a better name for this when it is exposed via the API? It is called `txn` internally, like a database transaction, but externally that may not make the most sense because it could be confused with an exchange of value transaction.

tynes:
Rebased on master. After thinking about this some more, I don't think its particularly unsafe to expose this. People are familiar enough with "6 confirmations before my deposit on Coinbase goes through".

tynes:
Some additional potential names for exposing `chain.db.txn`  via the API:

- `currentTreeRoot`
- `nextTreeRoot`
- `treeState`

pinheadmz:
Hi there almost 2-year-old pull request! @tynes what are your thoughts on this one now that its had LOTS of time to sit? My first impression is it's not needed, but what's the use case? The `txn` isn't used to Urkel-prove namestate to light clients and isn't committed to a block header until the actual tree update.

PR #175, opened 2019-05-14
node/http: add next interval

tynes:
Adds 2 new properties to Node `GET /`

- `nextTreeRootHeight` - blockheight at which the tree will be committed
- `blocksUntilNextTreeRoot` - number of blocks until the next time the tree is committed

This is useful for applications building on top of `hsd` that need to know the next time the tree will be committed. It also includes tests for off by one errors.

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -130,7 +130,8 @@ class HTTP extends Server {
         network: this.network.type,
         chain: {
           height: this.chain.height,
-          nextInterval: this.chain.height + blocksUntilNextInterval,
+          nextIntervalHeight: this.chain.height + blocksUntilNextInterval,
+          blocksUntilNextInterval: blocksUntilNextInterval,
           tip: this.chain.tip.hash.toString('hex'),
           treeRoot: this.chain.tip.treeRoot.toString('hex'),
           progress: this.chain.getProgress(),
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -121,11 +121,16 @@ class HTTP extends Server {
       if (!addr)
         addr = this.pool.hosts.address;
 
+      const treeInterval = this.network.names.treeInterval;
+      const blocksUntilNextInterval =
+        treeInterval - (this.chain.height % treeInterval);
+
       res.json(200, {
         version: pkg.version,
         network: this.network.type,
         chain: {
           height: this.chain.height,
+          nextInterval: this.chain.height + blocksUntilNextInterval,
           tip: this.chain.tip.hash.toString('hex'),
           treeRoot: this.chain.tip.treeRoot.toString('hex'),
           progress: this.chain.getProgress(),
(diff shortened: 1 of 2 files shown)


kilpatty:
Good catch, this is actually super helpful for HNScan!

tynes:
I wasn't sure what would be more useful, `blocksUntilNextInterval` or `nextInterval`.
Ended up going with `nextInterval` because its shorted and you can calculate `blocksUntilNextInterval` locally

kilpatty:
No test for OBOE??

Also thoughts about adding this into get name resource? I ran into some issues locally where seeing when a record that I added to a name would update is not explicitly clear. I wonder if we should think about adding this into the "info" part of a name when we pull that.

lib/node/http.js
```diff
@@ -118,11 +118,16 @@ class HTTP extends Server {
       if (!addr)
         addr = this.pool.hosts.address;
 
+      const treeInterval = this.network.names.treeInterval;
+      const blocksUntilNextInterval =
```
tynes: Add a relative value to match `ns.stats`, it includes both an absolute height and a relative height.

What to call this value? `blocksUntilNextInterval`?

tynes: - `nextTreeCommitment`?

tynes:
@kilpatty 

> No test for OBOE??

Added test coverage for an off by one error

> Also thoughts about adding this into get name resource?

We could definitely add this here, I think `ns.getJSON` is a good place for it. Out of scope for this PR though

lib/node/http.js
tynes: Ended up going with `nextIntervalHeight` and `blocksUntilNextInterval`

boymanjor: @tynes how do you feel about `nextTreeRootHeight` and `blocksUntilNextTreeRoot` it is the same length but matches the `treeRoot` attribute already returned.

test/util/common.js
```diff
@@ -85,6 +85,33 @@ common.writeTX = function writeTX(name, tx, view) {
...
+  });
+};
+
+common.forValue = async function(obj, key, val, timeout = 30000) {
+  assert(typeof obj === 'object');
+  assert(typeof key === 'string');
+
+  const ms = 10;
+  let interval = null;
+  let count = 0;
```
boymanjor: nit: can we add a newline between the last variable assignment and the return statement?

test/util/common.js
```diff
@@ -85,6 +85,33 @@ common.writeTX = function writeTX(name, tx, view) {
   common.writeFile(`${name}-undo.raw`, undoRaw);
 };
 
+common.event = async function event(obj, name) {
```
boymanjor: Should we remove this function? It isn't being called.

lib/node/http.js
tynes: I like calling it those, makes the API more uniform. Just pushed up the change

tynes:
@boymanjor Resolved issues, pushed up new code

pinheadmz:
@tynes I think this is a useful datum to return here, do you still wanna finish? I could take it over if you like.

Thoughts:
- Probably don't need the privatekey/wallet stuff in the test, lets remove anything we don't really need
- I think we could just add one value: `nextTreeUpdateHeight` with the height of the next update.

PR #177, opened 2019-05-15
wallet: handle multiple conflicting inputs

tynes:
Handle double spend with multiple inputs. Port of
https://github.com/bcoin-org/bcoin/commit/eba7bbeddc3f22c69ee249bb4f87e0dc9bec1bf9


kilpatty:
Is it possible this handles the problem you were dealing with in relation to the transfers? I recall you were getting an error that mentioned conflicting inputs?

PR #180, opened 2019-05-17
wallet: add locks for create code paths

tynes:
There are many different types of transactions regarding the auction system. Each covenant type has its own explicit code path for creation. Currently, only the `send` functionality has a lock around it which means that there is *no* lock being used for code that wants to create the transaction but not sign and/or broadcast it. This pull request adds a new method for each create transaction type method that wraps the actual method with a lock.

Since there is no enforced concept of private methods in Javascript, this makes each of the create transaction methods safe for external consumption. These functions were only being called from within the wallet before, but in an attempt to better support watch only wallets, the create transaction methods would like to be consumed from outside of the wallet itself.

An example:

```javascript

// original createBid functionality here
_createBid() {}


createBid() {
  // create lock
  // call _createBid
  // unlock
}
```

The above pattern was followed for each of the create methods. The send methods call the create methods and there is already a lock around the send methods (they follow a similar pattern as above), they were updated to directly use the methods with the underscores.

There are currently two pull requests that could use this functionality
- https://github.com/handshake-org/hsd/pull/163
- https://github.com/handshake-org/hsd/pull/145

### DIFF
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -2763,32 +2763,52 @@ class Wallet extends EventEmitter {
   }
 
   /**
-   * Create and finalize a cancel MTX.
+   * Create and finalize a cancel
+   * MTX without a lock.
    * @param {String} name
    * @param {Object} options
    * @returns {MTX}
    */
 
-  async createCancel(name, options) {
+  async _createCancel(name, options) {
     const mtx = await this.makeCancel(name);
     await this.fill(mtx, options);
     return this.finalize(mtx, options);
   }
 
   /**
-   * Create and send a cancel MTX.
+   * Create and finalize a cancel
+   * MTX with a lock.
+   * @param {String} name
+   * @param {Object} options
+   * @returns {MTX}
+   */
+
+  async createCancel(name, options) {
+    const unlock = await this.fundLock.lock();
+    try {
+      return await this._createCancel(name, options);
+    } finally {
+      unlock();
+    }
+  }
+
+  /**
+   * Create and send a cancel
+   * MTX without a lock.
    * @param {String} name
    * @param {Object} options
    */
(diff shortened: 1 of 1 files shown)


lib/wallet/wallet.js
```diff
@@ -1537,20 +1537,39 @@ class Wallet extends EventEmitter {
   }
 
   /**
-   * Create and finalize a open MTX.
+   * Create and finalize a open
```
boymanjor: "finalize an open"

lib/wallet/wallet.js
```diff
@@ -1537,20 +1537,39 @@ class Wallet extends EventEmitter {
...
-  async createOpen(name, force, options) {
+  async _createOpen(name, force, options) {
     const acct = options ? options.account || 0 : 0;
     const mtx = await this.makeOpen(name, force, acct);
     await this.fill(mtx, options);
     return this.finalize(mtx, options);
   }
 
+  /**
+   * Create and finalize a open
```
boymanjor: "finalize an open"

lib/wallet/wallet.js
```diff
@@ -2377,32 +2511,52 @@ class Wallet extends EventEmitter {
...
   async _sendRenewal(name, options) {
     const passphrase = options ? options.passphrase : null;
-    const mtx = await this.createRenewal(name, options);
+    const mtx = await this._createRenewal(name, options);
     return this.sendMTX(mtx, passphrase);
   }
 
   /**
-   * Create and send an renewal MTX.
+   * Create and send an renewal
```
boymanjor: "send a renewal"

tynes:
@boymanjor - All nits addressed, the commits fixing the nits were squashed into the appropriate commits.

PR #181, opened 2019-05-17
wallet: remove dead option

tynes:
The `witness` option in `lib/wallet/node.js` is left over from bcoin. This pull request deletes it because it is not used in the `walletdb`.


boymanjor:
LGTM

PR #187, opened 2019-05-25
node/http: coin amount for value/burned

tynes:
Other parts of the API return full coin amounts,
this switches the API for `GET /` to return
full coin amounts for the chain state at
`.chain.state.{value,burned}`.

The logger uses Amount.coin for both of these
values in the `chaindb.js` file.

```
this.logger.info(
  'Chain State: hash=%x tx=%d coin=%d value=%s burned=%s.',
  this.state.tip,
  this.state.tx,
  this.state.coin,
  Amount.coin(this.state.value),
  Amount.coin(this.state.burned));
```

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -20,6 +20,7 @@ const TX = require('../primitives/tx');
 const Claim = require('../primitives/claim');
 const Address = require('../primitives/address');
 const Network = require('../protocol/network');
+const Amount = require('../ui/amount');
 const pkg = require('../pkg');
 
 /**
@@ -129,8 +130,8 @@ class HTTP extends Server {
           state: {
             tx: this.chain.db.state.tx,
             coin: this.chain.db.state.coin,
-            value: this.chain.db.state.value,
-            burned: this.chain.db.state.burned
+            value: Amount.coin(this.chain.db.state.value, true),
+            burned: Amount.coin(this.chain.db.state.burned, true)
           }
         },
         pool: {


lib/node/http.js
```diff
@@ -129,8 +130,8 @@ class HTTP extends Server {
           state: {
             tx: this.chain.db.state.tx,
             coin: this.chain.db.state.coin,
-            value: this.chain.db.state.value,
-            burned: this.chain.db.state.burned
+            value: Amount.coin(this.chain.db.state.value, true),
+            burned: Amount.coin(this.chain.db.state.burned, true)
           }
```
tynes: The RPC interface often returns `Amount.coin` with the second argument to `true`. See bottom.

The code path for `Amount.coin` ends us in `Amount.encode` (below), where the second argument of `coin` is `num` in this context, which causes the number to be returned as a float.

https://github.com/handshake-org/hsd/blob/4cdaa33fa4483b3a6c74761479c90cb96bf9b36f/lib/ui/amount.js#L341-L345

https://github.com/handshake-org/hsd/blob/4a0ce94a390158e0d1e60838b818379403d1059e/lib/wallet/rpc.js#L505

pinheadmz:
I think I'm gonna call a NACK on this one - it's obviously a big breaking change. It's true the log messages often round off to nearest HNS `coin` but I think the HTTP API should stay consistent with its `dollarydoo` base unit.

Of course the RPC API typically uses `coin` and then, depending on input method the units might change again (see the warning in the docs here: https://hsd-dev.org/api-docs/#send-a-transaction). So we probably need to be very clear about this in the docs....

Anyway, gonna close for now. <3

PR #191, opened 2019-05-30
Log Identity Keys in net Module

tynes:
This PR adds the identity key to all of the log messages that log IP:Port in the `net` module. It does this by adding a `fullname` method to the `Peer` class which wraps `netAddress.fullname` which returns the `identitykey@ip:port` whereas `netAddress.hostname` returns just `ip:port`.

### DIFF
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -204,6 +204,15 @@ class Peer extends EventEmitter {
   }
 
   /**
+   * Getter to retrieve fullname.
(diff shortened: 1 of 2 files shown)


tynes:
This PR will be deprecated if https://github.com/handshake-org/hsd/pull/331 lands.

pinheadmz:
> This PR will be deprecated if #331 lands.

@tynes should this be closed?

PR #192, opened 2019-06-03
bench: add bench helper script

kilpatty:
This commit ports over the bench helper script from coin seen here: https://github.com/bcoin-org/bcoin/blob/master/bench/bench.js

### DIFF
--- a/bench/bench.js
+++ b/bench/bench.js
@@ -1,0 +1,13 @@
+'use strict';
+
+module.exports = function bench(name) {
+  const start = process.hrtime();
+  return function end(ops) {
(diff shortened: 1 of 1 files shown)


tynes:
Is the plan to use that bench function to wrap up this PR? https://github.com/handshake-org/hsd/pull/146

kilpatty:
Yes, that and I had some changes I was going to push into the Cipher State stuff in Brontide - Wanted to have bench available so I can test that my changes are an improvement in performance.

tynes:
This looks good to me.

cc: @boymanjor

PR #195, opened 2019-06-11
Wallet websocket covenant events

tynes:
Emit websocket events for each type of covenant when the wallet indexes the transaction. This is implemented by iterating over each output in the transaction. The topic is:

`${lowercase(type)} covenant`

For example `bid covenant`.

Includes tests for events over websocket using `bsock`.


lib/wallet/http.js
```diff
@@ -1303,6 +1303,21 @@ class HTTP extends Server {
         this.to('w:*', event, wallet.id, json);
     };
```
tynes: We could probably use `handleTX` instead of creating a new function here, but the difference is in `this.to`

lib/wallet/txdb.js
```diff
@@ -79,6 +79,52 @@ class TXDB {
...
+      const {type} = covenant;
+      const action = typesByVal[type].toLowerCase();
+      const channel = `${action} covenant`;
+      switch (type) {
+        case types.CLAIM:
+        case types.OPEN:
+        case types.BID:
+        case types.FINALIZE: {
+          const name = covenant.get(2).toString();
+          this.emit(channel, {name, output, tx}, details);
```
tynes: Emit `{ name, output, tx }` on the channel along with `details`

lib/wallet/txdb.js
```diff
@@ -79,6 +79,52 @@ class TXDB {
...
+          break;
+        }
+        case types.REVEAL:
```
tynes: For these covenants, the name is not in the covenant data, it requires a db lookup to look up the name based on the hash. I don't think this will result in too much io

boymanjor: To alleviate the added i/o, we could refactor [`this.connectNames`][1]. Currently, after processing incoming covenants, the method returns a boolean that indicates whether the name state was updated. We could instead have it return a summary object which accumulates all processed covenants.

We could then pass this summary object to `this.emitCovenants`.

[1]: https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/wallet/txdb.js#L1776

lib/wallet/txdb.js
```diff
@@ -79,6 +79,52 @@ class TXDB {
...
+   * Emit events for each type of covenant.
+   * @private
+   * @param {TXRecord} wtx
+   * @param {Details} details
+   */
+
+  async emitCovenants(tx, details) {
+    const {outputs} = tx;
+    for (const output of outputs) {
+      const {covenant} = output;
```
tynes: Could use something like:

```javascript
    if (!covenant.isName())
      continue;
```

lib/wallet/http.js
tynes: `handleTX` doesn't pass along enough arguments, we need to add another method or refactor `handleTX`. The problem is that `handleTX` does nothing with the `tx` argument, and the covenant related events pass along the argument at that position. A new function called `handleCovenant` was created.

lib/wallet/txdb.js
tynes: Replaced `{ name, output, tx }` with `NameState` json

test/wallet-http-test.js
```diff
@@ -1073,6 +1074,103 @@ describe('Wallet HTTP', function() {
...
+  it('should emit events for covenants', async () => {
+    const socket = bsock.connect(network.walletPort);
+
+    socket.on('connect', async () => {
+      await socket.call('auth', 'foo');
+      await socket.call('join', wallet.id);
+    });
+
+    const seen = {
+      open: false,
```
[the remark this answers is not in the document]
tynes: Added a test for `redeem`, using a second auction participant.

lib/wallet/txdb.js
```diff
@@ -79,6 +79,62 @@ class TXDB {
...
+      if (!covenant.isName())
+        continue;
+
+      const nameHash = covenant.get(0);
+      const ns = await view.getNameState(this, nameHash);
+
+      if (EMPTY.equals(ns.name) || !ns.name) {
```
tynes: The `view` accumulates state during `connectNames`. If the `name` was added to the `NameState` object, then skip the i/o involved with looking it up. Look it up in the covenant data if its present, otherwise query the wallet database for the name based on the `nameHash`.

lib/wallet/txdb.js
```diff
@@ -79,6 +79,62 @@ class TXDB {
...
+            break;
+          }
+          case types.REVEAL:
+          case types.REDEEM:
+          case types.REGISTER:
+          case types.UPDATE:
+          case types.RENEW:
+          case types.TRANSFER:
+          case types.REVOKE: {
+            const nameHash = covenant.get(0);
```
tynes: Delete `nameHash` variable assignment here, there is another above the `switch` statement.

test/wallet-http-test.js
```diff
@@ -70,9 +72,23 @@ describe('Wallet HTTP', function() {
     await wclient.createWallet('secondary');
     cbAddress = (await wallet.createAddress('default')).address;
     await wallet.createAccount(accountTwo);
+
```
tynes: `wsocket` listens on `wallet.id` and `wsocket2` listens on `wallet2.id`.

tynes:
> As discussed out of band, I think we should emit events for tree updates as well.

PR for emitting tree updates https://github.com/handshake-org/hsd/pull/201, the node websocket topic `"tree commit"`

tynes:
I don't think the additional i/o introduced by this PR is a problem because it first looks in the `nameview` that is populated in `connectNames` with a database lookup. It will only do another database lookup in the case that `nameview` does not already have the name.  We could add a runtime flag to disable the event emitting codepath (including the additional i/o) if we wanted to, but I don't think its necessary.

lib/wallet/txdb.js
tynes: See https://github.com/handshake-org/hsd/pull/195#discussion_r296451469

lib/wallet/txdb.js
```diff
@@ -79,6 +79,63 @@ class TXDB {
...
+          case types.TRANSFER:
+          case types.REVOKE: {
+            const nameState = await this.wallet.getNameState(nameHash);
+            ns.name = nameState.name;
+            break;
+          }
+        }
+      }
+
+      if (ns.name)
```
tynes: This is redundant because of the `getJSON` call below.

tynes: Fixed in ae70b8d461e1a5bb032195b208e0a363171991d4

pinheadmz:
concept ACK -- I'm gonna keep this on my to-review pile and maybe we can get it into v2.4.0

PR #196, opened 2019-06-17
tests: add unit tests for cipher state

kilpatty:
This commit adds some unit tests for cipher state as well as exchanges assert for bsert in the dependencies. The main point of these units tests are for other implementations to be able to copy so that they can insure their building blocks of brontide are correct.

### DIFF
--- a/test/brontide-test.js
+++ b/test/brontide-test.js
@@ -3,15 +3,169 @@
 
 'use strict';
 
-const assert = require('assert');
-const {Brontide} = require('../lib/net/brontide');
+const assert = require('bsert');
+const { CipherState, Brontide } = require('../lib/net/brontide');
 
 const HELLO = Buffer.from('hello', 'ascii');
 const PROLOGUE = 'lightning';
+const ROTATION_INTERVAL = 1000;
 
 /*
  * Tests
  */
+
+describe('CipherState', function() {
+  const cipher = new CipherState();
+  const key = Buffer.from(
+    '2121212121212121212121212121212121212121212121212121212121212121',
+    'hex'
+  );
+  const salt = Buffer.from(
+    '1111111111111111111111111111111111111111111111111111111111111111',
+    'hex'
+  );
+
+  it('should initalize with the key and salt', () => {
+    cipher.initSalt(key, salt);
+
+    assert.bufferEqual(cipher.key, key);
+
+    assert.bufferEqual(cipher.salt, salt);
+  });
+
+  it('should rotate the secret key', () => {
+    cipher.rotateKey();
+
+    assert.bufferEqual(
+      cipher.key,
+      Buffer.from(
+        '0b579ba44366e4d49ac7a44a8203925cb6d610e950aee7a23c47a5448173af11',
+        'hex'
+      )
+    );
+
+    assert.bufferEqual(
+      cipher.salt,
+      Buffer.from(
+        'be23775b41e7c67d1ec6dcfc21299f32461e145d4164f65943b4b99fcaff6dee',
+        'hex'
+      )
+    );
+
+    assert.strictEqual(cipher.nonce, 0);
+  });
+
+  it('should properly encrypt given text and empty ad', () => {
+    // Reset the cipher
+    const cipher = new CipherState();
+    cipher.initSalt(key, salt);
+
+    const hello = Buffer.from('hello', 'ascii');
+
+    const tag = cipher.encrypt(hello);
+
+    assert.bufferEqual(
+      tag,
+      Buffer.from('f11ae60b9df4c6ea25aea58ce1b6df83', 'hex')
+    );
+    assert.bufferEqual(hello, Buffer.from('0935b4c530', 'hex'));
+
+    // Round 2
+    const hello2 = Buffer.from('hello', 'ascii');
+
+    const tag2 = cipher.encrypt(hello2);
+
+    assert.bufferEqual(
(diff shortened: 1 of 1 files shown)


kilpatty:
Will fix lint tomorrow, my auto linting in Vim for some reason is not working.

test/brontide-test.js
```diff
@@ -13,6 +13,159 @@ const PROLOGUE = 'lightning';
...
+  });
+
+  it('should decrypt encrypted text throughout key rotation', () => {
+    const encryptionCipher = new CipherState();
+    encryptionCipher.initSalt(key, salt);
+
+    const decryptionCipher = new CipherState();
+    decryptionCipher.initSalt(key, salt);
+
+    for (let i = 0; i < 1001; i++) {
```
tynes: `1001` is a magic number, how do we know that the keys have actually been rotated?

test/brontide-test.js
```diff
@@ -13,6 +13,159 @@ const PROLOGUE = 'lightning';
...
+
+    assert.strictEqual(cipher.nonce, 0);
+  });
+
+  it('should properly encrypt given text and empty ad', () => {
+    // Reset the cipher
+    const cipher = new CipherState();
+    cipher.initSalt(key, salt);
+
+    const hello = Buffer.from('hello', 'ascii');
```
tynes: Can we use `HELLO`, defined at the top of the file?

test/brontide-test.js
```diff
@@ -13,6 +13,159 @@ const PROLOGUE = 'lightning';
...
+    // Round 2
+    const hello2 = Buffer.from('hello', 'ascii');
+
+    const tag2 = cipher.encrypt(hello2);
+
+    assert.bufferEqual(
+      tag2,
+      Buffer.from('d840242a1e817cd8374d45fb5621a5fc', 'hex')
+    );
+    assert.bufferEqual(hello2, Buffer.from('74898781da', 'hex'));
```
tynes: Where did the `74898781da` come from? It would be ideal if we had something similar to this - https://github.com/lightningnetwork/lightning-rfc/blob/master/08-transport.md#appendix-a-transport-test-vectors

I don't think it will work out of the box though because some of the parameters are different, right?

test/brontide-test.js
```diff
@@ -13,6 +13,159 @@ const PROLOGUE = 'lightning';
...
+    const tag = cipher.encrypt(hello);
+
+    assert.bufferEqual(
+      tag,
+      Buffer.from('f11ae60b9df4c6ea25aea58ce1b6df83', 'hex')
+    );
+    assert.bufferEqual(hello, Buffer.from('0935b4c530', 'hex'));
+
+    // Round 2
+    const hello2 = Buffer.from('hello', 'ascii');
```
tynes: Could we use `HELLO`, like above comment?

tynes:
Any additional comments you could leave around the tests would be super useful for people who want to implement against the test suite or review.

test/brontide-test.js
```diff
@@ -37,32 +190,41 @@ describe('Brontide', function() {
...
-      + 'ae';
+    const actTwo =
+      '' +
+      '0002466d7fcae563e5cb09a0d1870bb580344804617879a14' +
+      '949cf22285f1bae3f276e2470b93aac583c9ef6eafca3f730' +
+      'ae';
 
     initiator.recvActTwo(Buffer.from(actTwo, 'hex'));
 
     const actThree = initiator.genActThree();
```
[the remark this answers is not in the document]
kilpatty: Yeah, I'm going to go back and revert these so that it's easier. Not sure why the linter decided to newline them.

test/brontide-test.js
```diff
@@ -3,8 +3,8 @@
 
 'use strict';
 
-const assert = require('assert');
-const {Brontide} = require('../lib/net/brontide');
+const assert = require('bsert');
+const { CipherState, Brontide } = require('../lib/net/brontide');
```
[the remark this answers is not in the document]
kilpatty: Linter :/ will revert.

test/brontide-test.js
kilpatty: Good point, I copied the rotation interval constant from the brontide file into the test, that way it's a bit clearer that we are running this up to the rotation interval. 

Unfortunately it's not exported, so we can't pull the one automatically from the file, so if the rotation interval changes - it will need to be changed here as well.

test/brontide-test.js
kilpatty: Big fan of that document, I'm going to put one in my rust repo. I can include it in here as well if we think that would be a good idea. 

It actually does work out of the box with the LND tests. They just haven't tested the lower level stuff either, so no document for that yet.

test/brontide-test.js
kilpatty: The cipher encrypts the buffer in place, so using the constant would corrupt it for other tests. I suppose I could make a new buffer that copies the constant into itself so it's more clear.

test/brontide-test.js
kilpatty: See Above

test/brontide-test.js
kilpatty: Actually to clarify, it works out of the box w/ the LND tests if we change the prologue (which we do). We should probably build our own doc with the test results using Handshake's prologue that is used in production. I can work on that.

nodech:
needs rebase

PR #198, opened 2019-06-17
sighash `singlereverse`

tynes:
Implement a new sighash type `SINGLEREVERSE`. This sighash type is similar to `SINGLE`, but is different because instead of committing to the same output index as the input's index being signed, it commits to the index at `(outputs.length-1) - index`.

If the first input was signed with `SINGLEREVERSE`, then it would only commit to the final output.
This is useful because covenants are "linked", so that the next covenant in the series must be at the same index as the input. This new sighash type allows a user to commit to the covenant input but not the output for that covenant.

This is a building block for atomic name swaps on chain.

An additional way to do this would be `SINGLENEXT`, where the `i + 1`'th output is committed to.

### DIFF
--- a/lib/script/common.js
+++ b/lib/script/common.js
@@ -415,6 +415,15 @@ exports.hashType = {
   SINGLE: 3,
 
   /*
+   * Sign output at the reversed index (zero sequences).
+   * Input at index 0 would sign the output at the last
+   * index, input at index 1 would sign the second to
+   * last index and so on.
+   */
+
+  SINGLEREVERSE: 4,
+
+  /*
    * Sign no inputs.
    */
 
@@ -436,6 +445,7 @@ exports.hashTypeByVal = {
   1: 'ALL',
   2: 'NONE',
   3: 'SINGLE',
+  4: 'SINGLEREVERSE',
   0x40: 'NOINPUT',
   0x80: 'ANYONECANPAY'
 };
@@ -500,7 +510,7 @@ exports.isSignatureEncoding = function isSignatureEncoding(sig) {
   type &= ~exports.hashType.NOINPUT;
   type &= ~exports.hashType.ANYONECANPAY;
 
-  if (type < exports.hashType.ALL || type > exports.hashType.SINGLE)
+  if (type < exports.hashType.ALL || type > exports.hashType.SINGLEREVERSE)
     return false;
 
   if (!secp256k1.isLowS(sig.slice(0, -1)))
--- a/lib/primitives/tx.js
+++ b/lib/primitives/tx.js
@@ -297,6 +297,7 @@ class TX extends bio.Struct {
 
     if (!(type & hashType.ANYONECANPAY)
         && (type & 0x1f) !== hashType.SINGLE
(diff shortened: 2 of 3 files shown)


lib/script/common.js
```diff
@@ -414,6 +414,12 @@ exports.hashType = {
 
   SINGLE: 3,
 
+  /*
+   * Sign output at the n-i'th index (zero sequences).
```
tynes: I don't love this docstring, thinking of a better way to explain it

lib/script/common.js
```diff
@@ -414,6 +414,12 @@ exports.hashType = {
...
+   */
+
+  SINGLEREVERSE: 6,
```
tynes: I chose 6 to pad some space to allow porting of sighash types that come from Bitcoin more easily. Even though handshake is quite different from Bitcoin in some ways, its easier when things are more uniform. This may not be necessary and we could just use `4` because it impacts the `isSignatureEncoding` function below. Sighash types `4` and `5` will not cause this to return `false`:
https://github.com/handshake-org/hsd/pull/198/files#diff-4764bef9dd683c153cdffd9d3c06cf29R510

```javascript
if (type < exports.hashType.ALL || type > exports.hashType.SINGLEREVERSE)
```

chjj: I think we should either disallow 4 and 5 or make SINGLEREVERSE 3. Allowing 4 and 5 introduces malleability. I'm actually partial to the latter, but willing to go with either.

Thinking out loud here: when I was first looking at this, I thought for a second that we should make 4 or 5 disallowed by policy rules, but now I don't think it matters. This is a consensus check on handshake (as opposed to bitcoin where it's policy), but new sighash types can't actually be soft-forked in without a new witness version (or new opcode), so we don't really gain anything by making it policy only. We can disallow 4 and 5 by consensus rules and still add them in later with a new witness version.

chjj:
Code looks perfect aside from the 4 & 5 consideration.

lib/script/common.js
chjj: Hmm, on second thought this actually doesn't introduce malleability since the sighash itself commits to the hash type. I suppose bitcoin disallows it just because they want everything "weird" off the policy layer.

lib/script/common.js
tynes: Updated the docstring with something better. Its a bit longer but should be good enough for people to understand.

lib/script/common.js
tynes: > or make SINGLEREVERSE 3

Do you mean 4 here? `SINGLE` is already 3.

After thinking about it some more, skipping some numbers for the sighash types gives more headache than the small benefit gained from being able to more easily port Bitcoin based software to Handshake.  Who knows if/when Bitcoin will add new sighash types, its a definite premature optimization.

Just pushed up new commits that make `SINGLEREVERSE` 4

chjj: Yeah, sorry, 4 is what I meant. Not thinking clearly.

chjj:
Merging. Master is now unstable.

PR #201, opened 2019-06-19
node: event tree commit

tynes:
Emit an event over websockets for each time that a new tree root is set. 
The event is `tree commit` on the topic `chain`.

The event is added to `chain.setBestChain` which is called on every valid block that comes in. This means that this event could be triggered when a block that is not part of the best chain is received by the node.

Should we add additional information regarding if it is part of the best chain?
Or only make it so that the event is triggered when its part of the main chain?

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -1612,10 +1612,8 @@ class Chain extends AsyncEmitter {
   }
 
   /**
-   * Set the best chain. This is called on every valid block
-   * that comes in. It may add and connect the block (main chain),
-   * save the block without connection (alternate chain), or
-   * reorganize the chain (a higher fork).
+   * Set the best chain. This is called on every incoming
+   * block with greater chainwork than the current tip.
    * @private
    * @param {ChainEntry} entry
    * @param {Block} block
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -1671,6 +1671,9 @@ class Chain extends AsyncEmitter {
     this.emit('tip', entry);
     this.emit('block', block, entry);
 
+    if ((entry.height % this.network.names.treeInterval) === 0)
+      this.emit('tree commit', this.db.tree.rootHash(), entry, block);
+
     return this.emitAsync('connect', entry, block, view);
   }
 
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -1605,6 +1605,9 @@ class Chain extends AsyncEmitter {
     this.emit('tip', entry);
     this.emit('reconnect', entry, block);
 
+    if ((entry.height % this.network.names.treeInterval) === 0)
+      this.emit('tree commit', this.db.tree.rootHash(), entry, block);
+
     return this.emitAsync('connect', entry, block, view);
   }
 
--- a/test/node-http-test.js
+++ b/test/node-http-test.js
@@ -1,0 +1,176 @@
+/*!
+ * test/node-http-test.js - test for wallet http endoints
+ * Copyright (c) 2019, Mark Tyneway (MIT License).
+ * https://github.com/handshake-org/hsd
+ */
+
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+/* eslint no-return-assign: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const bio = require('bufio');
+const {NodeClient} = require('hs-client');
+const Network = require('../lib/protocol/network');
+const FullNode = require('../lib/node/fullnode');
+const Address = require('../lib/primitives/address');
+const Mnemonic = require('../lib/hd/mnemonic');
+const Witness = require('../lib/script/witness');
+const Script = require('../lib/script/script');
+const HDPrivateKey = require('../lib/hd/private');
+const Output = require('../lib/primitives/output');
+const Coin = require('../lib/primitives/coin');
+const MTX = require('../lib/primitives/mtx');
+const rules = require('../lib/covenants/rules');
+const common = require('./util/common');
+const mnemonics = require('./data/mnemonic-english.json');
+// Commonly used test mnemonic
+const phrase = mnemonics[0][1];
+
+const network = Network.get('regtest');
+const {types} = rules;
+
+const node = new FullNode({
+  network: 'regtest',
+  apiKey: 'foo',
+  walletAuth: true,
+  memory: true,
+  indexTx: true,
+  indexAddress: true,
+  rejectAbsurdFees: false
+});
+
+const nclient = new NodeClient({
+  port: network.rpcPort,
+  apiKey: 'foo'
+});
+
+let cbAddress, privkey, pubkey;
+let socketData, mempoolData;
+
+const {treeInterval} = network.names;
+
+describe('Node HTTP', function() {
+  this.timeout(15000);
+
+  before(async () => {
+    await node.open();
+    await nclient.open();
+    await nclient.call('watch chain');
+
+    const mnemonic = Mnemonic.fromPhrase(phrase);
+    const priv = HDPrivateKey.fromMnemonic(mnemonic);
+    const type = network.keyPrefix.coinType;
+    const key = priv.derive(44, true).derive(type, true).derive(0, true);
+    const xkey = key.derive(0).derive(0);
+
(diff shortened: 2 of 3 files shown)


boymanjor:
>Should we add additional information regarding if it is part of the best chain?
Or only make it so that the event is triggered when its part of the main chain?

To be clear, "best" and "main" are synonymous in this context, correct? You used them separately; I want to be sure my understanding is accurate. 

Are we sure that `setBestChain` handles blocks from alternate chains? The jsdoc comments lead me to believe this is true, but upon a first walkthrough of the code path, I cannot confirm the validity of this assumption.

`setBestChain` only gets called [here][1]. The enclosing branch either calls `setBestChain` or `saveAlternate`. The former emits the `tip`, `block`, and now information about the tree's root. The latter only emits `competitor` which signifies an alternate chain. Even the name of the method implies that this is only used to connect a block to the best chain. If you agree with this analysis we should update the comment and leave this change as-is. If I am wrong, I think we should indicate whether or not the `tree commit` event is tied to the best chain.

[1]: https://github.com/handshake-org/hsd/blob/a9d1742b58813dc0135e20b023ac4a43bd4e54a3/lib/blockchain/chain.js#L2023

tynes:
# Safety Report

The call stack for a block being added to the chain is as follows:

- `Chain.add` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1885)
- `Chain._add` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1904)
- `Chain.connect` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1979)
- `Chain.setBestChain` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1624)

The logic inside of `Chain.connect` determines if `Chain.setBestChain` will be called or if `Chain.saveAlternate` is called.

The relevant codepath:

https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L2008-L2021

If the incoming block's chainwork is less than or equal to the current tips chainwork, then `Chain.saveAlternate` is called. The only way that `Chain.setBestChain` is called is if the incoming block has greater chain work. This means that the jsdoc is __incorrect__ for `setBestChain` regarding the line `"This is called on every valid block that comes in."`. It is only called for blocks that extend
the current chain and not for alternative blocks.

https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1611-L1625

You can see here that the jsdoc is correct regarding `setBestChain` as a private function. The
only place that it is invoked is withing `lib/blockchain/chain.js` inside of `Chain.connect`.

```bash
$ basename $PWD
hsd
$ grep -rn setBestChain lib
lib/blockchain/chain.js:1464:    // That will be done outside in setBestChain.
lib/blockchain/chain.js:1624:  async setBestChain(entry, block, prev, flags) {
lib/blockchain/chain.js:2020:      await this.setBestChain(entry, block, prev, flags);
```

After doing this review, I believe that it is safe to include the event emitting exactly where it is
currently in this pull request. `Chain.setBestChain` is not called for every block, just blocks
extending the best chain. I will edit the commit messages to remove the safety warning
and include this report in the commit message.

### Additional Work

I think that reorgs should trigger the event `tree commit` as well. `Chain.connect` is only
called once for every block, if there is a reorg the codepath is as follows:

- `Chain.add` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1885)
- `Chain._add` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1904)
- `Chain.connect` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1979)
- `Chain.setBestChain` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1624)
- `Chain.reorganize` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1432)
- `Chain.reconnect` - [source](https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1571)

A `tree commit` event could be added inside of `Chain.reconnect` as well, it also emits events:
https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/blockchain/chain.js#L1605-L1606

This will ensure that the event is emitted when reorgs happen and different tree roots are committed.

tynes:
> > Should we add additional information regarding if it is part of the best chain?
> > Or only make it so that the event is triggered when its part of the main chain?
> 
> To be clear, "best" and "main" are synonymous in this context, correct? You used them separately; I want to be sure my understanding is accurate.

Yes that is what I meant. Sorry for using two words to mean the same thing.

> If you agree with this analysis we should update the comment and leave this change as-is

I agree with the analysis and laid out evidence above. Will update the commit appropriately.

chjj:
To make things clear: any time `verifyContext` is called, the block is being added to the main chain. The current code looks correct to me. There's probably a number of incorrect jsdocs in the codebase due to refactors and whatnot (I think I'm getting sick of maintaining jsdoc =/).

tynes:
Just pushed up a fix to the jsdoc.

> I think I'm getting sick of maintaining jsdoc =/

What do you think of having high level `README`s in each package similar to btcd/lnd?
The one thing that I do like about jsdoc is the type hints. Without those, things could become pretty messy. Can't wait until Typescript *is* Javascript.

lib/blockchain/chain.js
```diff
@@ -1605,14 +1605,15 @@ class Chain extends AsyncEmitter {
     this.emit('tip', entry);
     this.emit('reconnect', entry, block);
 
+    if ((entry.height % this.network.names.treeInterval) === 0)
+      this.emit('tree commit', this.db.tree.rootHash(), block, entry);
```
tynes: I notice that the `entry` seems to come before the `block` in the other events. Going to update these changes so that it is more consistent with the other events.

test/node-http-test.js
```diff
@@ -0,0 +1,185 @@
...
+  after(async () => {
+    await nclient.close();
+    await node.close();
+  });
+
+  describe('Websockets', function () {
+    describe('tree commit', () => {
+      it('should mine 1 tree interval', async () => {
+        await mineBlocks(treeInterval, cbAddress);
+
```
boymanjor: nit: let's get rid of this newline.

test/node-http-test.js
```diff
@@ -0,0 +1,185 @@
...
+        const mempool = await nclient.getMempool();
+        assert.equal(mempool[0], mtx.txid());
+
+        await mineBlocks(treeInterval, cbAddress);
+        assert.equal(socketData.length, 1);
+
+        const {root, block, entry} = socketData[0];
+
+        // the root is committed to in the next block
```
boymanjor: nit: this comment should be updated. We are asserting that a new root was committed and that the websocket event was fired.

test/node-http-test.js
```diff
@@ -0,0 +1,185 @@
...
+        assert.bufferEqual(node.chain.db.treeRoot(), root);
+
+        const info = await nclient.getInfo();
+
+        // the root has updated
```
boymanjor: nit: this comment seems irrelevant. What we are really checking is that the tip was updated.

tynes:
@boymanjor Nits are addressed, thanks for the review. I removed the comments in the test since it reads fairly straightforward.

PR #204, opened 2019-06-19
framer: fix jsdoc parameters

kilpatty:
This commit fixes a small jsdoc param that is not longer used.


tynes:
Looks good to me.

PR #206, opened 2019-06-21
pkg: link hs-client to bin

pinheadmz:
This is a port of https://github.com/bcoin-org/bcoin/pull/707

After installation, links hsd-cli and hsw-cli directly to hs-client bin, already installed. 

Adds full deprecation message to `cli`.
- [x] Should we just remove `cli` all together? Seems funny to have a deprecation warning on unreleased software.
- [ ] `hsd-rpc` and `hsw-rpc` are a little trickier since those commands are actually bash scripts that hotwire `hsd-cli rpc ...`, etc. Leaving them out for now.

This need was brought to my attention by a community post: https://handshake.community/t/help-chain-is-loading-mac-hsd/100

User seemed a little confused about CLI usage and I realized there is no explicit docs for CLI usage in HSD. (And actually in bcoin, it is a little [buried](https://github.com/bcoin-org/bcoin/blob/master/docs/cli.md) )

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,13 @@ Note that `node-gyp` must be installed. See the
 [node-gyp](https://github.com/nodejs/node-gyp) documentation for more
 information.
 
+## CLI
+
+HSD comes with command-line interface tools `hsd-cli` (to interact with the node
+server) and `hsw-cli` (to interact with the wallet server). These applications
+are available in `./bin` (for example the command `./bin/hsd-cli info` returns
+basic node info). CLI usage in the API docs refers to these applications.
+
 ## Documentation
 
 - Documentation Site: [https://handshake-org.github.io](https://handshake-org.github.io)
--- a/README.md
+++ b/README.md
@@ -27,6 +27,19 @@ server) and `hsw-cli` (to interact with the wallet server). These applications
 are available in `./bin` (for example the command `./bin/hsd-cli info` returns
 basic node info). CLI usage in the API docs refers to these applications.
 
+When `hsd` is installed globally, CLI commands are available without the path:
+
+```
+$ hsd-cli info
+```
+
+RPC commands are available with `hsd-cli rpc <command>` and `hsw-cli rpc <command>`.
+The shortcuts `hsd-rpc` and `hsw-rpc` are available if you install hs-client globally:
+
+```
+$ npm install -g hs-client
+```
+
 ## Documentation
 
 - Documentation Site: [https://handshake-org.github.io](https://handshake-org.github.io)
--- a/package.json
+++ b/package.json
@@ -57,7 +57,9 @@
     "hsd-node": "./bin/node",
     "hsd-spvnode": "./bin/spvnode",
     "hs-seeder": "./bin/hs-seeder",
-    "hs-wallet": "./bin/hsw"
+    "hs-wallet": "./bin/hsw",
+    "hsd-cli": "./bin/hsd-cli",
+    "hsw-cli": "./bin/hsw-cli"
   },
   "scripts": {
     "lint": "eslint $(cat .eslintfiles) || exit 0",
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -32,6 +32,14 @@ $ python --version
 
 See the [node-gyp](https://github.com/nodejs/node-gyp) documentation for more information.
 
+## CLI
+
+HSD comes with command-line interface tools `hsd-cli` (to interact with the node
+server) and `hsw-cli` (to interact with the wallet server). These applications
+are available in `/bin` (for example the command `./bin/hsd-cli info` returns
```
tynes: Nit: `/bin` is an absolute path which usually means the root of the filesystem and not the root of the git repo, I think `bin` is more appropriate.

boymanjor: Or even `./bin` (which is the relative location of the file w.r.t. README.md).

tynes:
Responding to your questions:

> Should we just remove cli all together? Seems funny to have a deprecation warning on unreleased software.

I agree and am not super opinionated. Leaning towards leaving it just to prevent problems with potentially out of date docs.

> `hsd-rpc` and `hsw-rpc` are a little trickier since those commands are actually bash scripts that hotwire hsd-cli rpc ..., etc. Leaving them out for now.

I agree with leaving them out, we could refactor `*-rpc` to a node.js script if we really wanted to, but I don't think it is all that high of a priority right now.

README.md
pinheadmz: added a `.`! Also specified the npm package name in this line:

> ... or they may be installed globally with `npm i -g hs-client`

pinheadmz:
@tynes rebased this and removed the unused `bin/cli`. I think this would be super helpful for https://github.com/Homebrew/homebrew-core/pull/51014

But the rpc is still an issue because we don't document anywhere the `hsd-cli rpc...` syntax. Thoughts?

package.json
```diff
@@ -57,7 +57,10 @@
     "hsd-node": "./bin/node",
     "hsd-spvnode": "./bin/spvnode",
     "hs-seeder": "./bin/hs-seeder",
-    "hs-wallet": "./bin/hsw"
+    "hs-wallet": "./bin/hsw",
+    "hs-wallet": "./bin/hsw",
```
boymanjor: This line appears twice.

pinheadmz:
RPC usage updated in README: https://github.com/handshake-org/hsd/pull/206/commits/157a5cd9c4f40c763b985fbe81f72c7b0ed4e52c

README.md
```diff
@@ -20,6 +20,21 @@ Note that `node-gyp` must be installed. See the
...
+## CLI
+
+HSD comes with command-line interface tools `hsd-cli` (to interact with the node
+server) and `hsw-cli` (to interact with the wallet server). These applications
+are available in `./bin` (for example the command `./bin/hsd-cli info` returns
+basic node info). CLI usage in the API docs refers to these applications.
+
+When `hsd` is installed globally, CLI commands are available without the path:
+
+`$ hsd-cli info`
```
boymanjor: Let's stick to convention and use ``` here, i.e.
```
$ hsd-cli info
```

README.md
```diff
@@ -20,6 +20,21 @@ Note that `node-gyp` must be installed. See the
...
+
+RPC commands are available with `hsd-cli rpc <command>` and `hsw-cli rpc <command>`.
+The shortcuts `hsd-rpc` and `hsw-rpc` are available if you install hs-client globally:
+`npm install -g hs-client`.
```
boymanjor: Ditto.

README.md
boymanjor: Or keep it on the same line without the $.

pinheadmz: 👍

README.md
pinheadmz: 👍  👍  👍

PR #209, opened 2019-07-02
rules: comments

tynes:
Add some comments to the file `lib/covenants/rules.js` and correct one comment that was not correct.


pinheadmz:
Merging this, we can always use more comments! This will be a good check to make sure CIrcleCI isn't running on master branch any more (should be github workflows only)

PR #210, opened 2019-07-02
node/http: get header by hash/height

tynes:
It can be useful to query for block headers directly without needing to query for full blocks.
Less disk i/o and less bandwidth would be needed when querying for just the block header.
It will also work for nodes in SPV mode.

This is a port of a PR to bcoin here https://github.com/bcoin-org/bcoin/pull/808/files

### DIFF
--- a/test/http-test.js
+++ b/test/http-test.js
@@ -344,6 +344,65 @@ describe('HTTP', function() {
     assert.equal(filterInfo.entries, 1);
   });
 
+  it('should generate 10 blocks from RPC call', async () => {
+    const blocks = await nclient.execute('generatetoaddress', [10, addr.toString(network)]);
+    assert.strictEqual(blocks.length, 10);
+  });
+
+  // depends on the previous test to generate blocks
+  it('should fetch block header by height', async () => {
+    // fetch corresponding header and block
+    const height = 7;
+    const header = await nclient.get(`/header/${height}`);
+    assert.equal(header.height, height);
+
+    const properties = [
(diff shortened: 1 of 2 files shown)


tynes:
Updated to match latest changes https://github.com/bcoin-org/bcoin/pull/808/files

tynes:
Accidentally closed when pruning old git branches. Still open for review, but not critical to have merged now.

tynes:
PR to docs here https://github.com/handshake-org/handshake-org.github.io/pull/68

test/http-test.js
```diff
@@ -344,6 +344,65 @@ describe('HTTP', function() {
...
+    assert.equal(block.prevBlock, header.prevBlock);
+    assert.equal(block.merkleRoot, header.merkleRoot);
+    assert.equal(block.time, header.time);
+    assert.equal(block.bits, header.bits);
+    assert.equal(block.nonce, header.nonce);
+  });
+
+  it('should fetch null for block header that does not exist', async () => {
+    // many blocks in the future
+    const header = await nclient.get(`/header/${40000}`);
```
pinheadmz, line 382: Curious why is `40000` in braces?
This could be foolproof by getting chain tip height and adding any value > 1
But no change needed really.

tynes:
> ACK, would be nice to have CLI update in hs-client as well.

Here is the method added to `hs-client`. Do you think there should also be a command added like:

```
$ hsd-cli header <heightOrHash>
```

https://github.com/handshake-org/hs-client/pull/28

PR #212, opened 2019-07-02
net: remove discoverExternal() from the pool

kilpatty:
This commit removes the discover external check in the pool. Currently
we already store all local addresses that are sent to us from peers in
the hostlist, so there is no reason to rely on an external service to
check for our local/external address. This was removed from bitcoin in
commit 845c86d.

EDIT: This is blocked by implementing the saving of local addresses via the remote property in the version packet.

### DIFF
--- a/lib/net/pool.js
+++ b/lib/net/pool.js
@@ -277,7 +277,6 @@ class Pool extends EventEmitter {
     await this.hosts.open();
 
     await this.discoverGateway();
-    await this.discoverExternal();
     await this.discoverSeeds();
 
     this.fillOutbound();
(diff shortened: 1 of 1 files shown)


kilpatty:
ping @pinheadmz

pinheadmz:
utACK. 

This function hasn’t really worked in a while anyway:

https://github.com/bcoin-org/bdns/issues/1

tynes:
utACK.

Has there been discussion around this potentially leaking information about people running nodes? Probably don't want extra external API calls that the user is not necessarily aware of.

pinheadmz:
@tynes plenty of discussion in bitcoin core. Especially when it got funny: https://github.com/bitcoin/bitcoin/issues/4679

nodech:
Merged in 6213460846142edcdad3835dad7473e5c83f9f3e from #692

PR #213, opened 2019-07-02
config: add discover option for peer discovery

kilpatty:
This commit adds a configuration boolean named discover. It mimics the
behavior provided in commit 845c86d. Discover allows the user to
configure their node to *not* share their own peer address with other
peers. This allows for a node to listen, but choose not to share their
local address.

### DIFF
--- a/lib/node/fullnode.js
+++ b/lib/node/fullnode.js
@@ -80,6 +80,8 @@ class FullNode extends Node {
       indexAddress: this.config.bool('index-address')
     });
 
+    this.listen = this.config.bool('listen');
+
     // Pool needs access to the chain and mempool.
     this.pool = new Pool({
       network: this.network,
@@ -104,8 +106,9 @@ class FullNode extends Node {
       publicPort: this.config.uint('public-port'),
       host: this.config.str('host'),
       port: this.config.uint('port'),
-      listen: this.config.bool('listen'),
-      memory: this.config.bool('memory')
+      listen: this.listen,
+      memory: this.config.bool('memory'),
+      discover: this.config.bool('discover', this.listen )
     });
 
     // Miner needs access to the chain and mempool.
--- a/lib/net/pool.js
+++ b/lib/net/pool.js
@@ -1333,7 +1333,7 @@ class Pool extends EventEmitter {
 
   async handleOpen(peer) {
     // Advertise our address.
-    if (!this.options.selfish && this.options.listen) {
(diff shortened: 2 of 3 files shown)


kilpatty:
ping @pinheadmz and @tynes Added the discover option from the same PR that removes External IP services from Bitcoin. I have defaulted it to false, which is also what Bitcoin does from my understanding, but I think this should be open to discussion as it requires another flag for users to run a fully fledged node that listens and is discoverable.

pinheadmz:
I'm actually not sure if we need this in addition to `listen`... is there any reason to config with
```
listen: false
discover: true
```
??

kilpatty:
No, but there is a reason to config with: 
```
listen: true
discover: false
```
In which case you would want to accept inbound connections, but only to peers that you share your address with. Additionally those peers you share with won't share your address to their peers.

pinheadmz:
Ah I see ok interesting.
Bitcoin Core also has "whitebind" options for both port and IP... I wonder if that approach is any more effective for that use case.

kilpatty:
Whitebind is interesting, but it still allows that address to be shared. So any peers that connect to it will share it with all their peers. 

On second thought, I might switch the default of discover to whatever listen is -> That is more inline with how core handles it: https://en.bitcoin.it/wiki/Running_Bitcoin

That way if you enable --listen, you won't have to also enable discover. The only time you would touch discover is if you want to listen and explicitly not be discovered. That is probably for more niche use cases, and defaulting to listen's value won't make it more cumbersome for current users. Thoughts?

lib/node/fullnode.js
```diff
@@ -105,7 +105,8 @@ class FullNode extends Node {
       host: this.config.str('host'),
       port: this.config.uint('port'),
       listen: this.config.bool('listen'),
-      memory: this.config.bool('memory')
+      memory: this.config.bool('memory'),
+      discover: this.config.bool('discover', this.config.bool('listen'))
```
pinheadmz: I think the style preference for these is to put `this.listen = this.config.bool('listen', false)` at the top, then `this.listen` can be used multiple times throughout. Makes it easier to change default values later I think.

kilpatty: Ah okay sounds good will change that!

kilpatty: Okay changed this to declare this.listen above, let me know if that looks better

pinheadmz:
see commit 53ed342749b605ea77e64ac16af1b43b167e8491 possibly closed already

PR #222, opened 2019-07-17
peer: set key if inbound and brontide succeeds

kilpatty:
Currently on hnscan we noticed that no inbound peers show their identity key when printed using "getpeerinfo". This change sets the key on the net address, inside of an inbound peer if the brontide Handshake is successful.

### DIFF
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -409,6 +409,8 @@ class Peer extends EventEmitter {
       this.brontide.once('connect', () => {
         this.time = Date.now();
         this.connected = true;
+        if (!this.outbound)
+          this.address.setKey(this.brontide.remoteStatic);
         this.emit('connect');
 
         cleanup();


pinheadmz:
Very cool! Confirmed the functionality by bypassing the [hard-coded regtest ID key](https://github.com/kilpatty/hsd/blob/257e70940c2d23d7b132665c175df36fa3c25df4/lib/node/node.js#L166-L167) and connecting two nodes in regtest. Both nodes returned their peer's identity key in `rpc getpeerinfo` :+1:

pinheadmz:
Brontide will be refactored an optional soon, this PR will need a rebase after that but I think it's still a good idea for nodes that HAVE a brontide connection, so I think we can leave this open for now.

tynes:
This functionality was merged as part of 340952867a4fdfbe0e0ee9ab9a6342376ae1f7f8

Thank you @kilpatty for figuring out how to do this.

PR #223, opened 2019-07-17
address: typo on from scripthash assertion

kilpatty:


lib/primitives/address.js
```diff
@@ -307,7 +307,7 @@ class Address extends bio.Struct {
    */
 
   fromScripthash(hash) {
-    assert(hash && hash.length === 32, 'P2WPKH must be 32 bytes.');
+    assert(hash && hash.length === 32, 'P2SH must be 32 bytes.');
```
tynes: As a community, do we refer to the scripts as P2SH and assume the W, or explicitly call it P2WSH?

I like including the W so that people from all communities can more easily reason about scripting with Handshake.

pinheadmz: +1 for `W`

kilpatty: Noted! Will change to W.

PR #226, opened 2019-07-17
test: opcodes that introspect tx outputs

pinheadmz:
Increase test coverage of `Script` module with the new handshake opcodes `OP_CHECKOUTPUT` and `OP_TYPE`.

These unit tests could be the basis for an integration test or separate application that uses hot/cold keys to build the update/transfer name-ownership script described in the whitepaper.

### DIFF
--- a/test/script-introspection-test.js
+++ b/test/script-introspection-test.js
@@ -1,0 +1,72 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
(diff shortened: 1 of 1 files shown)


pinheadmz:
Updated to include OP_TYPE only, assuming https://github.com/handshake-org/hsd/pull/230 will be merged and remove OP_CHECKOUTPUT from the protocol

pinheadmz:
rebase to  c5daaaadfdcb02e7b6c574fcfc1db533d59903df:
rebase on master, diff is null

Issue #224, opened 2019-07-17
Discussion: should OP_CHECKOUTPUT specify an output index?


pinheadmz:
OP_CHECKOUTPUT as it is currently implemented gets three values from the stack:

```
const version = stack.getInt(-3, minimal, 4);
const hash = stack.get(-2);
const num = stack.getNum(-1, minimal, 8);
```

... and then compares these with the output _at the same index as the input in which the opcode is executed_:

```
function checkOutput(tx, index, value, version, hash) {
  if (index >= tx.outputs.length)
    return false;

  const output = tx.outputs[index];

  if (output.value !== value)
    return false;

  if (output.address.version !== version)
    return false;

  if (!output.address.hash.equals(hash))
    return false;

  return true;
}
```

This is a cool scheme for coin-spending "covenants" (in the broader meaning of the term) but it is limited by the corresponding-index requirement. Since Handshake `covenants` all very specifically require inputs and outputs to be linked, it prevents them from being combined with OP_CHECKOUTPUT.

### Proposal

OP_CHECKOUTPUT takes one additional parameter off the stack, which specifies an output to introspect and check the other three values.

If that output does not exist, the script fails.

Some of this basic logic already exists. The code could be updated with the minimal diff below. Note that `index` is already declared in the `execute()` function as the index of the input being verified, so during the execution of OP_CHECKOUTPUT we'll need to declare `outputIndex` separately.

```diff
diff --git a/lib/script/script.js b/lib/script/script.js
index 45d6e2f2..d25a01e1 100644
--- a/lib/script/script.js
+++ b/lib/script/script.js
@@ -586,6 +586,7 @@ class Script extends bio.Struct {
           if (stack.length < 3)
             throw new ScriptError('INVALID_STACK_OPERATION', op, ip);

+          const outputIndex = stack.getInt(-4, minimal, 4);
           const version = stack.getInt(-3, minimal, 4);
           const hash = stack.get(-2);
           const num = stack.getNum(-1, minimal, 8);
@@ -600,11 +601,12 @@ class Script extends bio.Struct {
             throw new ScriptError('INVALID_VALUE', op, ip);

           const val = num.isZero() ? value : num.toDouble();
-          const res = checkOutput(tx, index, val, version, hash);
+          const res = checkOutput(tx, outputIndex, val, version, hash);

           stack.pop();
           stack.pop();
           stack.pop();
+          stack.pop();

           stack.pushBool(res);
```

### Use case

[...]

PR #227, opened 2019-07-18
[Consensus] Script: require output index for OP_CHECKOUTPUT 

pinheadmz:
Implementation of the proposal at https://github.com/handshake-org/hsd/issues/224.
Tests built off of https://github.com/handshake-org/hsd/pull/226

### Modified consensus rule

OP_CHECKOUTPUT and OP_CHECKOUTPUTVERIFY consume one extra value from the stack that indicates _which output_ of the spending transaction to check (for value, address hash and address version).

### Motivation

Handshake has two types of covenants: The literal primitive object that is associated with every transaction output and restricts name updates and transfers, and the more conceptual scheme that restricts payments of the hns coin. The latter has been proposed in other cryptocurrency systems including opcodes like OP_CHECKOUTPUT. The former is more specifically integrated in handshake, and these native covenants _require that input and output indexes be identical_. That rule is enforced in multiple places like `rules.js` and `chain.js`.

Unfortunately, the native covenant index-rule prevents combining both types of covenants in the same transaction. In other words, OP_CHECKOUT can not be used effectively in a native name-related covenant because the output at that index is already restricted by the native covenant rules.

Allowing the sender to define _which output_ should be checked by the opcode makes it way more flexible, and allows simple on-chain name-for-money atomic swaps (as outlined in issue #224)

### References

[Bitcoin Covenants by: Malte Moser, Ittay Eyal, and Emin Gun Sirer](https://maltemoeser.de/paper/covenants.pdf) describes OP_CHECKOUTPUTVERIFY:
> The opcode expects the index as the first parameter, allowing to place it in the input of the spending transaction. The creator of the spending transaction can therefore determine the output’s position.

[Chain also implemented OP_CHECKOUTPUT](https://chain.com/docs/1.2/protocol/specifications/vm1#introspection-instructions). Step 1 in the algorithm is:
> 1. Pops 6 items from the data stack: index, data, amount, assetid, version, prog.

### SINGLEREVERSE

This consensus change doesn't conflict with #198 and has different trade-offs in the on-chain atomic-swap application. We could have both features in Handshake. (TODO: exemplify both protocols and compare TX sizes and other trade-offs)

### Note to reviewers

I want to make sure that the `getInt()` line below is correct especially regarding the size (4 bytes) - is there such thing as an hsd transaction with more outputs than that? 
`const outputIndex = stack.getInt(-4, minimal, 4);`

### DIFF
--- a/lib/script/script.js
+++ b/lib/script/script.js
@@ -586,6 +586,7 @@ class Script extends bio.Struct {
           if (stack.length < 3)
             throw new ScriptError('INVALID_STACK_OPERATION', op, ip);
 
+          const outputIndex = stack.getInt(-4, minimal, 4);
           const version = stack.getInt(-3, minimal, 4);
           const hash = stack.get(-2);
           const num = stack.getNum(-1, minimal, 8);
@@ -600,8 +601,9 @@ class Script extends bio.Struct {
             throw new ScriptError('INVALID_VALUE', op, ip);
 
           const val = num.isZero() ? value : num.toDouble();
-          const res = checkOutput(tx, index, val, version, hash);
+          const res = checkOutput(tx, outputIndex, val, version, hash);
 
+          stack.pop();
           stack.pop();
           stack.pop();
           stack.pop();
--- a/test/atomic-name-sale-checkoutput-test.js
+++ b/test/atomic-name-sale-checkoutput-test.js
@@ -1,0 +1,272 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const Network = require('../lib/protocol/network');
+const Script = require('../lib/script/script');
+const Opcode = require('../lib/script/opcode');
+const FullNode = require('../lib/node/fullnode');
+const MTX = require('../lib/primitives/mtx');
+const Output = require('../lib/primitives/output');
+const Input = require('../lib/primitives/input');
+const Address = require('../lib/primitives/address');
+const rules = require('../lib/covenants/rules');
+const Resource = require('../lib/dns/resource');
+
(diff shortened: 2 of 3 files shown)


tynes:
> I want to make sure that the getInt() line below is correct especially regarding the size (4 bytes) - is there such thing as an hsd transaction with more outputs than that?
const outputIndex = stack.getInt(-4, minimal, 4);

Looking at `TX.read`, the number of outputs is encoded in a varint.

https://github.com/handshake-org/hsd/blob/cc1ef7ab8dfcadbcc1b18b934ccf2d2ec1776042/lib/primitives/tx.js#L1806

The number of outputs should depend on the maximum transaction size.
UTXOs with covenants must be linked, so that adds an extra input for each covenant type that is not `NONE`. 

----
Some background info:

An output is defined by:

| Field  | Type | Size (bytes) | Source |
| --- | ----- | ---- | --- |
| value | U64 | 8 | https://github.com/handshake-org/hsd/blob/master/lib/primitives/output.js#L274
| address | Address | 22/34 | https://github.com/handshake-org/hsd/blob/master/lib/primitives/address.js#L411 |
| covenant | Covenant |  Variable, 585 max | https://github.com/handshake-org/hsd/blob/master/lib/primitives/covenant.js#L511

The consensus maximum transaction size is:
https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/consensus.js#L232

There is also a policy maximum transaction size:
https://github.com/handshake-org/hsd/blob/4b04d9a2f9094b9c4c8e17346e7f24517f0a0336/lib/protocol/policy.js#L30
https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/consensus.js#L160

PR #229, opened 2019-07-27
test: add mining test for new proof of work

kilpatty:
This commit adds multiple tests to test the new proof of work function.
Mainly, mining a block is tested, as well as mining a block with a mask
that mimics what they would receive from a pool.

We also added in an extra parameter to the mine function to allow for a
nonce to be passed in instead of starting with 0. This should have zero
side effects as if it is not included, we default to the previous value
(0). We also add a funciton named mineRandomNonce that uses a random
nonce value to start iterating from - simulating a pool-like
environment.

Co-Authored-By: Mark Tyneway <mark@purse.io>
Co-Authored-By: Sean Kilgarriff <seanpkilgarriff@gmail.com>

### DIFF
--- a/lib/mining/cpuminer.js
+++ b/lib/mining/cpuminer.js
@@ -8,6 +8,7 @@
 
 const assert = require('bsert');
 const bio = require('bufio');
+const random = require('bcrypto/lib/random');
 const EventEmitter = require('events');
 const {Lock} = require('bmutex');
 const util = require('../utils/util');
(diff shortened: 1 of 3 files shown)


lib/mining/cpuminer.js
```diff
@@ -298,6 +299,37 @@ class CPUMiner extends EventEmitter {
...
+    const target = job.attempt.target;
+
+    if (!interval)
+      interval = CPUMiner.INTERVAL;
+
+    let nonce = 0;
+    let solved = false;
+
+    for (;;) {
+      nonce = random.randomRange(0, 0xffffffff);
```
tynes: We should add a constant to `lib/protocol/consensus` for both the size of the nonce and the extra nonce so that we don't have to hardcode in a magic number here

PR #230, opened 2019-07-29
[consensus] Script: remove OP_CHECKOUTPUT and OP_CHECKOUTPUTVERIFY

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/224
Replaces https://github.com/handshake-org/hsd/pull/227

- [x] : Update https://github.com/handshake-org/hsd/pull/226

It seems this opcode may be more trouble than it is worth for Handshake. We already have a native covenant system, and a proposed consensus update to enable on-chain atomic name sales (https://github.com/handshake-org/hsd/pull/198).

This PR removes the opcode from the scripting system and adjusts tests that called it to throw the appropriate error (`BAD_OPCODE`)

### DIFF
--- a/lib/script/common.js
+++ b/lib/script/common.js
@@ -159,8 +159,6 @@ exports.opcodes = {
 
   // More Ops
   OP_TYPE: 0xd0,
-  OP_CHECKOUTPUT: 0xd1,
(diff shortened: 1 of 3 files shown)


chjj:
Yeah, I agree. Going to merge this. Master should now be considered unstable.

Issue #231, opened 2019-07-31
"Could not resolve preferred inputs." Error when calling "sendreveal"


wi-ski:
### Context
I have a wallet - with two separate accounts, both bidding for the same name - calling sendreveal and passing an account name throws this error `"Could not resolve preferred inputs."`.  I'm totally able to call sendreveal when using only **1** account and calling sendreveal.

This guy: https://github.com/handshake-org/hsd/blob/e1c0281dd9e9bbbd23f5cd832c82ebe3808e9b4f/lib/primitives/mtx.js#L1710

I have no idea what a "preferred input" is in this context. Could I get some guidance here? :heart:  @tynes (we spoke about this over Telegram) cc. @chjj

wi-ski:
After reading the description - I thought I'd reiterate:

a. The wallets has two accounts (loser and winner)
b. Both accounts have placed a bid on a name ("foo")
c. Once the auction has reached the reveal period, calling sendReveal - and passing each account's name as a param - causes the error.

Worth Mentioning:
1. When the wallet only has one account involved in the auction, im able to call sendReveal and pass the account name.
2. Using the sendRevealAll 'works' - eg. it doesnt error.

pinheadmz:
I can confirm this issue with a brief test:

https://gist.github.com/pinheadmz/fc9b4d0fcbb4041ebd8e938f9f52b92d

`Error: Could not resolve preferred inputs.`

I'll go down this rabbit-hole with ya, @wi-ski !

pinheadmz:
@wi-ski got a WIP answer for ya: https://github.com/handshake-org/hsd/pull/233

wi-ski:
Oh heck yea! Thank you so much for taking a look :) Sincerely, you're the best. I just tested your fix and it totally worked :tada:

wi-ski:
I _believe_ Im following :thinking: Ill take a deeper look in the day time, but this feels straight forward.

Again - thank you for the help :)

wi-ski:
Alright - looks like we need the same for `sendupdate`: https://github.com/pinheadmz/hsd/pull/1/files

:point_down: Is that in the right spirit? Tests incoming if so.

PR #233, opened 2019-08-01
wallet: accept account argument for all covenant makers and verify account ownership during coin selection

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/231

Adds new method hasCoinByAccount(acct, hash, index) to txdb which
returns true if the coin (hash, index) is owned by account (number).
Adds this check to wallet.makeReveal() to prevent BIDs from other
wallet accounts being included as inputs to a new REVEAL TX from a
specific account.

TODO:
- [x] Put the test `double-reveal-test.js` in a more appropriate place... (`auction-test`? `wallet-test`?)
- [x] Finish the auction process in the test, make sure the same bug doesn't throw during any other auction or name-ownership function like `UPDATE`, `REGISTER`. etc...
- [x] Ensure RPC and HTTP APIs pass `account` parameters

### DIFF
--- a/lib/wallet/txdb.js
+++ b/lib/wallet/txdb.js
@@ -2251,6 +2251,20 @@ class TXDB {
   }
 
   /**
+   * Test whether an account owns a coin.
+   * @param {Number} acct
+   * @param {Hash} hash
+   * @param {Index} number
+   * @returns {Promise} - Returns Boolean.
+   */
+
+  hasCoinByAccount(acct, hash, index) {
+    assert(typeof acct === 'number');
+
+    return this.bucket.has(layout.C.encode(acct, hash, index));
+  }
+
+  /**
    * Get hashes of all transactions in the database.
    * @param {Number} acct
    * @returns {Promise} - Returns {@link Hash}[].
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -2626,15 +2626,21 @@ class Wallet extends EventEmitter {
    * Make a transfer MTX.
    * @param {String} name
    * @param {Address} address
+   * @param {String|Number} acct
    * @returns {MTX}
    */
 
-  async makeTransfer(name, address) {
+  async makeTransfer(name, address, acct) {
     assert(typeof name === 'string');
     assert(address instanceof Address);
 
     if (!rules.verifyName(name))
       throw new Error('Invalid name.');
+
+    if (acct != null) {
+      assert((acct >>> 0) === acct || typeof acct === 'string');
+      acct = await this.getAccountIndex(acct);
+    }
 
     const rawName = Buffer.from(name, 'ascii');
     const nameHash = rules.hashName(rawName);
@@ -2657,6 +2663,9 @@ class Wallet extends EventEmitter {
     // Is local?
     if (coin.height < ns.height)
       throw new Error(`Wallet does not own: "${name}".`);
+
+    if (acct != null && !await this.txdb.hasCoinByAccount(acct, hash, index))
+      throw new Error(`Account does not own: "${name}".`);
 
     const state = ns.state(height, network);
 
@@ -2696,7 +2705,8 @@ class Wallet extends EventEmitter {
    */
 
   async _createTransfer(name, address, options) {
-    const mtx = await this.makeTransfer(name, address);
+    const acct = options ? options.account : null;
+    const mtx = await this.makeTransfer(name, address, acct);
     await this.fill(mtx, options);
     return this.finalize(mtx, options);
   }
@@ -2876,14 +2886,20 @@ class Wallet extends EventEmitter {
    * Make a transfer-finalizing MTX.
    * @private
    * @param {String} name
+   * @param {String|Number} acct
    * @returns {MTX}
    */
 
-  async makeFinalize(name) {
+  async makeFinalize(name, acct) {
     assert(typeof name === 'string');
 
     if (!rules.verifyName(name))
       throw new Error('Invalid name.');
+
+    if (acct != null) {
+      assert((acct >>> 0) === acct || typeof acct === 'string');
+      acct = await this.getAccountIndex(acct);
+    }
 
     const rawName = Buffer.from(name, 'ascii');
     const nameHash = rules.hashName(rawName);
@@ -2906,6 +2922,9 @@ class Wallet extends EventEmitter {
     // Is local?
     if (coin.height < ns.height)
       throw new Error(`Wallet does not own: "${name}".`);
+
+    if (acct != null && !await this.txdb.hasCoinByAccount(acct, hash, index))
+      throw new Error(`Account does not own: "${name}".`);
 
     const state = ns.state(height, network);
 
@@ -2955,7 +2974,8 @@ class Wallet extends EventEmitter {
    */
 
   async _createFinalize(name, options) {
-    const mtx = await this.makeFinalize(name);
+    const acct = options ? options.account : null;
+    const mtx = await this.makeFinalize(name, acct);
     await this.fill(mtx, options);
     return this.finalize(mtx, options);
   }
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -2768,14 +2768,20 @@ class Wallet extends EventEmitter {
    * Make a transfer-cancelling MTX.
    * @private
    * @param {String} name
+   * @param {String|Number} acct
    * @returns {MTX}
    */
 
-  async makeCancel(name) {
+  async makeCancel(name, acct) {
     assert(typeof name === 'string');
 
     if (!rules.verifyName(name))
       throw new Error('Invalid name.');
+
+    if (acct != null) {
+      assert((acct >>> 0) === acct || typeof acct === 'string');
+      acct = await this.getAccountIndex(acct);
+    }
 
     const rawName = Buffer.from(name, 'ascii');
     const nameHash = rules.hashName(rawName);
@@ -2798,6 +2804,9 @@ class Wallet extends EventEmitter {
     // Is local?
     if (coin.height < ns.height)
       throw new Error(`Wallet does not own: "${name}".`);
+
+    if (acct != null && !await this.txdb.hasCoinByAccount(acct, hash, index))
+      throw new Error(`Account does not own: "${name}".`);
 
     const state = ns.state(height, network);
 
@@ -2831,7 +2840,8 @@ class Wallet extends EventEmitter {
    */
 
   async _createCancel(name, options) {
-    const mtx = await this.makeCancel(name);
+    const acct = options ? options.account : null;
+    const mtx = await this.makeCancel(name, acct);
     await this.fill(mtx, options);
     return this.finalize(mtx, options);
   }
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -3039,14 +3039,20 @@ class Wallet extends EventEmitter {
   /**
    * Make a revoke MTX.
    * @param {String} name
+   * @param {String|Number} acct
    * @returns {MTX}
    */
 
-  async makeRevoke(name) {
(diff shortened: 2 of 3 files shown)


lib/wallet/txdb.js
```diff
@@ -2240,6 +2240,19 @@ class TXDB {
...
   }
 
+  /**
+   * Test whether an account owns a transaction.
+   * @param {Number} acct
+   * @param {Hash} hash
+   * @returns {Promise} - Returns Boolean.
+   */
+
+  acctOwnsTX(acct, hash) {
```
tynes: I don't love this method name. It feels out of style.

pinheadmz: Maybe `acctHasTX()`?

lib/wallet/wallet.js
```diff
@@ -1781,6 +1786,9 @@ class Wallet extends EventEmitter {
       if (!coin)
         continue;
 
+      if (acct && !await this.txdb.acctOwnsTX(acct, hash))
```
tynes: It feels like the `Account` object should have a method called `hasTX`, making the `txdb` have a method `acctOwnsTX` doesn't feel like the best abstraction. Is it possible to return the account index on the `bid`? This would allow for something like:

```js
if (!own)
  continue;

if (acct && acct !== coin.accountId)
  continue;
```

tynes: Also potentially the `coin` could have the account id on it. We need to be careful to not conflate the account id and the bip 44 account index, re: watch only wallet bug.

pinheadmz: We could, but the bids come from txdb serialized like so: 

https://github.com/pinheadmz/hsd/blob/90837d8fe46eade6e45b66cb3acb09aafa21de7c/lib/wallet/txdb.js#L3729-L3773

So adding an acct id to the bid would mean changing that serialization. No problem if you think that's best, but it'd be a breaking change.

pinheadmz: Same thing with txdb `coin`:

`this.own = false;`

We could change that from a bool to a byte (or how ever many bytes) to store account number?

tynes: Is there a way to get account id by txid using the `txdb`? I think thats another solution.

pinheadmz: wallet layout has:
` *   T[account][hash] -> dummy (tx by account)`
... but nothing for _account by tx_

If it did, would that be cleaner than just getting a `has` bool from txdb?

pinheadmz:
- Calling this test `wallet-accounts-reveal-test` and keeping it in its own file for now

- Running test past the `REVEAL` point doesn't seem necessary - at that point only one coin owns a name, so there wouldn't be conflicting account issues.

- RPC and HTTP API endpoints are now covered.

lib/wallet/wallet.js
```diff
@@ -1781,6 +1786,9 @@ class Wallet extends EventEmitter {
       if (!coin)
         continue;
 
+      if (acct && !await this.txdb.hasCoinByAccount(acct, hash, index))
```
wi-ski: I might be slipping - but this is an acct index at this point right? Does that mean it could be 0 - and falsey - for the default account?

wi-ski:
```suggestion
      if (acct != null && !await this.txdb.hasCoinByAccount(acct, hash, index))
```
:thinking: Maybe?

test/wallet-accounts-reveal-test.js
```diff
@@ -0,0 +1,163 @@
...
+  before(async () => {
+    await node.open();
+    await wclient.open();
+
+    wallet = await wdb.create();
+
+    alice = await wallet.createAccount({name: 'alice'});
+    bob = await wallet.createAccount({name: 'bob'});
+
+    aliceMiner = await alice.receiveAddress();
```
wi-ski: Is there a reason you called this alice-"miner"? It being an address - how do you feel about `aliceReceiveAddress`?

test/wallet-accounts-reveal-test.js
```diff
@@ -0,0 +1,163 @@
...
+    bobMiner = await bob.receiveAddress();
```
wi-ski: `bobReceiveAddress` *

test/wallet-accounts-reveal-test.js
```diff
@@ -0,0 +1,163 @@
...
+const wclient = new WalletClient({
+  port: network.walletPort
+});
+
+const {wdb} = node.require('walletdb');
+const name = rules.grindName(10, 20, network);
+
+let wallet, alice, bob, aliceMiner, bobMiner;
+
+async function mineBlocks(n, addr) {
```
wi-ski: #168 introduces a `common` util for mining blocks: https://github.com/handshake-org/hsd/pull/168/files#diff-e169f8c8ca803ee76312e094ba6f17cbR120

I don't know what the convention is for setting up mining in tests - but thought it was worth a mention :bowing_man: man:

test/wallet-accounts-reveal-test.js
```diff
@@ -0,0 +1,163 @@
+/* eslint-env mocha */
```
wi-ski: These test are awesome! QQ about tests around auctions:

Is there a "standard" test setup for testing cases at different phases of an auction?

If not - #168 introduces what could be a good starting (at least for discussion?): https://github.com/handshake-org/hsd/pull/168/files#diff-bd762abfcd16ee1dedbbbe11817ebec1. It sets up an auction, moves it through different phases with nested `describe` blocks.

lib/wallet/wallet.js
pinheadmz: Oops! good catch thanks, will fix.

test/wallet-accounts-reveal-test.js
pinheadmz: This is a good reference, but it operates on HTTP API endpoints and `hs-client` so it is kind of more of an integration test. I'm trying to fix a bug in the `wallet` module here, so I think it makes sense to just isolate that unit.

test/wallet-accounts-reveal-test.js
pinheadmz: This is cool too but again, requires the client which I'm trying to avoid for this patch. The function in my test also has a quick-n-dirty default value for the miner address when I'm not trying to fund a specific wallet account.

test/wallet-accounts-reveal-test.js
pinheadmz: What do you mean here? I'm calling these addresses `__Miner` because they just receive block subsidies to the wallet accounts so they can afford to open name auctions. I'm not sending any funds to those accounts so I don't think calling them `__ReceiveAddress` is an improvement.

test/wallet-accounts-reveal-test.js
pinheadmz: Discussion on next comment.

test/wallet-accounts-reveal-test.js
pinheadmz: mm, never mind you're right :-) updating...

pinheadmz:
Updated to cover sending `UPDATE` from specified account

tynes:
@wi-ski Do these changes fix any issues you are seeing?

wi-ski:
@tynes - Yes sir they do.

wi-ski:
Awesome job @pinheadmz :clap:

wi-ski:
@tynes @pinheadmz - Dusting off this set of changes :) Is there any chance these could still be desired?

wi-ski:
@chjj @boymanjor @pinheadmz @tynes - Can we get eyes on these changes?

pinheadmz:
rebase to  5c09c707bf7c6345e34cd3d3c2113c2c5081bf3a
rebase on master, update resource serialization used in test.

wi-ski:
@pinheadmz - Hey! Ty so much for making these changes <3

lib/wallet/wallet.js
```diff
@@ -2328,13 +2337,18 @@ class Wallet extends EventEmitter {
    * @returns {MTX}
    */
 
-  async makeUpdate(name, resource) {
+  async makeUpdate(name, resource, acct) {
```
boymanjor: Why are we supporting individual accounts for updates, but not for the other covenants involving an owned name? I would think it would be all or nothing w.r.t. to these covenant types. Similarly, should we support individual accounts for redeems?

test/wallet-accounts-auction-test.js
```diff
@@ -0,0 +1,355 @@
...
+
+    for (const bid of walletBids)
+      assert(bid.own);
+
+    assert.strictEqual(node.mempool.map.size, 0);
+  });
+
+  describe('REVEAL', function() {
+    describe('Library methods', function() {
+      it('one tx per account', async () => {
```
boymanjor: nit: many of these `it` blocks do not follow the normal convention i.e. "should..."

lib/wallet/wallet.js
pinheadmz: Good point. I'll investigate adding account arguments to redeems. This shouldn't be an issue for register, transfer, finalize, etc. since it is impossible for more than one account to own a name -- unless we want to accept an account argument just for the sake of warning the user they think the wrong account owns a name.

pinheadmz:
I added account support for reveal and redeem, as well as update. I started having second thoughts about how to approach this for owned names: update, register, transfer, finalize, revoke, since once a name is owned, there is only one account that can send the covenant anyway.

I was thinking maybe those calls should just not accept an account argument and when it comes to filling the mtx, those calls should just force a `account = null` to avoid the "preferred inputs error".

But now I think we should still do this (and I'll have to add 3-4 more commits to this PR) and here's why: a user may want the account that owns a name to ALSO pay the mining fee. This and related coin-selection / privacy between accounts reasons.

SO for update, etc. we will still accept an account parameter but only TWO options are available: the account that owns the name, or null (fee will be paid by any coin in the wallet, and change always goes to default account). If the wrong account is given, that'll be an error.

chjj:
In the future we should consider adding an account index for bids.

pinheadmz:
@chjj yeah some users with lots of activity are reporting long processing time generate REVEALs

pinheadmz:
@chjj @boymanjor @wi-ski 
Thanks for bearing with me on this reviewers. I think we're ready for a final pass here. ALL covenant types are now accounted for (we'll need to update the docs to indicate that account is an option). Tests were majorly refactored and simplified.

pinheadmz:
@boymanjor we have a non-deterministic test in `Auction RPC`, care to re-run CI build? Everything passes locally for me.

wi-ski:
@pinheadmz - Wowza. You're a beast. Thank you for hacking at this.

@chjj's comment inspired me to investigate intro-ing account indices to bid layouts - this: https://github.com/pinheadmz/hsd/pull/2/files

:point_up: Is this in the right 'spirit' for intro-ing accounts to bids? Meaning - appending a new int [here](https://github.com/pinheadmz/hsd/pull/2/files#diff-0242d314a7bf2a92de65ad2dbf57a39cR98)? If so, can keep 'pulling that thread'.

wi-ski:
That PR is targeting these changes *

pinheadmz:
@wi-ski I think the idea behind indexing bids is not just adding an integer to the bids in layout, but adding a new index entirely like `tx by account` but for bids. A PR would also include implantation throughout txdb and wallet to read, write, and search by that index.

wi-ski:
Ok - My wishful thinking was wishful. :upside_down_face:

pinheadmz:
@wi-ski I encourage you to continue looking into this problem! The goal would be: when the wallet goes to make a reveal, instead of iterating through every bid in the wallet, it just looks for the bids in the specified account… How could this be accomplished? 🤔

wi-ski:
Yes sir, Ill continue poking :thinking:

wi-ski:
Ty for the guidance

PR #234, opened 2019-08-06
seeds: add hnscan/easyhandshake nodes to testnet seeds

kilpatty:
Unfortunately looks like some people are having issues connecting to the network currently. This should be a quick fix to help the new potential users coming from hn get onto the network.

### DIFF
--- a/lib/net/seeds/testnet.js
+++ b/lib/net/seeds/testnet.js
@@ -1,6 +1,17 @@
 'use strict';
 
 module.exports = [
+
+  // hnscan - Urkel Labs
+  'ak2hy7feae2o5pfzsdzw3cxkxsu3lxypykcl6iphnup4adf2ply6a@138.68.61.31',
+
+  // easyhandshake.com
+  'ajaqq7jwqrwixmvl64tzi3qfi7ynj3hmmtzmkyubtrljh4mwmh7ie@165.22.151.242',
+
+  // Mark Tyneway
+  'aovaq4vgqt6ttolmgyk4kmaybvpf66xni5xr2jjxjuvcwzda7clzi@45.79.33.75',
+
+  // Original Testnet Seeds
   'aoihqqagbhzz6wxg43itefqvmgda4uwtky362p22kbimcyg5fdp54@172.104.214.189',
   'ajdzrpoxsusaw4ixq4ttibxxsuh5fkkduc5qszyboidif2z25i362@173.255.209.126',
   'ajk57wutnhfdzvqwqrgab3wwh4wxoqgnkz4avbln54pgj5jwefcts@172.104.177.177',


lib/net/seeds/testnet.js
```diff
@@ -1,8 +1,6 @@
 'use strict';
 
 module.exports = [
-  'aoihqqagbhzz6wxg43itefqvmgda4uwtky362p22kbimcyg5fdp54@172.104.214.189',
-  'ajdzrpoxsusaw4ixq4ttibxxsuh5fkkduc5qszyboidif2z25i362@173.255.209.126',
-  'ajk57wutnhfdzvqwqrgab3wwh4wxoqgnkz4avbln54pgj5jwefcts@172.104.177.177',
-  'am2lsmbzzxncaptqjo22jay3mztfwl33bxhkp7icfx7kmi5rvjaic@139.162.183.168'
+  'ak2hy7feae2o5pfzsdzw3cxkxsu3lxypykcl6iphnup4adf2ply6a@138.68.61.31',
```
tynes: Add a comment saying which node is which

lib/net/seeds/testnet.js
```diff
@@ -1,8 +1,6 @@
 'use strict';
 
 module.exports = [
-  'aoihqqagbhzz6wxg43itefqvmgda4uwtky362p22kbimcyg5fdp54@172.104.214.189',
```
tynes: Also don't remove these ones, just move them to the bottom and put a comment around them saying they were the original testnet nodes

kilpatty: Will do!

lib/net/seeds/testnet.js
kilpatty: Yeah good call, will add yours as well.

lib/net/seeds/testnet.js
```diff
@@ -1,6 +1,17 @@
...
+
+  // hnscan - Urkel Labs
+  'ak2hy7feae2o5pfzsdzw3cxkxsu3lxypykcl6iphnup4adf2ply6a@138.68.61.31',
+
+  // easyhandshake.com
+  'ajaqq7jwqrwixmvl64tzi3qfi7ynj3hmmtzmkyubtrljh4mwmh7ie@165.22.151.242',
+
+  // Mark Tyneway
+  'aovaq4vgqt6ttolmgyk4kmaybvpf66xni5xr2jjxjuvcwzda7clzi@45.79.33.75',
+
```
tynes: Ideally we have some nodes addressable via IPV6 and Tor as well

pinheadmz:
We should probably double check this before mainnet launch. I imagine there will be a new corresponding testnet at that time as well (or likely just before). Brontide will be optional but for the seed nodes it makes sense to have it for authentication. We just need to make sure the keys and IPs are correct.

pinheadmz:
Closing -- the testnet referred to here is long gone. Unfortunately the current "mainnet" testnet has no hashrate and no nodes and no blocks. So maybe one day we can revisit this if HNS testnet becomes important enough to support permanent seeds.

PR #236, opened 2019-08-15
Wallet Name Alerts

tynes:
A desirable feature is to receive a notification when the next auction period starts. This PR creates that functionality by adding a new index to the wallet database that maps a `uint32` (height) -> `nameHashes`. When a new block comes in, the name hashes are looked up in the new index and the wallets that know about those names emit events over websocket. When name state is being indexed, the height of its next period is calculated and used as the key to index the name hash. 

This functionality could be hidden behind a feature flag to prevent the need for additional indexing.
This is useful for automated auction runners.

### DIFF
--- a/lib/wallet/client.js
+++ b/lib/wallet/client.js
@@ -13,7 +13,7 @@ const NameState = require('../covenants/namestate');
 
 const parsers = {
   'block connect': (entry, txs) => parseBlock(entry, txs),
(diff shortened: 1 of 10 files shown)


pinheadmz:
I think this is a useful feature as well but I don't think using the DB is the right approach. Might actually be better as a plugin that gets current auctions and emits events. We could leave this open still as a reminder that its a feature request.

Issue #202, opened 2019-06-19
On Adding Schnorr Signatures


tynes:
Consider adding or switching to Schnorr Signatures?

There are a few ways to do this, Bitcoin Cash went with overloading opcodes depending on the shape of the signature. ESDCA if the signature is DER encoded, otherwise use Schnorr which is expected to be in rs format.

Another way to do it would be to use the size of the address data. A 33 byte address data (similar to [bip-taproot](https://github.com/sipa/bips/blob/bip-schnorr/bip-taproot.mediawiki#cite_ref-1-0)) could alter the signature related opcodes to use Schnorr instead of ECDSA. The 33 bytes would be a public key, not a public key hash or script hash. It would add a new script type that no longer has the `OP_DUP OP_BLAKE_160 0x14... OP_EQUALVERIFY` policy as the witness would no longer need to provide a public key. This would cut down on witness data along with mempool gossip bandwidth. The witness could just be the signature in rs format. If Schnorr Signatures are added in a new witness version, it decreases the anonymity set which decreases fungibility. It does create another type of address which decreases the anonymity set for other address types in witness version 0. Things can easily fork from current Bitcoin software into supporting Handshake by continuing to support ECDSA.

Its also possible to just add schnorr related opcodes and use it in the script for the 33 byte address data. I like that more than overloading.

Another way is removing ECDSA completely and just using Schnorr. This would require any wallet software to be rewritten, including the [ledger application](https://github.com/boymanjor/ledger-app-hns).

chjj:
Yeah, I've been considering this as well. We have 4 well tested versions of bip-schnorr now:

- JS: https://github.com/bcoin-org/bcrypto/blob/master/lib/js/schnorr.js
- OpenSSL: https://github.com/bcoin-org/bcrypto/blob/master/src/ecdsa/ecdsa.c#L1780
- Nettle: https://github.com/bcoin-org/bcrypto/blob/nettle/src/ecdsa/ecdsa.c#L1684
- libsecp256k1: https://github.com/bcoin-org/bcrypto/blob/master/src/secp256k1.cc

[...]

PR #238, opened 2019-08-18
Switch to Schnorr

tynes:
Switch to using `bip-schnorr` as defined in [sipa/bip-schnorr](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki) and implemented in [bcrypto](https://github.com/bcoin-org/bcrypto/blob/master/lib/js/schnorr.js).

This pull request simply replaces ECDSA with Schnorr. It does not change the curve (secp256k1).
A complete replacement is the most simple solution as supporting both ECDSA and Schnorr adds complexity - no need for additional opcodes, no need to decide on which is "standard" (use Schnorr or ECDSA when the address data is 20 bytes?), no worry of depending on the weakest link.

This will require changes in https://github.com/boymanjor/ledger-app-hns @boymanjor 

All addresses that currently exist will **still** be valid addresses.

Closes https://github.com/handshake-org/hsd/issues/202

### DIFF
--- a/lib/workers/jobs.js
+++ b/lib/workers/jobs.js
@@ -142,7 +142,7 @@ jobs.signInput = function signInput(tx, index, coin, ring, type) {
  */
 
 jobs.ecVerify = function ecVerify(msg, sig, key) {
-  const result = secp256k1.verify(msg, sig, key);
+  const result = secp256k1.schnorrVerify(msg, sig, key);
   return new packets.ECVerifyResultPacket(result);
 };
 
@@ -156,7 +156,7 @@ jobs.ecVerify = function ecVerify(msg, sig, key) {
  */
 
 jobs.ecSign = function ecSign(msg, key) {
-  const sig = secp256k1.sign(msg, key);
+  const sig = secp256k1.schnorrSign(msg, key);
   return new packets.ECSignResultPacket(sig);
 };
 
--- a/lib/primitives/keyring.js
+++ b/lib/primitives/keyring.js
@@ -465,7 +465,7 @@ class KeyRing extends bio.Struct {
 
   sign(msg) {
     assert(this.privateKey, 'Cannot sign without private key.');
-    return secp256k1.sign(msg, this.privateKey);
+    return secp256k1.schnorrSign(msg, this.privateKey);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


lib/workers/jobs.js
```diff
@@ -142,7 +142,7 @@ jobs.signInput = function signInput(tx, index, coin, ring, type) {
  */
 
 jobs.ecVerify = function ecVerify(msg, sig, key) {
```
pinheadmz: No need to change the name of this function?

tynes: No, its still elliptic curve verify

chjj:
75% on board with this. I feel like I was 100% on board a few days ago, but now we'll be losing compatibility with BTC. =/

rojii:
In my humble opinion it's best to let schnorr run in the wild before adding it to handshake (although this impression goes under all of the previous knowledge / understanding of each individual distinctly).  I understand that it's a better algorithm, porting it over in the future to the handshake project isn't much of a concern there's plenty of ways to add it to handshake, some of which are kind of bizarre... unless the ultimatum is to completely remove DER encoded signatures from hsd' I'm not sure what to propose to the code to make this happen since it is quite simple to add since we aren't writing the algorithm (@chjj  is). But actually seeing the schnorr algorithm in the wild for other cryptocurrency projects that handle money will perhaps gauge the decision into a better direction . I'm no psychic, but in like manner in what @tynes  has done it doesn't hurt to add multiple implementations of adding it with / without DER signatures to atleast accelerate the process.

If we can be on the same page I got no issues adding some test vectors for checksig / checkmultisig but as I said I think we all agree this can go either way and perhaps the more rational decision is to wait it out for the algorithm to be in the wild a bit longer.

tynes:
Is public key recovery possible with bip-schnorr? This stackoverflow post says no: https://crypto.stackexchange.com/questions/60825/schnorr-pubkey-recovery

This would break the RPC `verifymessage` since it is based on public key recovery.
It would also break the replacement schemes of `OP_CHECKOUTPUT` involving the public key recovery.

tynes:
> Is public key recovery possible with bip-schnorr? This stackoverflow post says no: https://crypto.stackexchange.com/questions/60825/schnorr-pubkey-recovery
> 
> This would break the RPC `verifymessage` since it is based on public key recovery.
> It would also break the replacement schemes of `OP_CHECKOUTPUT` involving the public key recovery.

@chjj ^^^

pinheadmz:
Closing for now, schnorr may be adopted in the future by a soft fork, but it will not be rolled out in address version `0`.

PR #244, opened 2019-08-20
events: websocket event on chain reorganize

tynes:
Add a websocket event `chain reorganize`. The node client can `bind` to this event after doing a `call` to `watch chain`. This is useful for letting clients get a notification when there has been a reorg event on the `chain`. Currently, there is a `reset` event, but it is only called when there is a rescan caused by the user. I need to be alerted when there is a reorg.

Will follow up with tests.

See PR in `bcoin`: https://github.com/bcoin-org/bcoin/pull/826

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -680,6 +680,15 @@ class HTTP extends Server {
       this.to('chain', 'chain reset', tip.encode());
     });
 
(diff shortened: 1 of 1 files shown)


pinheadmz:
ACK
@tynes you wanna write a test? If not I can get to it, this is a simple PR and seems like an important feature to have for http socket clients.

PR #248, opened 2019-09-10
chaindb: add trigger to force rollback the tree for reorgs

kilpatty:
May or may not be the best way to fix this, just pushing to start some discussion


kilpatty:
This is definitely not a good solution, and it looks like the real fix might be even simpler so just going to close this.

kilpatty:
ping @boymanjor to merge these into testnet4-stable if possible

PR #249, opened 2019-09-17
fix locations of prevBlock and treeRoot within the _raw block shared …

alexsmith540:
…by peers used in memblock.js abstraction of padding method.

substring locations used in memblock.js .padding() method are incorrect for powNG _raw block format. This bug didnt cause an issue when mining on one single node (because it uses .padding method in abstractblock.js), however this becomes a huge problem when you try to sync with a node and the peer's blocks received dont hash properly on the local node and the local node gets lots of 'unrequested blocks' errors and never syncs.

How to verify the bug: Create 2 simnet nodes, both on #pow-ng. On miner node: setgenerate=true for a hot second to mine some blocks. Then stop mining because logs. Now on the second node, add miner node as a peer and watch the 'unrequested blocks' errors fly in. Even more fun: Turn setgenerate back on in the miner node while watching peer's logs and watch the orphan blocks fly in.

### DIFF
--- a/lib/primitives/memblock.js
+++ b/lib/primitives/memblock.js
@@ -68,8 +68,8 @@ class MemBlock extends AbstractBlock {
     assert((size >>> 0) === size);
 
     const pad = Buffer.alloc(size);
-    const prevBlock = this._raw.slice(4, 4 + 32);
-    const treeRoot = this._raw.slice(100, 100 + 32);
+    const prevBlock = this._raw.slice(12, 12 + 32);
+    const treeRoot = this._raw.slice(44, 44 + 32);
 
     for (let i = 0; i < size; i++)
       pad[i] = prevBlock[i % 32] ^ treeRoot[i % 32];
--- a/test/block-test.js
+++ b/test/block-test.js
@@ -4,19 +4,65 @@
 
 'use strict';
 
-const {BloomFilter} = require('bfilter');
 const assert = require('bsert');
-const common = require('./util/common');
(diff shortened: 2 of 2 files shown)


pinheadmz:
Phew! Excellent catch. I wrote a test for this case here:

https://github.com/pinheadmz/hsd/commit/655eac4dc01e10950defeb4d03a13934861f044c

Test will fail without this patch.

tynes:
@alexsmith540 Thanks for the PR! Do you mind reformatting the commit message so that it better matches the repo style?

`package name: quick description`

If you want to add a couple paragraphs in the commit message explaining the change, that would be helpful as well.

Could you also cherry pick https://github.com/pinheadmz/hsd/commit/655eac4dc01e10950defeb4d03a13934861f044c into your PR?

alexsmith540:
@tynes @pinheadmz Thanks much for the feedback and the tests! I went ahead and merged those tests in and squashed into one nice commit + message, word.

tynes:
This was fixed in https://github.com/handshake-org/hsd/commit/953b60fc2dbb62ad3063149a580f4d8bdf2e8c14 but the tests are still useful. I think that creating a block `fromOptions` instead of mining it would help to reduce the number of dependencies in the test, so that the correctness of the miner isn't tied to the correctness of the test, since its a block test.

tynes:
Closing for now since this is a PR on a branch that was merged, feel free to reopen on `master` with updated tests if you would like.

Issue #246, opened 2019-08-28
Compile error


zichenglang:
## Env
- System : macOS Mojave 10.14.6
- NPM : 6.11.2
- Node.js : v12.7.0
- Python : 2.7.10

## Step
Follow the instructions exactly
```
$ git clone git://github.com/handshake-org/hsd.git
$ cd hsd
$ npm install --production
```
## Error
When I run command `npm install --production`, I got this

```log
  CXX(target) Release/obj.target/bcrypto/src/aead.o
../src/aead.cc:39:68: error: too few arguments to function call, single argument 'context' was not specified
  target->Set(Nan::New("AEAD").ToLocalChecked(), ctor->GetFunction());
                                                 ~~~~~~~~~~~~~~~~~ ^
/Users/lacdon/Library/Caches/node-gyp/12.7.0/include/node/v8.h:5955:3: note: 'GetFunction' declared here
  V8_WARN_UNUSED_RESULT MaybeLocal<Function> GetFunction(
  ^
/Users/lacdon/Library/Caches/node-gyp/12.7.0/include/node/v8config.h:351:31: note: expanded from macro 'V8_WARN_UNUSED_RESULT'
#define V8_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
                              ^
1 error generated.
make: *** [Release/obj.target/bcrypto/src/aead.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:196:23)
gyp ERR! stack     at ChildProcess.emit (events.js:203:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
gyp ERR! System Darwin 18.7.0
gyp ERR! command "/usr/local/Cellar/node/12.7.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/lacdon/github/hsd/node_modules/bcrypto
gyp ERR! node -v v12.7.0
gyp ERR! node-gyp -v v5.0.3
gyp ERR! not ok
Build failed.

> mrmr@0.1.8 install /Users/lacdon/github/hsd/node_modules/mrmr
> node-gyp rebuild || echo 'Build failed.'

[...]

PR #251, opened 2019-09-24
package: update bcrypto

boymanjor:
See #246.


tynes:
@boymanjor This can be closed

PR #252, opened 2019-09-24
Dockerfile: gmp-dev is required by bcrypto at runtime

jacobhaven:
Without GMP in the base image, attempting to run the image leads to `Error loading shared library libgmp.so.10` which gets suppressed and reported confusingly as `Bindings for bcrypto were not built`


boymanjor:
LGTM. @jacobhaven do you mind back porting this change to the master branch as well?

PR #253, opened 2019-09-25
Sync master with testnet4-stable

jacobhaven:
@boymanjor I think this is the easiest way to continually bring in bugfixes`testnet4-stable`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /code
 CMD "hsd"
 
 RUN apk upgrade --no-cache && \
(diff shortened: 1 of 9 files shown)


jacobhaven:
Ah, but the README conflicts and needs a manual merge.

boymanjor:
Also, we don't want the additional seeds on `master`. I can cherry pick your commit onto `master`.

jacobhaven:
Okay, I'll decline this. I also don't particularly care is the Dockerfile works on master.

Issue #255, opened 2019-09-26
Blacklist should include homographs of ICANN TLDs


jacobhaven:
We expect Handshake names to be resolved, trusted, and displayed in an equivalent way to regular ICANN TLDs.
We also support registration of [IDNA](https://en.wikipedia.org/wiki/Internationalized_domain_name) (punycode) names.

Because of this, simply blacklisting exact string matches of ICANN TLDs is not sufficient. It's extremely easy to conduct [IDN homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack) by registering a punycode domain name that visually resembles (or is identical to) an ICANN TLD.

Depending on how you configure the "confusable" homoglyph database, this would involve blacklisting _at least_ 100s of millions of names (my broad search found ~3 billion or ~92 GiBs of hash data). Initially I hoped, we could store everything in a static data structure committed to in the genesis block but it's probably necessary to just commit to a description of the blacklist grammar for and use that for dynamic detection of homographs collisions when handling (OPEN) TXOs.

My current work on enumerating the homographs is here: https://github.com/namebasehq/idn-homographs

I'll also write some much simpler code to detect homographs, but I need guidance on exactly how we should encode that algorithm in consensus.

Out-of-scope: 
1. Homographs of other (non-TLD) reserved names.
2. Similar attacks on **Handshake** names that have been registered.

jacobhaven:
@boymanjor

tynes:
I agree that blacklisting homographs is a way to protect users, but I am not sure if doing it at the level of blockchain consensus is the correct way to do it. That means that changing the blacklist is likely a hardfork, which adds a lot of risk to the system as a contentious hardfork will split the namespace and make the system much less usable as a whole.

I think that blacklists should be implemented at the policy level, which means that different versions of the software can implement different blacklists and the blacklists can change over time. The cost for a user to change blacklists should be very low, so that competition drives the best blacklist to the top. If a user cannot convince their software or service provider to change the blacklist, then they should be able to easily run their own node with a new blacklist or change to a different service provider.

[...]

PR #256, opened 2019-09-26
[consensus] gTLDs are always reserved

pinheadmz:
After a discussion with @jacobhaven, a good deal of concern was brought up that root zone TLDs like `com`, `org`, and `info` could be auctioned and owned by users after the claim period ends (4 years on mainnet).

I'm not sure I entirely agree with this position, but came up with this patch to implement the rule and maybe we can have public discussion about it here.

### DIFF
--- a/lib/covenants/rules.js
+++ b/lib/covenants/rules.js
@@ -368,10 +368,15 @@ rules.isReserved = function isReserved(nameHash, height, network) {
   if (network.names.noReserved)
     return false;
 
+  const name = reserved.get(nameHash);
+  // gTLDs are always reserved, forever.
+  if (name && name.root)
+    return true;
+
   if (height >= network.names.claimPeriod)
     return false;
 
-  return reserved.has(nameHash);
+  return Boolean(name);
 };
 
 /**
--- a/test/reserved-test.js
+++ b/test/reserved-test.js
@@ -3,6 +3,7 @@
 const assert = require('bsert');
 const fs = require('bfile');
 const reserved = require('../lib/covenants/reserved');
+const rules = require('../lib/covenants/rules');
 
 describe('Reserved', function() {
   it('should get a domain', () => {
@@ -94,5 +95,36 @@ describe('Reserved', function() {
     }
 
     assert.strictEqual(total, 203999999937640 - (10200000 * 1e6));
+  });
+
+  it('should always reserve a root TLD', async () => {
+    const network = {
+      names: {
+        noReserved: false,
+        claimPeriod: 10
+      }
+    };
+
+    // Reserved names are only reserved for a limited time
+    assert(rules.isReserved(
+      rules.hashString('facebook'),
+      5,
+      network));
+
+    assert(!rules.isReserved(
+      rules.hashString('facebook'),
+      50,
+      network));
+
+    // Root names (gTLDs) are always reserved
+    assert(rules.isReserved(
+      rules.hashString('com'),
+      5,
+      network));
+
+    assert(rules.isReserved(
+      rules.hashString('com'),
+      50,
+      network));
   });
 });


tynes:
I feel like allowing anybody to register `com` after 4 years could result in a lot of scams. Even if `com` is blacklisted by the mainline software at the DNS resolver level, its possible to change the blacklists, so users would need to run their fullnodes and also run signed releases. I think its more likely that Verisign doesn't claim `com` within 4 years. Also think about TLDs that do not have the infrastructure in place to claim their names because they do not support DNSSEC - its possible that they do not have the infrastructure ready in 4 years.

kilpatty:
Follow up from Telegram chat - 

While I agree that it's pretty unlikely Verisign or many TLD owners claim their names within 4 years, I don't know if making the TLDs permanently reserved is a good solution to this. 

There is a possibility Handshake fails to find any product market fit in the DNS/Root Zone ecosystem, but does find a fit in something completely unrelated. If we make DNS specific preferences permanent, those will still exist should the community no longer use Handshake for DNS. 

As far as I'm aware, we can change the reserved time with a soft fork? If we get 4 years down the road, and Handshake hasn't either A. attracted enough TLD owners to claim or B. Found a fit in some other use case, then I would suggest we increase the reserved time at that point. 

That being said, I think 4 years is really too soon to start for the reserved names. I think we should consider increasing that time to start, and go from there. 

I think it may also be a good idea to have 2 reserved times. One for the Alexa top 100k names, and another for the current TLDs/gTLDs. Something along the lines of 5 years for the first and 10 years for the second.

pinheadmz:
> then I would suggest we increase the reserved time at that point.

That would be making the rules _less_ restrictive, a hard fork. We can reduce the `claimPeriod` as a soft fork (older software won't care the claims have stopped early) but not extend it.

kilpatty:
@pinheadmz I thought it would be making the rules more restrictive?

We would be saying you can't open these names for an even longer period of time than previously thought, which to me is a more restrictive change. 

Mining nodes would reject any opens to the names, and the old nodes would still see that as valid (since there would be no opens to the names).

pinheadmz:
Huh yeah between `OPEN`s being allowed and `CLAIM`s being rejected, that value is something that can not be soft-forked in either direction!

pinheadmz:
Huh yeah between OPENs being allowed and CLAIMs being rejected, that value is something that can not be soft-forked in either direction ...?!

jacobhaven:
Having given more thought, if we're not going to protect users at the consensus level, then we should not reserver **any** names (besides the existing blacklist: `example`, `invalid`, `local`, `localhost`, and `test`). 

This would make it clear to users that _all_ of the governance on conflicts actually happens at the resolver/user-agent level, not the blockchain level.

I think it's very _confusing and dangerous_ for HS users to have names treated as a protected class only to have that protection stripped away.

tynes:
> Having given more thought, if we're not going to protect users at the consensus level, then we should not reserver **any** names (besides the existing blacklist: `example`, `invalid`, `local`, `localhost`, and `test`).

Could you clarify? Are you talking about the Alexa Top 100k as well as the claimed trademarks as well as the reserved ICANN names?

jacobhaven:
I'm mainly just talking about the ICANN TLDs. That's the real attack vector.
As I understood it, the Trademark claims don't expire like the other reserved names. They exist on the chain and can't be OPENed. Do they have to be renewed?
And it might make economic sense to temporarily reserve the Alexa 100k to encourage them to claim their names.

jacobhaven:
Thanks @pinheadmz I knew there were different flags, but not how they were all handled.

I think it would be in the best interest of user-safety to no pretend to handle ICANN collisions with the this reserved list (it clearly falls short, and always will).

The user should be able to choose (via their user-agent or DNS resolver) how to prioritize HS or ICANN names, especially when ICANN introduces new TLDs, someone registers a homograph (#255), or a HS name is serving some other deceptive or unwanted content.

pinheadmz:
By design, these consensus rules will not be changed and it will be the responsibility of the user/application layer to handle blacklisting past the consensus layer.

PR #258, opened 2019-09-28
dns/server: allow configurable use of unbound

tynes:
This PR adds a new config option `rs-no-unbound` that is expected to be a boolean. If set to true, it will not use `unbound` and instead use the node.js based `RecursiveResolver`.

Since `initNode` sets listeners on `this.hns`, it is moved to after the call to `initOptions`, since `initOptions` can will overwrite `this.hns` if `noUnbound` is set to true.

### DIFF
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -11,6 +11,7 @@ const IP = require('binet');
 const Logger = require('blgr');
 const bns = require('bns');
 const UnboundResolver = require('bns/lib/resolver/unbound');
+const RecursiveResolver = require('bns/lib/resolver/recursive');
 const RootResolver = require('bns/lib/resolver/root');
 const secp256k1 = require('bcrypto/lib/secp256k1');
 const LRU = require('blru');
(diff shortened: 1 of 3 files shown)


tynes:
@boymanjor This PR is meant to allow for deterministic testing of the DNS server. It is possible that the `unbound` bindings differ slightly from the node.js DNS server implementation and this config option would allow testing of both configurations without the need to delete the `unbound` header files and rebuild `hsd`.

boymanjor:
Minor nit: the commit prefix should use directory structure, i.e. `dns/server`.

tynes:
Updated commit message to use directory structure

lib/dns/server.js
```diff
@@ -535,6 +536,16 @@ class RecursiveServer extends DNSServer {
       this.stubPort = options.stubPort;
     }
 
+    if (options.noUnbound === true) {
```
boymanjor: I know we discussed this out of band, but now I think we should fully validate `noUnbound` here, i.e.:
```js
if (options.noUnbound != null) {
  assert(typeof options.noUnbound === 'boolean')
  if (options.noUnbound) {
    ...
  }
}
```

The motivation is to protect against one-off imports of this module that do not use `bcfg` to validate certain properties of the options hash.

tynes: Good call. Pushed up a fix for this.

PR #259, opened 2019-09-28
test: rename reorg test describe blocks

tynes:
This PR simply updates the names of the `describe` blocks in `test/auction-reorg-test.js` so that they do not conflict with the `describe` blocks in `test/auction-test.js`. This prevents developer confusion when running tests, as the names of the `describe` blocks are printed to the console by the test runner. Developers will no longer potentially be thrown off and waste time looking in the wrong test file.

Note that no actual test functionality is modified, re:

```
// DO NOT TOUCH THESE TESTS
// They trigger the tree interval reorg bug.
```

### DIFF
--- a/test/auction-reorg-test.js
+++ b/test/auction-reorg-test.js
@@ -65,10 +65,10 @@ function createNode() {
   };
 }
 
-describe('Auction', function() {
+describe('Auction Reorg', function() {
   this.timeout(15000);
 
-  describe('Vickrey Auction', function() {
+  describe('Vickrey Auction Reorg', function() {
     const node = createNode();
     const orig = createNode();
     const comp = createNode();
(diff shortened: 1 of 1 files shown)


test/auction-reorg-test.js
```diff
@@ -395,7 +395,7 @@ describe('Auction', function() {
     });
   });
 
-  describe('Claim', function() {
+  describe('Claim Reorg', function() {
```
pinheadmz: I could be wrong here but I do not think this test re-org's the chain. In fact I think this test is a complete unaltered copy of the `Claim` test in the original file `auction-test.js` and could probably just be removed.

boymanjor: ACK, but let's heed JJ's warning for now. This PR only changes the names of the describe blocks and leaves program execution as-is. We can open a new PR if this test is safe to remove.

PR #261, opened 2019-09-30
node/rpc: getdnssecproof

tynes:
Add a `getdnssecproof` RPC method that accepts three arguments (two of which are optional), a domain name, an estimate `boolean` and a verbose `boolean`. Setting `estimate` to false will cause an error if the full proof cannot be made. Setting `verbose` to `false` will result in raw hex being returned, otherwise JSON is returned. This follows the same pattern as in the `bitcoind` RPC, see `getblock`.

This method uses the built in DNS resolver to query out to 8.8.8.8 or 8.8.4.4 and builds a DNSSEC proof.

Try it against any of the names in `lib/dns/tld.json`.

For example:

```$
$ ./bin/cli rpc getdnssecproof brave.com
```

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -37,6 +37,7 @@ const pkg = require('../pkg');
 const rules = require('../covenants/rules');
 const {Resource} = require('../dns/resource');
 const NameState = require('../covenants/namestate');
+const ownership = require('../covenants/ownership');
 const AirdropProof = require('../primitives/airdropproof');
 const {EXP} = consensus;
 const RPCBase = bweb.RPC;
@@ -237,6 +238,7 @@ class RPC extends RPCBase {
     this.add('getnameinfo', this.getNameInfo);
     this.add('getnameresource', this.getNameResource);
     this.add('getnameproof', this.getNameProof);
+    this.add('getdnssecproof', this.getDNSSECProof);
     this.add('getnamebyhash', this.getNameByHash);
     this.add('grindname', this.grindName);
     this.add('sendrawclaim', this.sendRawClaim);
@@ -2405,6 +2407,24 @@ class RPC extends RPCBase {
       key: key.toString('hex'),
       proof: proof.toJSON()
     };
+  }
+
+  async getDNSSECProof(args, help) {
+    if (help || args.length < 1 || args.length > 3)
+      throw new RPCError(errs.MISC_ERROR,
+        'getdnssecproof "name" ( estimate ) ( verbose )');
+
+    const valid = new Validator(args);
+    const name = valid.str(0);
+    const estimate = valid.bool(1, false);
+    const verbose = valid.bool(2, true);
+
+    const proof = await ownership.prove(name, estimate);
+
+    if (!verbose)
+      return proof.toHex();
+
+    return proof.toJSON();
   }
 
   async getNameByHash(args, help) {


pinheadmz:
Interesting that it only proves TLDs. On-chain claims involve recursion all the way down. I tried both `bitcoin` and `bitcoin.com`, the response is... unexpected!

```
--> bin/cli rpc getdnssecproof bitcoin.com
Warning: The `hsd cli` interface is deprecated.
Please use `hsd-cli` ($ npm install hs-client).
No DS records for bitcoin.com.
--> bin/cli rpc getdnssecproof bitcoin
Warning: The `hsd cli` interface is deprecated.
Please use `hsd-cli` ($ npm install hs-client).
No DS records for bitcoin.
```

Corresponding logs:
```
[debug] (rpc) Handling RPC call: getdnssecproof.
[debug] (rpc) [ 'bitcoin' ]
[error] (rpc) RPC internal error.
[error] (rpc) No DS records for bitcoin.
[error] (node) No DS records for bitcoin.
```

tynes:
Updated to expose the `estimate` argument to `ownership.prove`

pinheadmz:
Ok I think this is cool and useful to help users claim reserves names. I'm not sure if it fits into this PR, but I also wish it kinda checked for "weak" claims.

I've been in contact with a user trying to claim a name that has a `SHA1` in a DS record somewhere in the chain:

### Get DNSSEC proof: no errors

```
$ hsd-rpc getdnssecproof keys-catalog-softwares.gdn
{
  "zones": [
    ...
      "referral": [
        {
          "name": "gdn.",
          "ttl": 86400,
          "class": "IN",
          "type": "DS",
          "data": {
            "keyTag": 51082,
            "algorithm": 8,
            "digestType": 1,
            "digest": "0bcd14f8566a49f2ed28c3b9e4f95c11f4e622e1",
            "algName": "RSASHA256",
            "hashName": "SHA1"
          }
        },
    ...
    {
      ...
      "claim": [
        ...
        {
          "name": "keys-catalog-softwares.gdn.",
          "ttl": 86400,
          "class": "IN",
          "type": "TXT",
          "data": {
            "txt": [
              "hns-claim:aakbmljljwzbndtw27mkjirvxz5xokse5mz755bhaiaaaaaaaaaklzaororjdpl6qzexi75h7ofhv447lownw5btzuxvs4ibaaaaaxnxfina"
            ]
          }
        },
        ...
      ]
    }
  ]
}
```

### Send actual claim: error

```
$ hsw-rpc sendclaim keys-catalog-softwares
Proof failed signature check.
```

pinheadmz:
Ok I think this looks really cool actually and helpful. IMO, `estimate` should be `false` by default.

I've been helping a user claim `keys-catalog-softwares.gdn` and they have their claim in a TXT record and all that, but there is a DS record with SHA1 somewhere in the chain. I think the most helpful API response for this user would be "Error: Insecure signature" or "Error: Invalid algorithm" but that might be out of scope for this PR.

It might make more sense to explain what `estimate` really means in https://github.com/handshake-org/handshake-org.github.io/pull/66 and provide examples of why a proof might fail in the docs. Then that section of the api-docs could be an important step in the claim process, and we could add it to the guides on hsd-dev.org and https://gist.github.com/tynes/230f715c9710a089ee190b28585b6596

tynes:
> `estimate` should be `false` by default

Agree with this now that you mention it, makes more sense to fail by default since the proof is going into creating a transaction and we don't want to give a sense of security to people by having this RPC method "just work".

I agree that the error messages would be useful, but that would involve a change to `chjj/bns` or duplicating code.

Thank you for the review!

tynes:
@pinheadmz The estimate param now defaults to `false`

tynes:
Pushed updates to the docs at https://github.com/handshake-org/handshake-org.github.io/pull/66

PR #97, opened 2019-02-05
Wallet: require public key for watchonly 

pinheadmz:
This is a cherry-pick of https://github.com/bcoin-org/bcoin/pull/639

> Addresses #636 in which a user creates private keys they could never access due to a wallet being watch-only.
> 
> This PR goes along with #637 and [bcoin-org/bclient#15](https://github.com/bcoin-org/bclient/pull/15), maybe we can decide if we need all three safeties in place or which ones we do or don't need.
> 
> In this pull request, we prevent a watch-only wallet from ever generating a master (private) key which could lead to a user seeing addresses they can not spend from.
> 
> I think `watchOnly` should always require an `accountKey`. The reverse isn't necessarily required, for example in `bclient`, creating a wallet with `accountKey` automatically sets `watchOnly` to `true`. In bcoin `wallet._createAccount()`, `accountKey` is just ignored if `watchOnly` is false, but at least that doesn't lead to unrecoverable keys!

**Note in `hsw-cli` the params for `watch-only` and `account-key` (in bcoin/bwallet-cli) are `watch` and `key` (in hsw-cli)**

Example:
```
$ hsw-cli mkwallet --id=watchtest --watch=true
Error: Must add HD public keys to watch only wallet.
    at WalletClient.request (/Users/matthewzipkin/Desktop/work/hs-client/node_modules/bcurl/lib/client.js:213:19)
    at process._tickCallback (internal/process/next_tick.js:68:7)

$ hsw-cli mkwallet --id=watchtest --watch=true --key=tpubDDuC5zidzkFVJSETSQnpzNv5yESw3TKSAJsYmYT1XaLYmr7kZa3bZ7CFKSQGb2E9AmktZf1QMj9FuxmBEDZ6U5TAYNLtgfZcqQ
uLqJhoZ6c
{
  "network": "testnet",
  "wid": 6,
  "id": "watchtest",
  "watchOnly": true,
  "accountDepth": 1,
  "token": "69a02638a83dae00d6583974fc7862a9ec3afc85d258e86601a678cc631f287d",
  "tokenDepth": 0,
  "master": {
    "encrypted": false
  },
  "balance": {
    "account": -1,
    "tx": 0,
    "coin": 0,
    "unconfirmed": 0,
    "confirmed": 0,
    "lockedUnconfirmed": 0,
    "lockedConfirmed": 0
  }
}
```

PR #262, opened 2019-10-01
Create txdb credit for known key paths when confirming

pinheadmz:
This is a port of https://github.com/bcoin-org/bcoin/pull/829

Refer to original post for summary and discussion.

Note the hsd wallet has one extra detail in the `txdb.confirm()` function, it adds coins from the tx input to a `view`, which gets passed to `connectNames()` -- clearly not necessary in bcoin.

Since this PR only affects the handling of tx **_outputs_**, we should be ok. But reviewers more familiar with hsd might wanna double check me on that.

The reason we don't need to do anything for the inputs was discussed offline:

When txdb confirms a tx, it assumes that all input coins are confirmed. In fact, this is consensus for obvious reasons. If it sees an input it doesn't own, it ignores it. Even if the tx gets unconfirmed and re-confirmed, txdb will not recognize that credit or create a new credit. That makes sense because if we want to get that output/credit/coin into the txdb - we *must* go back and txdb.confirm() the original tx that created it, and I believe that is the case the PR already addresses. We wouldn't really be able to create a new coin at this point anyway, since all we know if the outpoint (hash/index), not the script or sender address or value...

### DIFF
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -18,11 +18,7 @@ const KeyRing = require('../lib/primitives/keyring');
 const Input = require('../lib/primitives/input');
 const Outpoint = require('../lib/primitives/outpoint');
 const Script = require('../lib/script/script');
-const PrivateKey = require('../lib/hd/private.js');
 const policy = require('../lib/protocol/policy');
-const HD = require('../lib/hd');
-const Wallet = require('../lib/wallet/wallet');
-const nodejsUtil = require('util');
 const HDPrivateKey = require('../lib/hd/private');
 
 const KEY1 = 'xprv9s21ZrQH143K3Aj6xQBymM31Zb4BVc7wxqfUhMZrzewdDVCt'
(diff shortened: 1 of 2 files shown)


test/wallet-test.js
```diff
@@ -18,7 +18,7 @@ const KeyRing = require('../lib/primitives/keyring');
 const Input = require('../lib/primitives/input');
 const Outpoint = require('../lib/primitives/outpoint');
 const Script = require('../lib/script/script');
-const PrivateKey = require('../lib/hd/private.js');
+const HDPrivateKey = require('../lib/hd/private');
```
pinheadmz: lol - this `.js` was leftover from an earlier port https://github.com/handshake-org/hsd/pull/97 I guess I opened the PR on hsd before it was finished on bcoin!

tynes:
Hey @pinheadmz, looks like this needs a rebase

pinheadmz:
@tynes thanks, pushed.

PR #263, opened 2019-10-01
wallet: better error for send() if no outputs given

pinheadmz:
port of https://github.com/bcoin-org/bcoin/pull/664

Just changing the error message to make more sense:

`No outputs available.` --> `At least one output required.`


boymanjor:
@pinheadmz you mind rebasing this?

pinheadmz:
@boymanjor done, thanks.

Issue #264, opened 2019-10-06
DNSSEC doesn't sign Handshake Onion records


tynes:
Since `Resource.toA` can populate the `Message` (seen as `res` below) with multiple record types, both `A` and `TXT` in the case of of using the built in Onion functionality, it doesn't sign the `TXT` records due to the second argument to `key.signZSK` meaning "commit to this type of record". It is currently hardcoded to `A` records, as seen below. If there are `TXT` records populated in the response (hsd/lib/resource line 472) then they will not be authenticated.

https://github.com/handshake-org/hsd/blob/7c172277476ea2a41501413f4f43a4a0a040b100/lib/dns/resource.js#L1204-L1205

https://github.com/handshake-org/hsd/blob/7c172277476ea2a41501413f4f43a4a0a040b100/lib/dns/resource.js#L454-L473

This is also a problem with `Resource.toAAAA`, as it has similar logic to `Resource.toA`.

pinheadmz:
I'm having trouble grokking this - could you give an example or test with expected and actual behavior?

PR #265, opened 2019-10-08
test: add dns resolver tests

tynes:
This PR includes a test that purchases names and then writes one of each of the different Handshake records into the tree. A DNS stub resolver then queries for the name and asserts on the correct data being returned.

The DNSSEC test tests that all resource records sets in the answer and authority sections are signed. If all resource record sets in the additional section were signed as well, it could cause problems because there could potentially be many signatures and would force the DNS server to fall back to TCP. Ideally a different authentication mechanism is devised, because the DNSSEC key is public information anyways.

@boymanjor happy to split this up into smaller PRs if it would make it easier to review

Closes https://github.com/handshake-org/hsd/issues/264

### DIFF
--- a/lib/dns/records.js
+++ b/lib/dns/records.js
@@ -104,6 +104,39 @@ const typesByVal = {
 };
 
 /**
+ * DNS RR Query Types by Handshake
+ * record type. This allows a lookup
+ * to determine which kind of DNS
+ * query to send based on which type
+ * of Handshake record would like
+ * to be resolved.
+ * @constant
+ */
+
+const dnsByType = {
+  [types.INET4]: wire.types.A,
+  [types.INET6]: wire.types.AAAA,
+  [types.ONION]: wire.types.A,    // encoded as TXT
+  [types.ONIONNG]: wire.types.A,  // encoded as TXT
+  [types.NAME]: wire.types.CNAME,
+  [types.CANONICAL]: wire.types.CNAME,
+  [types.DELEGATE]: wire.types.DNAME,
+  [types.NS]: wire.types.NS,
+  [types.SERVICE]: wire.types.SRV,
+  [types.URI]: wire.types.URI,
+  [types.EMAIL]: wire.types.RP,
+  [types.TEXT]: wire.types.TXT,
+  [types.LOCATION]: wire.types.LOC,
+  [types.MAGNET]: wire.types.URI,
+  [types.DS]: wire.types.DS,
+  [types.TLS]: wire.types.TLSA,
+  [types.SMIME]: wire.types.SMIMEA,
+  [types.SSH]: wire.types.SSHFP,
+  [types.PGP]: wire.types.OPENPGPKEY,
+  [types.ADDR]: wire.types.URI
+};
+
+/**
  * Target
  * @extends {Struct}
  */
@@ -1341,6 +1374,7 @@ function isSingle(label) {
 
 records.types = types;
 records.typesByVal = typesByVal;
+records.dnsByType = dnsByType;
 
 records.Target = Target;
 records.Service = Service;
--- a/lib/dns/records.js
+++ b/lib/dns/records.js
@@ -64,14 +64,14 @@ const types = {
   TEXT: 13, // TXT
   LOCATION: 14, // LOC
   // new record
-  MAGNET: 15, // TXT
+  MAGNET: 15, // URI
   DS: 16, // DS
   TLS: 17, // TLSA
   SMIME: 18, // SMIMEA
   SSH: 19, // SSHFP
   PGP: 20, // OPENPGPKEY
   // new record
-  ADDR: 21 // TXT
+  ADDR: 21 // URI
 };
 
 /**
--- a/lib/dns/records.js
+++ b/lib/dns/records.js
@@ -75,6 +75,35 @@ const types = {
 };
 
 /**
+ * Record Types By Value
+ * @constant
+ */
+
+const typesByVal = {
+  [types.INET4]: 'INET4',
+  [types.INET6]: 'INET6',
+  [types.ONION]: 'ONION',
+  [types.ONIONNG]: 'ONIONNG',
+  [types.NAME]: 'NAME',
+  [types.GLUE]: 'GLUE',
+  [types.CANONICAL]: 'CANONICAL',
+  [types.DELEGATE]: 'DELEGATE',
+  [types.NS]: 'NS',
+  [types.SERVICE]: 'SERVICE',
+  [types.URI]: 'URI',
+  [types.EMAIL]: 'EMAIL',
+  [types.TEXT]: 'TEXT',
+  [types.LOCATION]: 'LOCATION',
+  [types.MAGNET]: 'MAGNET',
+  [types.DS]: 'DS',
+  [types.TLS]: 'TLS',
+  [types.SMIME]: 'SMIME',
+  [types.SSH]: 'SSH',
+  [types.PGP]: 'PGP',
+  [types.ADDR]: 'ADDR'
+};
+
+/**
  * Target
  * @extends {Struct}
  */
@@ -1311,6 +1340,7 @@ function isSingle(label) {
  */
 
 records.types = types;
+records.typesByVal = typesByVal;
 
 records.Target = Target;
 records.Service = Service;
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -1203,13 +1203,13 @@ class Resource extends Struct {
       case types.A:
         res.answer = this.toA(name);
         key.signZSK(res.answer, types.A);
-        if (this.hasTOR())
+        if (this.hasTor())
           key.signZSK(res.answer, types.TXT);
         break;
       case types.AAAA:
         res.answer = this.toAAAA(name);
         key.signZSK(res.answer, types.AAAA);
-        if (this.hasTOR())
+        if (this.hasTor())
           key.signZSK(res.answer, types.TXT);
         break;
       case types.CNAME:
--- a/test/data/resources-v0.json
+++ b/test/data/resources-v0.json
@@ -1,0 +1,206 @@
+{
+  "INET4": [{
+    "type": "A",
+    "resource": {
+      "hosts": [
+        "192.168.0.91",
+        "192.168.1.19"
+      ],
+      "ttl": 64
+    }
+  }],
+  "INET6": [{
+    "type": "AAAA",
+    "resource": {
+      "hosts": ["fc00::"],
+      "ttl": 64
+    }
+  }],
+  "ONION": [{
+    "_NOTE": "Returned as TXT record from A/AAAA query",
+    "type": "A",
+    "resource": {
+      "hosts": ["blockchainbdgpzk.onion"],
+      "ttl": 64
+    }
+  }],
+  "ONIONNG": [{
+    "_NOTE": "Returned as TXT record from A/AAAA query",
+    "type": "A",
+    "resource": {
+      "hosts": [
+        "sik5nlgfc5qylnnsr57qrbm64zbdx6t4lreyhpon3ychmxmiem7tioad.onion"
+      ],
+      "ttl": 64
+    }
+  }],
+  "NAME": [{
+    "type": "CNAME",
+    "resource": {
+      "canonical": "foo.",
+      "ttl": 64
+    }
+  }],
+  "CANONICAL": [{
+    "type": "CNAME",
+    "resource": {
+      "canonical": "foo.bar@1.2.3.4",
+      "ttl": 64
+    }
+  }],
+  "DELEGATE": [{
+    "type": "DNAME",
+    "resource": {
+      "delegate": "bar.foo@4.3.2.1",
+      "ttl": 64
+    }
+  }],
+  "NS": [{
+    "_NOTE": "Includes glue A records",
+    "type": "NS",
+    "resource": {
+      "ns": [
+        "ns1.cloudflare.com@1.2.3.4",
+        "ns2.cloudflare.com@1.2.4.4"
+      ],
+      "ttl": 64
+    }
+  }],
+  "SERVICE": [{
+    "_NOTE": "Includes 2 SRV records, the target is a name or an IP",
+    "type": "SRV",
+    "resource": {
+      "service": [{
+        "service": "a2.",
+        "protocol": "lseed.",
+        "target": "1.2.3.4",
+        "priority": 0,
+        "weight": 5,
+        "port": 7868
+      }, {
+        "service": "a2.",
+        "protocol": "lseed.",
+        "target": "btcseeder",
+        "priority": 0,
+        "weight": 2,
+        "port": 1337
+      }],
+      "ttl": 64
+    }
+  }],
+  "URI": [{
+    "type": "URI",
+    "resource": {
+      "uri": ["ftp://ftp1.example.com/public"],
+      "ttl": 64
+    }
+  }],
+  "EMAIL": [{
+    "type": "RP",
+    "resource": {
+      "email": ["foo.gmail.com"],
+      "ttl": 64
+    }
+  }],
+  "TEXT": [{
+    "type": "TXT",
+    "resource": {
+      "text": ["this is a txt record", "foobar"],
+      "ttl": 64
+    }
+  }],
+  "LOCATION": [{
+    "type": "LOC",
+    "resource": {
+      "location": [{
+        "version": 0,
+        "size": 0,
+        "horizPre": 1,
+        "vertPre": 2,
+        "latitude": 128,
+        "longitude": 256,
+        "altitude": 16
+      }],
+      "ttl": 64
+    }
+  }],
+  "MAGNET": [{
+    "type": "URI",
+    "resource": {
+      "magnet": [
+        "magnet:?xt=urn:btih:8c271f4d2e92a3449e2d1bde633cd49f64af888f"
+      ],
+      "ttl": 64
+    }
+  }],
+  "DS": [{
+    "type": "DS",
+    "resource": {
+      "ds": [{
+        "keyTag": 35215,
+        "algorithm": 13,
+        "digestType": 2,
+        "digest": "7C50EA94A63AEECB65B510D1EAC1846C973A89D4AB292287D5A4D715136B57A3"
+      }],
+      "ttl": 64
+    }
+  }],
+  "TLS": [{
+    "type": "TLSA",
+    "resource": {
+      "tls": [{
+        "protocol": "tcp.",
+        "port": 443,
+        "usage": 0,
+        "selector": 0,
+        "matchingType": 0,
+        "certificate": "0102348790"
+      }],
+      "ttl": 64
+    }
+  }],
+  "SMIME": [{
+    "type": "SMIMEA",
+    "resource": {
+      "smime": [{
+        "hash": "92faa0601d44a3c2b98318d8f7e4cd3b17f25caa6753cc04365d753d",
+        "usage": 0,
+        "selector": 0,
+        "matchingType": 0,
+        "certificate": "85fc1da9c82de277abd69f65b962aa55c335697a8bb8418894b65f24d86c6de4"
+      }],
+      "ttl": 64
+    }
+  }],
+  "SSH": [{
+    "type": "SSHFP",
+    "resource": {
+      "ssh": [{
+        "algorithm": 0,
+        "digestType": 1,
+        "fingerprint": "9be3ba6df1bc232ad4642d33ac09b8dc991e9b76f4f74bc96dd2ff17aa94500d9af2bd8f73a9eee9"
+      }],
+      "ttl": 64
+    }
+  }],
+  "PGP": [{
+    "type": "OPENPGPKEY",
+    "resource": {
+      "pgp": [{
+        "hash": "0e9baa5e5ce8cffd7bf13c37423957ed5e0061c9c8807628979ef475",
+        "publicKey": "b796a94d827d3b3142dde13d945a0467b5c9afa53ae551bdf898b1f3101ef883"
+      }],
+      "ttl": 64
+    }
+  }],
+  "ADDR": [{
+    "type": "URI",
+    "resource": {
+      "addr": [
+        "bitcoin:bc1q6yteaqmza4h7en72pwu4hk952c03wr32xxlp7v",
+        "bitcoin:tb1q6yteaqmza4h7en72pwu4hk952c03wr32vqyj9l"
+      ],
+      "ttl": 64
+    }
+  }]
+}
--- a/test/dns-test.js
+++ b/test/dns-test.js
@@ -1,0 +1,300 @@
+/**
+ * dns-test.js - DNS Server Testing for hsd
+ * Copyright (c) 2019, Mark Tyneway (MIT License).
+ * https://github.com/handshake-org/hsd
+ */
+
+'use strict';
+
+const assert = require('bsert');
+const StubResolver = require('bns/lib/resolver/stub');
+const bcrypto = require('bcrypto');
+const dnssec = require('bns/lib/dnssec');
+const util = require('bns/lib/util');
+const wire = require('bns/lib/wire');
+const FullNode = require('../lib/node/fullnode');
+const Resource = require('../lib/dns/resource');
+const Records = require('../lib/dns/records');
+const Network = require('../lib/protocol/network');
+const NameState = require('../lib/covenants/namestate');
+const rules = require('../lib/covenants/rules');
+const {types} = wire;
+
+const json = require('./data/resources-v0.json');
+
+const network = Network.get('regtest');
+
+const node = new FullNode({
+  network: 'regtest',
+  apiKey: 'foo',
+  walletAuth: true,
+  rsNoUnbound: true,
+  memory: true,
+  workers: true,
+  plugins: [require('../lib/wallet/plugin')]
+});
+
+const nstub = new StubResolver({
+  rd: true,
+  cd: true,
+  edns: true,
+  ednsSize: 4096,
+  maxAttempts: 2,
+  maxTimeout: 3000,
+  dnssec: true,
+  servers: [`127.0.0.1:${network.nsPort}`]
+});
+
+describe('DNS Servers', function() {
+  this.timeout(15000);
+
+  before(async () => {
+    await node.open();
+    await nstub.open();
+  });
+
+  after(async () => {
+    await node.close();
+    await nstub.close();
+  });
+
+  describe('Authoritative Resolver', () => {
+    // write each resource json to the tree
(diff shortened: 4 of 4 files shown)


lib/dns/records.js
```diff
@@ -64,14 +64,72 @@ const types = {
...
+  [types.ADDR]: 'ADDR'
+};
+
+/**
+ * DNS RR Types by Handshake
+ * record type.
+ * @constant
+ */
+
+const rrTypeByType = {
```
tynes: Nit: `rrTypesByType`

Also open to an alternative name to this if there are any suggestions.

lib/dns/resource.js
```diff
@@ -1125,6 +1125,8 @@ class Resource extends Struct {
             for (const rr of this.toGlue(types.SRV))
               res.additional.push(rr);
             key.signZSK(res.answer, types.SRV);
+            key.signZSK(res.additional, types.A);
+            key.signZSK(res.additional, types.AAAA);
```
tynes: `key.signZSK` will throw an error if there are 2 RRs of the same type with a different name. Depending on `toSRVIP` and `toGlue`, I think this is possible. Removing DNSSEC from the `additional` section is one option, or handling that case in `key.signZSK` is another option. Down the codepath of `key.signZSK`, `bns.util.extractSet` is called without a name.

lib/dns/resource.js
```diff
@@ -1224,13 +1228,27 @@ class Resource extends Struct {
...
         key.signZSK(res.additional, types.AAAA);
         break;
-      case types.NS:
+      case types.NS: {
         res.authority = this.toNS(name);
         res.additional = this.toNSIP(name);
         for (const rr of this.toGlue(types.NS))
           res.additional.push(rr);
         key.signZSK(res.authority, types.NS);
+
```
tynes: This logic handles the case when there are multiple names of the same RR type. I think this logic should move inside of `key.signZSK` so that the abstraction here can stay the same. A single `key.signZSK` would be able to append multiple `RRSIG` records if there are multiple rrsets in the same section, ie they have different names but the same rr type (since that is the second argument to `key.signZSK`.

lib/node/rpc.js
```diff
@@ -2366,7 +2366,7 @@ class RPC extends RPCBase {
 
     try {
       const res = Resource.decode(ns.data);
-      return res.toJSON();
```
tynes: This change is required to correctly render the JSON response with the `name` field filled.

lib/dns/records.js
tynes: `hnsTodns`

lib/dns/records.js
```diff
@@ -850,8 +908,8 @@ class SMIME extends Struct {
     assert((json.usage & 0xff) === json.usage);
     assert((json.selector & 0xff) === json.selector);
     assert((json.matchingType & 0xff) === json.matchingType);
-    assert(typeof json.fingerprint === 'string');
-    assert((json.fingerprint.length >>> 1) <= 255);
+    assert(typeof json.certificate === 'string');
```
tynes: TODO: break this into its own PR

lib/dns/server.js
```diff
@@ -48,17 +50,6 @@ const {
 const TYPE_MAP = Buffer.from('000722000000000380', 'hex');
 const RES_OPT = { inet6: false, tcp: true };
 
-const blacklist = new Set([
```
tynes: TODO: break this into its own PR

lib/node/rpc.js
tynes: TODO: break this into its own PR

lib/dns/records.js
tynes: Merged in https://github.com/handshake-org/hsd/commit/0da753980c0899667d52188883ccd728f9355e96

lib/dns/server.js
tynes: Closed, considering different options. Remove this commit from this PR

lib/node/rpc.js
tynes: Merged in https://github.com/handshake-org/hsd/commit/a90a45bb9ea2c267e47bbaa173da1f181123f35f

lib/dns/records.js
```diff
@@ -64,14 +64,72 @@ const types = {
...
+  [types.PGP]: 'PGP',
+  [types.ADDR]: 'ADDR'
+};
+
+/**
+ * DNS RR Types by Handshake
+ * record type.
+ * @constant
+ */
+
```
tynes: Not sure what to call this data structure. It is used to dynamically map between Handshake record type and DNS record type. Any feedback would be appreciated.

tynes: @boymanjor suggested `hnsToDns`

test/dns-test.js
```diff
@@ -0,0 +1,392 @@
...
+    await nclient.execute('generatetoaddress', [1, address]);
+    await common.forValue(obj, 'complete', true);
+  }
+}
+
+/**
+ * Verify DNSSEC
+ */
+
+function verifyDNSSEC(resource, pubkey, qtype, name) {
```
tynes: Refactor this to only work on answer, potentially authority.

lib/dns/resource.js
tynes: Don't sign the additional section here

test/dns-test.js
tynes: This has been completed

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+      additional.data.signature = null;
+  }
+}
+
+/**
+ * Parse the ZSK and KSK from
+ * a record.
+ */
+
+function getSigningKeys(record) {
```
tynes: It is possible to do this without calling `rr.toJSON` and looking at the `keyType`, but it is a lot more simple to do it this way when performance isn't critical.

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+
+  return {
+    zsk: zsk,
+    ksk: ksk
+  };
+}
+
+/**
+ * Build a name to query for the types
+ * that require additional labels.
```
tynes: I'd like to turn this into a helper function as I think I'd find it useful building applications on top of Handshake. Not quite sure where it would live or how to generalize it correctly - notice that it uses a magic index into an array.

lib/dns/resource.js
tynes: > Don't sign the additional section here

This has been applied, if the additional section is signed then the response will be too large and fall back to TCP. The key is public anyways.

For more, see:
https://security.stackexchange.com/questions/91038/dnssec-why-are-rrsigs-returned-in-the-answer-not-additional-section

lib/dns/resource.js
tynes: This is probably good to note for the future, but is no longer relevant.

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+    await node.close();
+    await nstub.close();
+    await rstub.close();
+  });
+
+  describe('Recursive Resolver', () => {
+    it('should return an A record', async () => {
+      const name = 'google.com.';
+
+      const res = await rstub.lookup(name, types.A);
```
pinheadmz: Is this the first time a test in bcoin or hsd actually makes an external internet connection? I wonder what the implications of that are, and what kind of failures could happen with CI. I think `google.com` seems like it will be a safe bet to query for a long time, but it still makes this test rely on an external factor. Do we want to consider a separate testing script like what I did here: https://github.com/bcoin-org/bcoin/pull/759 -- putting the test that requires an internet connection outside the main testing suite, but still running it on CI. Just 🍕 for 🤔 .

tynes: Would agree that removing it is probably a good idea, I added it as the first test in this file to figure out how the APIs work together. Thanks for the feedback. It is a risk of nondeterminism, but if `google.com` doesn't resolve (assuming no CI firewall) there is a major problem happening with the internet

lib/dns/records.js
tynes: I haven't been able to find a better name for this data structure and have began to like its current name. It is very descriptive of what it contains.

test/dns-test.js
pinheadmz: Both `"should return no answer..."` tests also hit the internet. Can be seen by logging the queries in `bns/lib/resolver/dns.js` `exchange()`

pinheadmz:
```
Querying server: 127.0.0.1 (43708) (tcp=false)
[ { name: 'com.', class: 'IN', type: 'A' } ]
Querying server: 192.112.36.4 (18481) (tcp=false)
[ { name: 'com.', class: 'IN', type: 'NS' } ]
      ✓ should return no answer for an ICANN rooted name (125ms)
Querying server: 127.0.0.1 (47554) (tcp=false)
[ { name: 'foo.', class: 'IN', type: 'A' } ]
Querying server: 202.12.27.33 (24311) (tcp=false)
[ { name: 'foo.', class: 'IN', type: 'NS' } ]
      ✓ should return no answer for HNS name pre-update (195ms)
```

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+    it('should return no answer for an ICANN rooted name', async () => {
+      const name = 'com.';
+
+      const res = await nstub.lookup(name, types.A);
+
+      assert.equal(res.question[0].name, name);
+      assert.equal(res.answer.length, 0);
+
+      // Handshake zone is not authoritative for ICANN based name
+      assert.equal(res.aa, false);
```
pinheadmz: Should we test that the nameserver returned an `authority` section from the ICANN lookup?

lib/dns/records.js
tynes: `dnsTypeToType`

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+
+    // record to save in the authenticated tree
+    for (const [hstype, item] of Object.entries(json)) {
+      const type = types[item.type];
+      const resource = Resource.fromJSON(item.resource);
+
+      // args: size, height, network
+      const name = rules.grindName(5, 1, network);
+
+      it(`should update ${name}`, async () => {
```
pinheadmz: I don't think the `${name}` is useful here since it's randomly generated, and each pass of the test will be different.
Perhaps "should update name with record type ${item.type}"

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+
+        const info = await nclient.execute('getnameinfo', [name]);
+        const data = Buffer.from(info.info.data, 'hex');
+
+        const returned = Resource.decode(data);
+        assert.deepEqual(returned.getJSON(name), json);
+
+        // this commits the tree state to disk
+        // updates served by the dns server
+        await mineBlocks(treeInterval, coinbase);
```
pinheadmz: There's a few places around the testing suite where a complete auction process must be run, I think there's also one that uses RPC calls like this. Anyway, that code is _covered_, and is a bit of a waste here, especially since every one of these tests takes ~500ms on my machine. 

I think we should find a simple-unit-test way to just inject a single record into the tree without any of the blockchain BS. This entire test could be 90% lighter. Maybe it's possible to even drop the `FullNode` entirely and just run the StubResolvers, RootServer, and NameServer, with a hacked-in urkel tree.

pinheadmz: Looks like `FullNode` instantiates a `RootResolver` with the option:
`lookup: key => this.chain.db.tree.get(key)`
I wonder if that can just be stubbed with a simple map. All the updates in this test could be simple insertions!

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+        const response = await nstub.lookup(query, type);
+        const res = response.toJSON();
+        const dns = resource.toDNS(util.fqdn(name), type).toJSON();
+
+        // query for the zone signing key
+        const dnskey = await nstub.lookup('.', types.DNSKEY);
+
+        // parse the zone signing key and
+        // key signing key out of the response
+        const {zsk, ksk} = getSigningKeys(dnskey);
```
pinheadmz: Do these need to run on each loop? Or can they be hoisted

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+
+        assert(zsk instanceof wire.Record);
+        assert(ksk instanceof wire.Record);
+
+        // validate the signature on the ZSK
+        verifyDNSSEC(dnskey, ksk, types.DNSKEY, '.');
+        // validate the signature over the rrsets
+        verifyDNSSEC(response, zsk, type, query);
```
pinheadmz: OK I see you are testing the keys here, maybe it should be a separate `it` block, or even run in `before()`

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+/**
+ * Build a name to query for the types
+ * that require additional labels.
+ */
+
+function buildQuery(name, resource, type) {
+  switch (type) {
+    case types.SRV: {
+      const service = resource.service[0];
+      return '_' + service.service
```
tynes: Nit: fix the style here to match the other cases, 1 item per line

test/dns-test.js
tynes: It is testing that the zsk was signed by the ksk. I wouldn't trust that the zsk is correct until I know that it has been signed by the ksk, so that is why I test that it is signed each time. In actuality, it is not possible for the zsk to be man in the middle'ed in this test, so it could be done once. In practice, each request should test this, and as end to end tests I'd like to emulate as close as possible what would happen in reality. If it were more computationally complex I'd definitely agree.

If you ask me to refactor it so that it doesn't bid on a new name each time but instead uses a single name and just updates that name, I'd be happy to do that refactor. It slows down the tests quite a bit, but I didn't want any side effects from previous tests to have a role.

test/data/resources-v0.json
```diff
@@ -0,0 +1,200 @@
...
+          "certificate": "0102348790"
+        }],
+        "ttl": 64
+      }
+    },
+    "SMIME": {
+      "type": "SMIMEA",
+      "resource": {
+        "smime": [{
+"hash": "92faa0601d44a3c2b98318d8f7e4cd3b17f25caa6753cc04365d753d",
```
pinheadmz: nit: indentation?

test/data/resources-v0.json
```diff
@@ -0,0 +1,200 @@
...
+          "hash": "0e9baa5e5ce8cffd7bf13c37423957ed5e0061c9c8807628979ef475",
+          "publicKey": "b796a94d827d3b3142dde13d945a0467b5c9afa53ae551bdf898b1f3101ef883"
+        }],
+        "ttl": 64
+      }
+    },
+    "ADDR": {
+      "type": "URI",
+      "resource": {
+        "addr": ["bitcoin:bc1qdsr2022f90pum2ptd8ftp4nvpgdhk656ue33qz"],
```
pinheadmz: Recommend using a regtest address for this, who knows... This address has history in the blockchain, belongs to somebody.

test/data/resources-v0.json
tynes: Good catch, must have missed it in my editor when it overflowed to the next line

test/dns-test.js
tynes: Going to remove all external requests. Thanks for the feedback

test/dns-test.js
tynes: Going to replace the whole auction process with just writing to the tree directly. This will speed up the tests, remove the need for a bunch of code and better decouple the tests. Good feedback

test/dns-test.js
tuxcanfly: Might be a silly question but in cases where there's multiple DNSKEY records (and multiple ZSK/KSK for some reason), this would select the last. Is that fine or maybe we could add a `break`?

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+        // when the native backend is being used
+        // because it uses OpenSSL which does not
+        // yet use RFC 6979, so nullify the
+        // signatures before comparing them
+
+        if (bcrypto.native === 2)
+          assert.deepEqual(nullSig(res), nullSig(dns));
+        else
+          assert.deepEqual(res.name, dns.name);
+      });
```
tuxcanfly: shouldn't this be `assert.deepEqual(res, dns);`?

test/data/resources-v0.json
```diff
@@ -0,0 +1,200 @@
...
+        "ttl": 64
+      }
```
tuxcanfly: All these test cases have only one host item. Would multiple be a good test case?

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+        verifyDNSSEC(dnskey, ksk, types.DNSKEY, '.');
+        // validate the signature over the rrsets
+        verifyDNSSEC(response, zsk, type, query);
+
+        // NOTE: the signatures are not canonical
+        // when the native backend is being used
+        // because it uses OpenSSL which does not
+        // yet use RFC 6979, so nullify the
+        // signatures before comparing them
+
```
tuxcanfly: nit: new line after comments

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+        assert(zsk instanceof wire.Record);
+        assert(ksk instanceof wire.Record);
+
+        // validate the signature on the ZSK
+        verifyDNSSEC(dnskey, ksk, types.DNSKEY, '.');
+        // validate the signature over the rrsets
+        verifyDNSSEC(response, zsk, type, query);
+
+        // NOTE: the signatures are not canonical
```
tuxcanfly: does this mean `verifyDNSSEC` manages to verify non-canonical signatures? If yes, is it possible to convert to canonical or otherwise verify them instead of skipping them altogether (like nullSig is doing now I presume).

lib/dns/records.js
tuxcanfly: indicating that values are DNS types is a good idea. `hnsToDnsTypes` or similar might be good.

test/dns-test.js
tynes: I can do that but would need to set the edns field along with a few bits in the header. You can see here when the tests run, there are additional fields that are set by the DNS server itself that are not set when calling `resource.toDNS`. Trimmed for brevity. This would expand the test coverage to include those fields being used by the DNS server.

```
  DNS Servers
    Authoritative Resolver
      ✓ should return no answer for HNS name pre-update
      1) should return authenticated INET4 record

  2 passing (948ms)
  20 failing

  1) DNS Servers
       Authoritative Resolver
         should return authenticated INET4 record:

      AssertionError [ERR_ASSERTION]: [Object] deepStrictEqual [Object]

      Expected values to be strictly deep-equal:
      
        {
          aa: true,
      ...
          authority: [],
      +   cd: true,
      -   cd: false,
          code: 'NOERROR',
          code: 'NOERROR',
      +   edns: {
      +     code: 0,
      +     dnssec: true,
      +     enabled: true,
      +     options: [],
      +     size: 4096,
      +     version: 0
      ...
      -   edns: undefined,
      -   id: 0,
      -   opcode: 'QUERY',
      -   qr: false,
      -   question: [],
      -   ra: false,
      -   rd: false,
      ...
      
      assert.deepEqual(res, dns);
             ^

      at Context.it (/home/tynes/Projects/github.com/handshake-org/hsd/test/dns-test.js:170:18)

```

test/data/resources-v0.json
tynes: Good thinking. You are correct in the sense that there are many different permutations of records inside of a resource. There are some code paths inside of `resource.toDNS` that these mocks do not cover, like reverse resolution. (Address -> Name possible?)

test/dns-test.js
tynes: Removed this external requests.

test/dns-test.js
tynes: These are all written directly to the tree now in a `before` block, so this is on longer relevant.

test/dns-test.js
tynes: `verifyDNSSEC` does work on non-canonical signatures, the problem was that a `deepEqual` did not work because the signatures were not canonical when using the native backend that wraps openssl.

tynes:
This PR depends on https://github.com/handshake-org/hsd/pull/286/files

test/dns-test.js
tynes: This code block was refactored out of the test suite

test/dns-test.js
```diff
@@ -0,0 +1,396 @@
...
+      assert.equal(question.name, name);
+
+      const [answer] = res.answer;
+      assert(typeof answer === 'object');
+      assert.equal(answer.name, name);
+    });
+  });
+
+  describe('Authoritative Resolver', () => {
+    it('should return no answer for an ICANN rooted name', async () => {
```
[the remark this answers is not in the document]
tynes: > I wonder if this test should be duplicated to simulate the scenario in which `com.` actually claims their name on hns.

I like this idea but think that it would expand the scope of this PR

test/data/resources-v0.json
tynes: Using both a mainnet and a testnet address that I generated locally

test/data/resources-v0.json
tynes: Added a new test case that covers multiple. The way that the test file is set up is that the key is the name of the handshake record and there is a single object following, so it is not possible to have multiple test cases per handshake record type. I could refactor the file to be a list for each handshake record type, and the test could iterate through each object in that list. Do you think that is a good idea? @tuxcanfly

tynes:
> Might be a silly question but in cases where there's multiple DNSKEY records (and multiple ZSK/KSK for some reason), this would select the last. Is that fine or maybe we could add a break?

I added an assertion for 1 zsk and 1 ksk being returned. I think that will be the expected behavior into the foreseeable future, as the dnssec keys aren't super important to the security of the network since the private key is public information.

It would be cool if we could allow a config option to set a dnssec key, then each resolver would be able to have their own. I think this is possible because the keys themselves are not kept on chain and are not part of consensus.

lib/dns/records.js
tynes: Ended up calling it `dnsByType` and have assertions in the test suite that the values are correct. Every time that we talk about what to call it, we come up with a different name :man_shrugging:

test/dns-test.js
```diff
@@ -0,0 +1,298 @@
...
+        // because it uses OpenSSL which does not
+        // yet use RFC 6979, so nullify the
+        // signatures before comparing them
+        if (bcrypto.native === 2) {
+          dns = nullSig(dns);
+          res = nullSig(res);
+          assert(dns && res);
+        }
+
+        assert.deepEqual(dns.answer, res.answer);
```
tynes: @tuxcanfly I am no longer checking equality of the dns message header, only the `answer`, `authority` and `additional`. The rest of the message may be different based on the configuration of the dns resolver/server.

test/dns-test.js
tynes: I'd like to check through `hsd` and `bns` to make sure that it already doesn't exist. Would the fn signature be:

`name (str), hsd resource (Resource), dns type (int) -> List[str] (list of query strings)`

A Resource can have multiple query strings depending on the Records inside of it, but this also may not return the name number of strings in the return value as the number of Records in the Resource which may be confusing to a library consumer.

test/dns-test.js
```diff
@@ -0,0 +1,298 @@
...
+        ns.set(raw, 0);
+        ns.setData(resource.encode());
+
+        // Create a database transaction, as
+        // writing directly to the database is
+        // discouraged. Write to the txn and
+        // then commit it.
+        const txn = node.chain.db.tree.transaction();
+        await txn.insert(nameHash, ns.encode());
+        await txn.commit();
```
pinheadmz: Awesome!

test/dns-test.js
tuxcanfly: OK if we not maybe `test/util/common` is a good place.

test/data/resources-v0.json
tuxcanfly: Yeah that would work :+1:

test/dns-test.js
tuxcanfly: Sounds good! :+1:

test/dns-test.js
tuxcanfly: OK.

tynes:
Need to rebase and squash these commits

test/dns-test.js
tynes: We know that technically the `DNSKEY` record shouldn't change, sending a request for it each time is extra paranoia. In practice the `DNSKEY` could be rolled over for a zone. It takes awhile for the root zone key to roll over, I wonder how often other zone keys roll over.

test/data/resources-v0.json
tynes: This has been resolved

test/dns-test.js
boymanjor: I think we should leave this as a specific test helper for now. We can abstract this into a proper library method when the functionality is needed elsewhere.

tynes: Ok this makes sense to me. Its not clearly documented anywhere in the Handshake docs that you need to use a particular domain name (with subdomains) to query for some of the rr types. Maybe the `bns` stub resolver could be extended for this repo where its wrapped with an easier to use API that will handle these cases as well as use handshake native types instead of dns rr types. I could imagine that being very helpful for onboarding new developers, especially if it works in the browser.

lib/dns/resource.js
```diff
@@ -1203,10 +1203,12 @@ class Resource extends Struct {
       case types.A:
         res.answer = this.toA(name);
         key.signZSK(res.answer, types.A);
+        key.signZSK(res.answer, types.TXT);
```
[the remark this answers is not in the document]
tynes: Opened a PR here: https://github.com/handshake-org/hsd/pull/292

Note that the test coverage for this is actually in this PR. I will rebase this PR on top of `master` once that is merged

tynes:
> I also think we can clean up the commit history.

I can rebase on master and rebase out a bunch of the intermediate commits after https://github.com/handshake-org/hsd/pull/292 gets merged.

Thanks for the review @boymanjor 

Note that I wanted to add SIG0 validation to this PR as well, I started in https://github.com/handshake-org/hsd/pull/289 but ran into a bug with `KEY` records. Right now I do not see a way to validate the SIG0 signature without being able to query for the key, which the [RFC](https://tools.ietf.org/html/rfc2931) defines `KEY` as the RR type used to return the SIG0 key. The authoritative resolver doesn't have logic for `KEY` RR types at the name `.`

See: https://github.com/handshake-org/hsd/blob/master/lib/dns/server.js#L228

lib/dns/resource.js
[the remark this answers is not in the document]
boymanjor: @tynes #292 has been merged.

tynes:
Needs rebase now that https://github.com/handshake-org/hsd/commit/91f1c2780999e2b9b22e3b3a540bed72a04cf803 has been merged

tynes:
@boymanjor Has been rebased to remove intermediate commits

PR #268, opened 2019-10-09
node/http: add mempool details param

tynes:
GET /mempool now accepts a query param
`details=bool`. When set to true, it
will return the TX JSON instead of just
the hashes. This is useful for block
explorers and also anybody who is interested
in more details about what exactly is in the
mempool at any given time.

I had fun watching the mempool using this
functionality.

bcoin: https://github.com/bcoin-org/bcoin/pull/869

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -334,11 +334,22 @@ class HTTP extends Server {
     this.get('/mempool', async (req, res) => {
       enforce(this.mempool, 'No mempool available.');
 
-      const hashes = this.mempool.getSnapshot();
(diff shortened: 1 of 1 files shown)


tynes:
See progression here: https://github.com/bcoin-org/bcoin/pull/869

nodech:
This is similar to: https://github.com/handshake-org/hsd/pull/641
>  it's pretty easy to collect necessary mempool transactions with additional requests, if used with decent HTTP client (can reuse connection - keep-alive) you can easily replicate the same thing.

PR #269, opened 2019-10-09
address: add toScript method and expose to api

tynes:
An address is a commitment to some program that will run in the Script interpreter. Every address has a mapping to a program. Depending on the type of address, it may or may not need an opening to the commitment. For example, p2pkh does not need an opening while p2sh does (the program preimage).

The abstract function that maps an address to a program accepts a version, data (hash in this codebase) and an opening to the commitment (top of Witness stack in this codebase). I find reasoning about addresses in this way to be useful. The `hsd` codebase calls things `Script` when in reality I think they should be called `Program`. There is a lot of confusion between what is a script/program/redeem script/witness etc.

This PR addresses this mapping by:

- Adding a `toScript` method to the `Address` class that encompasses the functionality of converting an address to a script that can be ran in the Script interpreter. See codepath mentioned above here:

https://github.com/handshake-org/hsd/blob/7c172277476ea2a41501413f4f43a4a0a040b100/lib/script/script.js#L2262

This method must accept the witness as an argument in the case of pay to witness script hash, as it is impossible to know the program that will be ran without the preimage. In the case of pay to pubkey hash, you do not need the witness.

For each of `Coin`, `Input` and `Output` the `getJSON` method additionally renders `program` and `asm`. The program is the `address.toScript` where the program is the raw hex bytecode of the program and the asm is the assembly printed for human consumption.

This will allow block explorers to better visualize the locking scripts of addresses. It will also allow block explorers to display the redeem script (program preimage) when p2sh outputs are spent. @kilpatty

### DIFF
--- a/lib/primitives/address.js
+++ b/lib/primitives/address.js
@@ -8,6 +8,8 @@
 
 const assert = require('bsert');
 const bio = require('bufio');
+const Script = require('../script/script');
+const Opcode = require('../script/opcode');
 const {bech32} = require('bstring');
 const blake2b = require('bcrypto/lib/blake2b');
 const sha3 = require('bcrypto/lib/sha3');
@@ -190,6 +192,38 @@ class Address extends bio.Struct {
     const hrp = network.addressPrefix;
 
     return bech32.encode(hrp, version, hash);
+  }
+
+  /**
+   * Create Script from Address.
+   * @param {Witness} witness
+   * @returns {Script|null}
+   */
+
+  toScript(witness) {
+    const hash = this.getHash();
+
+    if (this.version === 0) {
+      if (hash.length === 20)
+        return Script.fromPubkeyhash(hash);
+      else if (hash.length === 32) {
+        if (!witness)
+          return null;
+
+        const preimage = witness.get(-1);
+        if(sha3.digest(preimage).equals(hash))
+          return Script.decode(preimage);
+      }
+    }
+
+    if (this.version === 31) {
+      return Script.fromArray([
+        Opcode.fromSymbol('return'),
+        Opcode.fromPush(hash)
+      ]);
+    }
+
+    return null;
   }
 
   /**
--- a/lib/primitives/coin.js
+++ b/lib/primitives/coin.js
@@ -205,17 +205,22 @@ class Coin extends Output {
    * for JSON serialization.
    * @param {Network} network
    * @param {Boolean} minimal
+   * @param {Witness} witness
    * @returns {Object}
    */
 
-  getJSON(network, minimal) {
+  getJSON(network, minimal, witness) {
     network = Network.get(network);
+
+    const script = this.address.toScript(witness);
 
     return {
       version: this.version,
       height: this.height,
       value: this.value,
       address: this.address.toString(network),
+      program: script ? script.toHex() : undefined,
+      asm: script ? script.toASM() : undefined,
       covenant: this.covenant.toJSON(),
       coinbase: this.coinbase,
       hash: !minimal ? this.txid() : undefined,
--- a/lib/primitives/input.js
+++ b/lib/primitives/input.js
@@ -176,9 +176,11 @@ class Input extends bio.Struct {
   getJSON(network, coin, path) {
     network = Network.get(network);
 
-    let addr;
+    let addr, script;
     if (!coin) {
       addr = this.getAddress();
+      script = addr.toScript(this.witness);
+
       if (addr)
         addr = addr.toString(network);
     }
@@ -188,6 +190,8 @@ class Input extends bio.Struct {
       witness: this.witness.toJSON(),
       sequence: this.sequence,
       address: addr,
+      program: script ? script.toHex() : undefined,
+      asm: script ? script.toASM() : undefined,
       coin: coin ? coin.getJSON(network, true) : undefined,
       path: path ? path.getJSON(network) : undefined
     };
--- a/lib/primitives/output.js
+++ b/lib/primitives/output.js
@@ -169,15 +169,20 @@ class Output extends bio.Struct {
    * Convert the output to an object suitable
    * for JSON serialization.
    * @param {Network} network
+   * @param {Witness} witness
    * @returns {Object}
    */
 
-  getJSON(network) {
+  getJSON(network, witness) {
     network = Network.get(network);
+
+    const script = this.address.toScript(witness);
 
     return {
       value: this.value,
       address: this.address.toString(network),
+      program: script ? script.toHex() : undefined,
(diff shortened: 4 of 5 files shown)


tynes:
See example `TX.toJSON` with these changes applied:

```
{
  "hash": "7a03af60884104ef9f27ce0ba11b8f59754bb2585743cbcf76ecd9cc72aafe86",
  "witnessHash": "91a1360e61826b95531c8ec89d3d62768526125a4737390a16a79a6e272aee51",
  "mtime": 1570653688,
  "version": 0,
  "inputs": [
    {
      "prevout": {
        "hash": "2dbdf026de53fdde97a932e5cf23f401fb671157666ce1bc1a3c1640071c48b7",
        "index": 1
      },
      "witness": [
        "654c3fa70b5e3e54b925b97f9cf516066fe2719a83074842b778d1d62501f69a2b59ed279821d8c0e43fc1222e1e7df32dc2c234cff114bf04b0e8688a16ec9401",
        "035bd7808c83826cad9ab1997c88be3564117e27a083ea442e359779c3b4d84675"
      ],
      "sequence": 4294967295,
      "address": "hs1qhrc264e234wxqhl5tugvxp7lvllwc2phsmfdp6",
      "program": "76c014b8f0ad572a8d5c605ff45f10c307df67feec283788ac",
      "asm": "OP_DUP OP_BLAKE160 b8f0ad572a8d5c605ff45f10c307df67feec2837 OP_EQUALVERIFY OP_CHECKSIG"
    }
  ],
  "outputs": [
    {
      "value": 0,
      "address": "hs1qrsmnssdez3mp6kadvvwegfzh8h6fe06y6fhwjk",
      "program": "76c0141c373841b914761d5bad631d9424573df49cbf4488ac",
      "asm": "OP_DUP OP_BLAKE160 1c373841b914761d5bad631d9424573df49cbf44 OP_EQUALVERIFY OP_CHECKSIG",
      "covenant": {
        "type": 3,
        "action": "BID",
        "items": [
          "bbd9e905510ec6ca554591336a8ef507390c1fdb3200ea00a3af637535acfb63",
          "9d5e0000",
          "74687265652d68756e647265642d646f6d61696e732d74656e2d626964732d332d3634",
          "342714d9ea2c4eeb81209ebcedbc7d760cbff00a4edea5e67f297848b1f8051d"
        ],
        "asm": "TYPE:BID NAMEHASH:bbd9e905510ec6ca554591336a8ef507390c1fdb3200ea00a3af637535acfb63 HEIGHT:24221 NAME:74687265652d68756e647265642d646f6d61696e732d74656e2d626964732d332d3634 HASH:342714d9ea2c4eeb81209ebcedbc7d760cbff00a4edea5e67f297848b1f8051d"
      }
    },
    {
      "value": 59998932140,
      "address": "hs1q2a6yys0p22eennza42awfhtzd0wg6prczyaehc",
      "program": "76c01457744241e152b399cc5daabae4dd626bdc8d047888ac",
      "asm": "OP_DUP OP_BLAKE160 57744241e152b399cc5daabae4dd626bdc8d0478 OP_EQUALVERIFY OP_CHECKSIG",
      "covenant": {
        "type": 0,
        "action": "NONE",
        "items": [],
        "asm": "TYPE:NONE"
      }
    }
  ],
  "locktime": 0,
  "hex": "00000000012dbdf026de53fdde97a932e5cf23f401fb671157666ce1bc1a3c1640071c48b701000000ffffffff02000000000000000000141c373841b914761d5bad631d9424573df49cbf44030420bbd9e905510ec6ca554591336a8ef507390c1fdb3200ea00a3af637535acfb63049d5e00002374687265652d68756e647265642d646f6d61696e732d74656e2d626964732d332d363420342714d9ea2c4eeb81209ebcedbc7d760cbff00a4edea5e67f297848b1f8051dac0c37f80d000000001457744241e152b399cc5daabae4dd626bdc8d04780000000000000241654c3fa70b5e3e54b925b97f9cf516066fe2719a83074842b778d1d62501f69a2b59ed279821d8c0e43fc1222e1e7df32dc2c234cff114bf04b0e8688a16ec940121035bd7808c83826cad9ab1997c88be3564117e27a083ea442e359779c3b4d84675"
}
```

Need to test the p2wsh case before ready to merge. Looking for acks

tynes:
Updated with tests

pinheadmz:
I think this may be more problematic than its worth. 

> This method must accept the witness as an argument, as it is impossible to know the program that will be ran without the preimage.

I think this is a pretty good sign that this method doesn't belong in `Address`, I think `Script.decode()` is good where it is, and doesn't need to be linked here. Addresses, as a concept, are for the sender, who only needs minimal information about the spending conditions (pubkey vs script -- even easier than bitcoin 😄 ). If you know the witness, you are already one step past the address.

tynes:
There is no good way to distinguish what actual bitcoin script will be executed without knowing the protocol. People are not going to know to look at line 2200+ in `lib/script/script.js` to understand how addresses correspond to the actual bitcoin script that executes.

I also updated what you quoted above, as that applies in the p2wsh case. In the pay to pubkey hash case, you can always get back the correct script.

pinheadmz:
Actually, probably a more important point is that when spending from a witness pubkey hash output - _there is no script at all_. `Script.fromPubKeyHash()` is just used internally. SegWit removed the need for anyone to ever write OPDUP HASH160 etc on the chain...

tynes:
The `program` is the script that runs to determine if it is a valid state transition. The `witness` is the initial stack for the VM execution. This happens for every transaction, and just seeing the address does not give insight into what is happening under the hood. You have the `witness` for the `input` so you are guaranteed to have a correctly rendered program.

For outputs, It is not guaranteed that you will have the program in this case, depending on the type of UTXO it is. If it is pay to pubkey hash, then you do have the program, it is simply filling a template. If it is a pay to script hash, you will likely not have the script. For a project like Bitcoin, you generally want to keep you script preimages hidden because you want all BTC to be fungible. I do not think that Handshake is money, so I think that in some cases it is ok to reveal your script preimages. Based on a counterparty learning the script and then hashing it and checking it against the hash encoded in the address, (which should be accessible as a hash and not just as an address because that is an additional encoding/decoding step which is annoying) they can determine their level of trust when transacting with you. This is how the non interactive nameswap scheme works - the counterparty must know that the name holder cannot back out of the deal by spending to a `UPDATE` and cancelling the atomic swap.

lib/primitives/address.js
```diff
@@ -192,6 +194,38 @@ class Address extends bio.Struct {
     return bech32.encode(hrp, version, hash);
   }
 
+  /**
+   * Create Script from Address.
+   * @param {Witness} witness
```
tynes: This argument is optional `{Witness?}`

PR #272, opened 2019-10-09
policy: move blacklist to protocol/policy

tynes:
Moving the already defined policy based blacklist to the `protocol/policy` file. It is defined as a `Set`, so a smarter blacklist class can be defined if it implements the `Set` interface.


boymanjor:
We spoke out of band, but I believe the dns server implementation is the right place for the blacklist. We've decided to close this PR. @tynes mentioned creating `lib/dns/policy.js` to store policy related constants. Perhaps we can open discussion on this in an issue or within another PR.

PR #275, opened 2019-10-11
net: getproof - add dynamic decaying dos score

tuxcanfly:
TODO:

- [x] finetune decay params
- [x] add tests

### DIFF
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -105,11 +105,11 @@ exports.DOS_RESET_TIME = 30 * 60;
  * @default
  */
 
-const dos_decay_factor = {};
+const dosDecayFactor = {};
 for (let i=0; i<64;i++)
-    dos_decay_factor[i] = Math.exp(-1.0 * i * Math.log(2) / 60);
+    dosDecayFactor[i] = Math.exp(-1.0 * i * Math.log(2) / 60);
 
-exports.DOS_DECAY_FACTOR = dos_decay_factor;
+exports.DOS_DECAY_FACTOR = dosDecayFactor;
 
 /**
  * Ban score threshold before ban is placed in effect.
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -92,6 +92,26 @@ exports.MAX_MESSAGE = 8 * 1000 * 1000;
 exports.BAN_TIME = 24 * 60 * 60;
 
 /**
+ * Amount of time after which to reset DOS peers.
+ * @const {Number}
+ * @default
+ */
+
+exports.DOS_RESET_TIME = 30 * 60;
+
+/*
+ * Precomputed decay factor for first 64 seconds.
+ * @const {Object}
+ * @default
+ */
+
+const dos_decay_factor = {};
+for (let i=0; i<64;i++)
+    dos_decay_factor[i] = Math.exp(-1.0 * i * Math.log(2) / 60);
+
+exports.DOS_DECAY_FACTOR = dos_decay_factor;
+
+/**
  * Ban score threshold before ban is placed in effect.
  * @const {Number}
  * @default
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -106,7 +106,7 @@ exports.DOS_RESET_TIME = 30 * 60;
  */
 
 const dosDecayFactor = {};
-for (let i=0; i<64;i++)
+for (let i=0; i < 64; i++)
     dosDecayFactor[i] = Math.exp(-1.0 * i * Math.log(2) / 60);
 
 exports.DOS_DECAY_FACTOR = dosDecayFactor;
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -120,6 +120,14 @@ exports.DOS_DECAY_FACTOR = dosDecayFactor;
 exports.BAN_SCORE = 100;
 
 /**
+ * DOS score threshold before ban is placed in effect.
+ * @const {Number}
+ * @default
+ */
+
+exports.DOS_SCORE = 100;
+
+/**
  * Create a nonce.
  * @returns {Buffer}
  */
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -125,7 +125,16 @@ exports.BAN_SCORE = 100;
  * @default
  */
 
-exports.DOS_SCORE = 1000;
+exports.DOS_SCORE = 144;
+
+/**
+ * Requests per second for P2P messages. Breaching this will constitute a DOS
+ * ban. Currently only used for getproof.
+ * @const {Number}
+ * @default
+ */
+
+exports.P2P_RPS = 100;
 
 /**
  * Create a nonce.
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -97,17 +97,17 @@ exports.BAN_TIME = 24 * 60 * 60;
  * @default
  */
 
-exports.DOS_RESET_TIME = 30 * 60;
+exports.DOS_RESET_TIME = 30;
 
 /*
- * Precomputed decay factor for first 64 seconds.
+ * Precomputed decay factor for first 1024 seconds.
  * @const {Object}
  * @default
  */
 
 const dosDecayFactor = {};
-for (let i=0; i < 64; i++)
-    dosDecayFactor[i] = Math.exp(-1.0 * i * Math.log(2) / 60);
+for (let i=0; i < 1024; i++)
+    dosDecayFactor[i] = Math.exp(-1.0 * i * Math.log(2));
 
 exports.DOS_DECAY_FACTOR = dosDecayFactor;
 
@@ -125,7 +125,7 @@ exports.BAN_SCORE = 100;
  * @default
  */
 
-exports.DOS_SCORE = 100;
+exports.DOS_SCORE = 1000;
 
 /**
  * Create a nonce.
--- a/lib/net/common.js
+++ b/lib/net/common.js
@@ -101,6 +101,9 @@ exports.DOS_RESET_TIME = 30;
 
 /*
  * Precomputed decay factor for first 1024 seconds.
+ * The decay is logarithmic with respect to time.
+ * See: http://mathworld.wolfram.com/ExponentialDecay.html
+ * N(t) = N0 * (e ^ (-lambda * t))
  * @const {Object}
  * @default
  */
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -1916,7 +1916,7 @@ class Peer extends EventEmitter {
   increaseDos(dosScore, time) {
     // dosScore is a time-bound transient banscore useful to detect DOS
     // attacks by repeated requests for a message (e.g. getproof).  Calling
-    // increaseBan with a dosScore multiple times will increase dosScore in a
+    // increaseDos with a dosScore multiple times will increase dosScore in a
     // logarithmically decaying fashion wrt time between requests to penalize
     // short bursts of repeated requests.  The dosScore decays back to 0 in
     // 30s.
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -1921,18 +1921,17 @@ class Peer extends EventEmitter {
     if (seconds > common.DOS_RESET_TIME) {
       this.dosScore = 0;
     } else if (this.dosScore > 1 && seconds > 0) {
-      if (seconds < 64) {
-        this.dosScore *= common.DOS_DECAY_FACTOR[Math.floor(seconds)];
-      } else {
+      if (seconds < 64)
+        this.dosScore *= common.DOS_DECAY_FACTOR[seconds | 0];
+      else
         this.dosScore *= Math.exp(-1.0 * seconds * (Math.log(2) / 60));
-      }
     }
     this.lastDos = time;
 
-    this.dosScore = Math.floor(this.dosScore) + dosScore;
+    this.dosScore = (this.dosScore | 0) + dosScore;
 
     if (this.dosScore >= this.options.dosScore) {
-      this.logger.debug('Ban threshold exceeded (%s).', this.hostname());
+      this.logger.debug('DOS threshold exceeded (%s).', this.hostname());
       this.ban();
       return true;
     }
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -1907,6 +1907,7 @@ class Peer extends EventEmitter {
 
   /**
    * Increase dosscore on peer.
+   * See: https://github.com/btcsuite/btcd/blob/master/connmgr/dynamicbanscore.go
    * @param {Number} dosScore
    * @param {Number} time
    * @returns {Boolean}
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -1888,34 +1888,50 @@ class Peer extends EventEmitter {
   /**
    * Increase banscore on peer.
    * @param {Number} banScore
+   * @returns {Boolean}
+   */
+
+  increaseBan(banScore) {
+    this.banScore += banScore;
+
+    if (this.banScore >= this.options.banScore) {
+      this.logger.debug('Ban threshold exceeded (%s).', this.hostname());
+      this.ban();
+      return true;
+    }
+
+    return false;
+  }
+
+  /**
+   * Increase dosscore on peer.
    * @param {Number} dosScore
    * @param {Number} time
    * @returns {Boolean}
    */
 
-  increaseBan(banScore, dosScore, time) {
-    this.banScore += banScore;
-
-    if (dosScore > 0) {
-      // dosScore is a time-bound transient banscore useful to detect DOS
-      // attacks by repeated requests for a message (e.g. getproof).  Calling
-      // increaseBan with a dosScore multiple times will increase the overall
-      // banscore in a logarithmically decaying fashion to penalize short
-      // bursts of repeated requests.  The dosScore decays back to 0 in 30mins,
-      // i.e. 1800s.
-      const seconds = (time - this.lastDos) / 1000;
-      if (1800 < seconds) {
-        this.dosScore = 0;
-      } else if (this.dosScore > 1 && seconds > 0) {
+  increaseDos(dosScore, time) {
+    // dosScore is a time-bound transient banscore useful to detect DOS
+    // attacks by repeated requests for a message (e.g. getproof).  Calling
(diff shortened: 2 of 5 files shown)


lib/net/peer.js
```diff
@@ -81,6 +81,9 @@ class Peer extends EventEmitter {
     this.invQueue = [];
     this.onPacket = null;
 
+    this.dosScore = 0;
```
tynes: What is your reasoning around adding a new property `dosScore` when `banScore` already exists to handle banning peers?

tynes:
I think that network usage over time will change and that testnet usage will be different than mainnet usage, so getting real world data to fine tune on testnet will not fully apply to mainnet. Maybe the banning function is lenient at first and then made more strict when it becomes a problem? That is not ideal because then people would have to update their software, but early in the blockchain's lifecycle it isn't the biggest problem. Another way would be to dos `hsd` on multiple "hardware types" and see how well it holds up and use that information to tune it so that it prevents dos on the more lightweight hardware. Large providers may want to ease the parameters if they are serving a lot of legitimate requests in that case. Maybe some of the variables can be policy based?

lib/net/peer.js
tuxcanfly: The idea is to have a constant score like `bansScore` which increases linearly, and this can continue to be used as usual. However, dos attempts generally need to be distinguised form regular client messages with further inspection usually based on time (eg a regular peer might request 10 `getproof` per minute, whereas a malicious peer might attempt 1000 `getproof` per minute). This new score needs to increase when requests/minute increase but decay and reset to 0 as requests decrease. They are conceptually different but can be added together to get a score indicating both different types of malicious requests as well mutliple attempts at abusing some of these requests.

tuxcanfly:
Yeah the initial parameters need to be estimated based on usage. I think policy is a bit too steep for this, config should likely suffice.

tynes:
Concept ack on parameters as config

tynes:
I notice that the implementation is similar to: https://github.com/btcsuite/btcd/blob/86fed781132ac890ee03e906e4ecd5d6fa180c64/connmgr/dynamicbanscore.go

Do you think its a good idea to precompute a score table, in a similar fashion to how they do it?

tuxcanfly:
Yeah that can be done, could speed up incrementing the score.

tynes:
Hey @tuxcanfly, any updates on this?

tuxcanfly:
@tynes I've left `dosScore` as an option in `PeerOptions` which defaults to `BAN_SCORE` for now. With the current scores, 10 consecutive `getproof` requests are allowed and anything beyond that gets banned. After that the score decays slowly each second, eventually being reset to 0 after 30mins.

I think this is good enough for a start but if there are any issues we can tweak these params as required.

tynes:
I felt like I didn't see `Math.floor` used in the codebase very often, and saw `(Number >>> 0)` or (Number | 0) being used more frequently, so I put together this benchmark.

```javascript
/**
 * Benchmarking ways to round down
 * to the nearest integer in JavaScript.
 */

'use strict';

const assert = require('assert');

function bench(name) {
  const start = process.hrtime();
  return function end(ops) {
    const elapsed = process.hrtime(start);
    const time = elapsed[0] + elapsed[1] / 1e9;
    const rate = ops / time;

    console.log('%s: ops=%d, time=%d, rate=%s',
      name, ops, time, rate.toFixed(5));
  };
};

const N = 100;
// create a bunch of random numbers
const numbers = [];
for (let i = 0; i < N; i++)
  numbers.push(Math.random() * 10);

{
  const end = bench('Math.floor');

  for (let number of numbers)
    number = Math.floor(number);

  end(N);
}

{
  const end = bench('>>>');

  for (let number of numbers)
    number = (number >>> 0);

  end(N);
}

{
  const end = bench('|');

  for (let number of numbers)
    number = (number | 0);

  end(N);
}

// make sure that it is rounding down as expected
for (let number of numbers) {
  assert.deepEqual(Math.floor(number), (number >>> 0));
  assert.deepEqual(Math.floor(number), (number | 0));
}
```

Running it locally on my machine produces:

```
Math.floor: ops=100, time=0.00006423, rate=1556904.87311
>>>: ops=100, time=0.00000812, rate=12315270.93596
|: ops=100, time=0.000006987, rate=14312294.26077
```

You can see that `|` is the most performant when `N` is 100. The performance benefits begin to go away and become more comparable to `Math.floor` when `N` becomes larger.

I ran assertions at the end to make sure that it is behaving as expected.
I think we should remove `Math.floor` in favor of `(Number | 0)`

lib/net/common.js
```diff
@@ -91,6 +91,26 @@ exports.MAX_MESSAGE = 8 * 1000 * 1000;
...
+
+exports.DOS_RESET_TIME = 30 * 60;
+
+/*
+ * Precomputed decay factor for first 64 seconds.
+ * @const {Object}
+ * @default
+ */
+
+const dosDecayFactor = {};
```
tynes: I benchmarked an array vs an object as well, the object was faster :+1:

lib/net/common.js
```diff
@@ -91,6 +91,26 @@ exports.MAX_MESSAGE = 8 * 1000 * 1000;
...
+for (let i=0; i<64;i++)
```
tynes: Nit: missing space

lib/net/peer.js
```diff
@@ -1900,6 +1903,43 @@ class Peer extends EventEmitter {
...
+    // attacks by repeated requests for a message (e.g. getproof).  Calling
+    // increaseBan with a dosScore multiple times will increase dosScore in a
+    // logarithmically decaying fashion wrt time between requests to penalize
+    // short bursts of repeated requests.  The dosScore decays back to 0 in
+    // 30mins, i.e. 1800s.
+    const seconds = (time - this.lastDos) / 1000;
+    if (seconds > common.DOS_RESET_TIME) {
+      this.dosScore = 0;
+    } else if (this.dosScore > 1 && seconds > 0) {
+      if (seconds < 64) {
```
tynes: Nit: no need for curly braces with 1 liners

lib/net/peer.js
```diff
@@ -1900,6 +1903,43 @@ class Peer extends EventEmitter {
...
+      } else {
+        this.dosScore *= Math.exp(-1.0 * seconds * (Math.log(2) / 60));
+      }
+    }
+    this.lastDos = time;
+
+    this.dosScore = Math.floor(this.dosScore) + dosScore;
+
+    if (this.dosScore >= this.options.dosScore) {
+      this.logger.debug('Ban threshold exceeded (%s).', this.hostname());
```
tynes: Could you update this message so we know it was caused by the `dosScore`?

tuxcanfly:
https://groups.google.com/d/msg/public-dns-discuss/p1o62SJElck/w0flYsmqBQAJ

Google's Public DNS limits 100 queries per second per IP. Borrowing from this I can make each query increase the DOS score by 1 point and set the threshold to 100.

lib/net/peer.js
tuxcanfly: Fixed.

lib/net/peer.js
tuxcanfly: Fixed.

lib/net/common.js
tuxcanfly: Fixed.

lib/net/common.js
tuxcanfly: Cool :+1:

lib/net/peer.js
```diff
@@ -1900,6 +1903,42 @@ class Peer extends EventEmitter {
...
+   * Increase dosscore on peer.
+   * @param {Number} dosScore
+   * @param {Number} time
+   * @returns {Boolean}
+   */
+
+  increaseDos(dosScore, time) {
+    // dosScore is a time-bound transient banscore useful to detect DOS
+    // attacks by repeated requests for a message (e.g. getproof).  Calling
+    // increaseBan with a dosScore multiple times will increase dosScore in a
```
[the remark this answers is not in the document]
boymanjor: Also, should we cite the btcd implementation as our reference?

lib/net/common.js
```diff
@@ -91,6 +91,26 @@ exports.MAX_MESSAGE = 8 * 1000 * 1000;
...
+
+/*
+ * Precomputed decay factor for first 1024 seconds.
+ * @const {Object}
+ * @default
+ */
+
+const dosDecayFactor = {};
+for (let i=0; i < 1024; i++)
+    dosDecayFactor[i] = Math.exp(-1.0 * i * Math.log(2));
```
boymanjor: nit: what do you think about a comment referencing the origin of this exponential function?

test/peer-test.js
```diff
@@ -0,0 +1,74 @@
...
+    assert.strictEqual(peer.dosScore, 25);
+  });
+
+  it('should reset dos score after decay', async () => {
+    peer.increaseDos(10, Date.now());
+    assert.strictEqual(peer.dosScore, 10);
+
+    // override reset time to 0s
+    const dosResetTime = common.DOS_RESET_TIME;
+    common.DOS_RESET_TIME = 0;
```
boymanjor: I want to suggest allowing a configurable dos reset time on the pool and peer objects, but I noticed the ratio of the transient dos score's lifespan to its half-life is 30/1. This ratio is held in the btcd implementation, as well. How sensitive is this ratio to manipulation? Would we need to keep this ratio constant and update the exponential function to account for a dynamic half-life?  Related, how would updating these parameters affect the precomputed decay factors? Specifically, is the size of the precomputed table a magic number, or can it be calculated dynamically?

boymanjor: I'm trying to grok whether it is better to keep this value constant and only give the user one parameter to play around with (the dos score limit).

tuxcanfly: So, we have the following equations:

    dN/dt = −λN

    N = N0e−λt

    T1/2 = 0.693/λ

Here N = max dos score, λ = decay constant and T1/2 = half-life

In our case, we want to provide rate of decay (dN/dt) as a config and tweak constant `λ` accordingly. Latest commit adds a `PeerOptions` option `p2prps`, which is a generic option for tweaking max rps peers are allowed, using which we compute `λ` and use it calculate the half-life.

I've tweaked default params such that we maintain a steady rate of 100 requests per second, but obviously callers can tweak this.

(Note: temporarily removed precomputed constants since it depends on half-life now).

tuxcanfly: So, our scenario is a bit more complicated. In the above equations, the score `N` never increases. However for us, `N` can increase because of new requests linearly as well as decrease logarithmically because of the decay.

In practice it appears that the rps is about 10% of the configured rps. It might be possible to compute the 'practical' rps by using some computations. Not sure if it's worth the time investment. If the rps always scales 0.1x we could just use a rule of thumb and 10x the config parameter. :man_shrugging:

tuxcanfly: OK, so the calculation is not too difficult:

At time `t` seconds, assuming an `rps` requests per second, we have incoming:

`t * rps` requests each adding 1 dos score. At the same time, assuming half-life 1s, the cumulative decay of previous scores (each rps decaying in 1s) is:

    rps / 2^1 + rps / 2^2 + rps / 2^3 ... rps / 2^t = rps / 2 ^ t

so total:

    t * rps - rps * 2 ^ t  = rps * (t - 1/2^t)

So we conclude that at any give time `t` we will have `rps * (t - 1/2^t)` dos score which is the actual or practical rps. This is the scaling factor we got above (0.1x), so we can generalize it.

tuxcanfly: Correction: since the incoming scores are already accounted for in the decaying scores, the total score at time `t` is simply:

    rps * (1 - 1/2^t)

tuxcanfly: So, all this means is if you are aiming to serve `rps` requests per second, when `t` is small you need to overshoot by a multiple of about `1-2x` to actually achieve the required rps. This is because recent rps will clog up more dos score when starting up. As `t` approaches infinity, you will get the required rate as all the old dos scores have decayed. I hope the explanation makes sense.

lib/net/peer.js
```diff
@@ -2145,6 +2185,16 @@ class PeerOptions {
       this.banScore = options.banScore;
     }
 
+    if (options.dosScore != null) {
+      assert(typeof options.dosScore === 'number');
+      this.dosScore = options.dosScore;
+    }
+
+    if (options.p2prps != null) {
```
tynes: This should be parsed from the config and passed through in the file https://github.com/handshake-org/hsd/blob/master/lib/node/fullnode.js

lib/net/peer.js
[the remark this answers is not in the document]
boymanjor: @tuxcanfly did you see these comments?

lib/net/common.js
boymanjor: @tuxcanfly did you see this comment?

tuxcanfly: Sorry, added a note and the formula for explanation.

lib/net/peer.js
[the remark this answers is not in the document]
tuxcanfly: OK added a link.

lib/net/peer.js
tuxcanfly: OK, added a config option `p2prps`. Could be worded and explained better imo since it only affects getproof messages atm.

lib/net/common.js
boymanjor: @tuxcanfly given the updated calculation with a dynamically calculated half-life, should we just remove this pre-computation table? It is not being used anymore.

tuxcanfly: Yeah it's not being used atm. But it would nevertheless be useful if it could use a given half-life, maybe after config is parsed for example, as a minor optimization. Will benchmark and move or remove it depending on the results.

lib/net/peer.js
```diff
@@ -1900,6 +1905,40 @@ class Peer extends EventEmitter {
...
+   * See: https://github.com/btcsuite/btcd/blob/master/connmgr/dynamicbanscore.go
+   * @param {Number} dosScore
+   * @param {Number} time
+   * @returns {Boolean}
+   */
+
+  increaseDos(dosScore, time) {
+    // dosScore is a time-bound transient banscore useful to detect DOS
+    // attacks by repeated requests for a message (e.g. getproof).  Calling
+    // increaseBan with a dosScore multiple times will increase dosScore in a
```
boymanjor: nit: @tuxcanfly this comment mentions "increaseBan" instead of "increaseDos."

tuxcanfly: Thanks, fixed.

test/peer-test.js
boymanjor: > So, all this means is if you are aiming to serve rps requests per second, when t is small you need to overshoot by a multiple of about 1-2x to actually achieve the required rps.

@tuxcanfly are you saying that, in practice, if I want to serve `x` requests per second, but my overall decay reset period is quite small, I will need to configured `p2prps` as `2x` because small `t` result in a bloated dos score due to the nature of exponential drop offs?

tuxcanfly: Yep, that's right

lib/net/peer.js
```diff
@@ -81,6 +81,11 @@ class Peer extends EventEmitter {
     this.invQueue = [];
     this.onPacket = null;
 
+    this.dosScore = 0;
+    this.lastDos = 0;
+
+    this.halflife = (0.693 / (this.options.p2prps / this.options.dosScore));
```
tuxcanfly: Reminder to use const.

lib/net/peer.js
```diff
@@ -1900,6 +1905,40 @@ class Peer extends EventEmitter {
...
+    // attacks by repeated requests for a message (e.g. getproof).  Calling
+    // increaseDos with a dosScore multiple times will increase dosScore in a
+    // logarithmically decaying fashion wrt time between requests to penalize
+    // short bursts of repeated requests.  The dosScore decays back to 0 in
+    // 30s.
+    const seconds = ((time - this.lastDos) / 1000) | 0;
+    if (seconds > common.DOS_RESET_TIME) {
+      this.dosScore = 0;
+    } else {
+        this.dosScore *= Math.exp(-1.0 * this.halflife * seconds * Math.log(2));
```
tynes: Nit: 4 spaces instead of 2, also unnecessary curly braces (`{`,`}`)

lib/net/pool.js
```diff
@@ -3157,6 +3157,7 @@ class Pool extends EventEmitter {
   async handleGetProof(peer, packet) {
     const {root, key} = packet;
     const proof = await this.chain.db.prove(root, key);
+    peer.increaseDos(1, Date.now());
```
tynes: Thoughts on using `Date.now` vs `util.ms`?

`util.ms` simply wraps `Date.now`.

I'm not particularly opinionated on either, it looks like `Date.now` is used more often so maybe its better to stick with that.

lib/node/fullnode.js
```diff
@@ -93,6 +93,7 @@ class FullNode extends Node {
       identityKey: this.identityKey,
       maxOutbound: this.config.uint('max-outbound'),
       maxInbound: this.config.uint('max-inbound'),
+      p2prps: this.config.uint('p2prps'),
```
tynes: What do you think of making the config option be `p2p-rps`? I think that is a lot easier to read.

test/peer-test.js
```diff
@@ -0,0 +1,77 @@
...
+
+    await sleep(1000);
+
+    peer.increaseDos(10, Date.now());
+    assert.strictEqual(peer.dosScore, 10);
+
+    common.DOS_RESET_TIME = dosResetTime;
+  });
```
tynes: Maybe add a unit test that asserts the ban happened after the multiple `increaseDos` calls? Can hook into the `ban` event on the `Peer`

https://github.com/handshake-org/hsd/blob/8beae620353ec7461944d8074ca0cb841a7b4372/lib/net/peer.js#L1946-L1948

boymanjor:
@tuxcanfly this will be ready to merge once the remaining comments are addressed.

tynes:
Would like to deprecate this PR in favor of https://github.com/handshake-org/hsd/pull/334

boymanjor:
Closing in favor of #334.

boymanjor:
Reopening this PR. Either this or #334 should be merged before mainnet launch.

tuxcanfly:
If we want to reconsider this I'd like to redo this PR in the same manner as the other sliding window PR. It would make it easier to review and test, benchmarka as well.

boymanjor:
@tuxcanfly we're going reclose this in favor of #334.

Issue #197, opened 2019-06-17
Port bcoin Net Tests


tynes:
Net tests were added to bcoin in this commit: https://github.com/bcoin-org/bcoin/commit/a35516d4e0096314948a41ecba1e81c7467ceb61

These tests cover net packets and would be useful to have in `hsd`. The packet classes are just different enough to make it so that its not a simple copy/paste from `bcoin`.

PR #277, opened 2019-10-15
test: add lib/net tests

tynes:
Net tests with a couple of changes to packets for safety

Port of https://github.com/bcoin-org/bcoin/commit/a35516d4e0096314948a41ecba1e81c7467ceb61

Closes https://github.com/handshake-org/hsd/issues/197

### DIFF
--- a/lib/net/packets.js
+++ b/lib/net/packets.js
@@ -303,7 +303,7 @@ class PingPacket extends Packet {
   /**
    * Create a `ping` packet.
    * @constructor
-   * @param {BN?} nonce
+   * @param {Buffer} nonce
    */
 
   constructor(nonce) {
@@ -311,7 +311,7 @@ class PingPacket extends Packet {
 
     this.type = exports.types.PING;
 
-    this.nonce = nonce || null;
+    this.nonce = nonce || common.ZERO_NONCE;
   }
 
   /**
@@ -355,7 +355,7 @@ class PongPacket extends Packet {
   /**
    * Create a `pong` packet.
    * @constructor
-   * @param {BN?} nonce
+   * @param {Buffer} nonce
    */
 
   constructor(nonce) {
@@ -1839,7 +1839,8 @@ class UnknownPacket extends Packet {
   /**
    * Inject properties from serialized data.
    * @private
-   * @param {Buffer} data
+   * @param {BufferReader} data
+   * @param {Number} type
    */
 
   read(data, type) {
--- a/lib/net/packets.js
+++ b/lib/net/packets.js
@@ -1843,8 +1843,8 @@ class UnknownPacket extends Packet {
    * @param {Number} type
    */
 
-  read(data, type) {
-    this.data = data;
+  read(br, type) {
+    this.data = br.readBytes(br.getSize());
     this.type = type;
     return this;
   }
--- a/test/net-test.js
+++ b/test/net-test.js
@@ -1,0 +1,788 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const {resolve} = require('path');
+const fs = require('fs');
+const {BloomFilter} = require('bfilter');
+const {nonce} = require('../lib/net/common');
+const consensus = require('../lib/protocol/consensus');
+const Framer = require('../lib/net/framer');
+const packets = require('../lib/net/packets');
+const NetAddress = require('../lib/net/netaddress');
+const {CompactBlock, TXRequest, TXResponse} = require('../lib/net/bip152');
+const InvItem = require('../lib/primitives/invitem');
+const Headers = require('../lib/primitives/headers');
+const Block = require('../lib/primitives/block');
+const MemBlock = require('../lib/primitives/memblock');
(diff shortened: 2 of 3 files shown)


lib/net/packets.js
```diff
@@ -320,7 +320,7 @@ class PingPacket extends Packet {
    */
 
   getSize() {
-    return 8;
+    return this.nonce ? 8 : 0;
```
kilpatty: I think this is an intention change from JJ to not allow 0 byte ping packets. With how it's currently setup, attempting to create a ping packet without a nonce will result in an assertion error. Likewise, attempting to read a ping packet without 8 bytes will be an assertion error.

kilpatty: Same goes for the two instances below this `write` and `read`

test/net-test.js
```diff
@@ -0,0 +1,802 @@
...
+      check(pkt);
+
+      pkt = packets.VerackPacket.decode(pkt.encode());
+      check(pkt);
+    });
+
+    it('ping', () => {
+      const check = (pkt, nonce) => {
+        assert.equal(pkt.type, packets.types.PING);
+        if (nonce)
```
tynes: Eliminate this nonce check and always assert there is a nonce

lib/net/packets.js
tynes: Letting `nonce` fallback to `null` is unsafe. Maybe we fail early and assert that `nonce` is a `Buffer` in the constructor? I know there are no assertions in other constructors, so the other options include generating the nonce on the fly (which I don't think will work well) or initializing the fallback value to 8 bytes of zero value

tynes: Addressed by following the same pattern in the `PongPacket` where the fallback value is `common.ZERO_NONCE`

tynes:
@boymanjor This is ready for you to look at

test/net-test.js
```diff
@@ -0,0 +1,785 @@
...
+      const framer = new Framer('regtest');
+      assert.strictEqual(framer.network, Network.get('regtest'));
+    });
+
+    it('throw with long command', () => {
+      const framer = new Framer('regtest');
+      let err = null;
+
+      try {
+        framer.packet('abcdefghijklm', Buffer.alloc(2, 0x00));
```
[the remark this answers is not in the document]
tynes: Good catch, will update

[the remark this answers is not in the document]
tynes: > We should also assert the upper bound on the size of the payload.

To do this, it would require allocating a buffer of ~4 GB.

It is only possible to allocate a buffer of ~2 GB using `Buffer.alloc` because of the max buffer length constant.

https://nodejs.org/api/buffer.html#buffer_buffer_constants_max_length

This value is not possible to alter, an error is thrown when it is altered.

I feel like 4 GB max size leaves open a dos vector where large packets are spammed and held in memory concurrently, causing an oom error.

lib/net/packets.js
```diff
@@ -1839,11 +1839,12 @@ class UnknownPacket extends Packet {
   /**
    * Inject properties from serialized data.
    * @private
-   * @param {Buffer} data
+   * @param {BufferReader} data
```
tynes: replace `data` with `br`

tynes: Fixed

boymanjor:
@tynes this needs to be rebased. Upon a local rebase, one of the Framer testcases failed. However, the coverage does get a bump. I'm not sure why the ci run is showing a downgrade. Relevant output is below:
```
  1410 passing (37s)
  1 pending
  1 failing

  1) Net
       Framer
         will construct with network (primary):

      AssertionError [ERR_ASSERTION]: [Network] strictEqual [Network]

      Expected "actual" to be reference-equal to "expected":

        [Network] {
      +   activationThreshold: 108,
      +   addressPrefix: 'rs',
      -   activationThreshold: 1916,
      -   addressPrefix: 'hs',
          block: {
          block: {
      +     keepBlocks: 10000,
      +     maxTipAge: 4294967295,
      -     keepBlocks: 288,
      -     maxTipAge: 43200,
            pruneAfterHeight: 1000,
      ...
          checkpoints: [],
      +   claimPrefix: 'hns-regtest:',
      -   claimPrefix: 'hns-claim:',
          coinbaseMaturity: 2,
      ...
              required: false,
      +       startTime: 0,
      -       startTime: 1199145601,
              threshold: -1,
              threshold: -1,
      +       timeout: 4294967295,
      ...
      -       timeout: 1230767999,
      ...

      assert.strictEqual(framer.network, Network.get('main'));
             ^

      at Context.it (/Users/boymanjor/code/hsd/test/net-test.js:737:14)

=============================================================================
Writing coverage object [/Users/boymanjor/code/hsd/coverage/coverage.json]
Writing coverage reports at [/Users/boymanjor/code/hsd/coverage]
=============================================================================

=============================== Coverage summary ===============================
Statements   : 61.28% ( 7843/12798 )
Branches     : 51.75% ( 2366/4572 )
Functions    : 57.31% ( 1043/1820 )
Lines        : 61.3% ( 7843/12795 )
================================================================================

```

tynes:
@boymanjor Rebased and found a side effect in a different test. Pushed up the changes

Issue #152, opened 2019-04-22
dumpzone RPC Call


tynes:
## dumpzone RPC

A Node RPC `dumpzone` that dumps the state of the Handshake Zone into a file could be useful for building DNS infrastructure on top of `hsd`.

By default, the file could be output in the zone file format.
The definition of a zone file is here - https://tools.ietf.org/html/rfc1035

## Arguments 

- filename - name of file to write to
- json - output the file in json
- detailed - add additional handshake specific information like auction state

Detailed mode should only be able to work with JSON output, since the point of outputting a zone file is to interoperate directly with DNS servers

kilpatty:
@tynes Thoughts about making this a plugin vs an RPC call? Perhaps both will work well, but I was thinking a plugin might give us more flexibility with configuration e.g. file format, write to disk every x minutes.

tynes:
I think that the base functionality should be a RPC. Then everybody has the functionality built into the daemon without needing to install another package for the plugin. I think that a plugin could be useful for extending the base functionality, in particular the write to disk every so often job. You can also get this functionality with a cron job or an event that `hs-client` listens to. I think the proper solution would depend on the particular security constraints. Getting feedback from people who run large scale DNS infrastructure or people that have a particular use case for this RPC would be helpful for helping us decide exactly how it should work.

My usecase involves dumping the state of the Handshake zone in JSON with extra details so that it is possible to make visualizations of the tree. I'd personally like to listen to an event using `hs-client`, maybe something like `tree update`, so that I can call `dumpzone /*json=*/true /*detail=*/true` right after the tree is updated

Bitcoin has a similar RPC method called `dumpwallet`
https://bitcoincore.org/en/doc/0.16.0/rpc/wallet/dumpwallet/

kilpatty:
Ah yes good point, I think plugin to extend base functionality would be best as well. 

Happy to help work on this if it's needed!

chjj:
@tynes, we have a full implementation of the zone file format in bns: 

- https://github.com/chjj/bns/blob/master/lib/internal/scan.js
- https://github.com/chjj/bns/blob/master/lib/internal/schema.js

As far as dumping to zone file goes, I have some code written for this currently (along with the new record types). Will push soon.

jacobhaven:
@chjj Encoding a name's wire data to a Zone file would make public resolvers like `hs.zone` a lot more reliable. Any update on that?

pinheadmz:
[...]

PR #280, opened 2019-10-16
WIP: node/http: dumpzone

tynes:
Experimenting with dumping the entire Handshake tree as a zonefile. This would allow a traditional DNS server to serve the Handshake zone. A tradeoff in security for scalability.

We need to figure out how often this would be called and ensure that its performant enough that it will not cause any problems with staying in consensus. It needs to be tested on a very large tree, planning on testing it on the current testnet.

Closes https://github.com/handshake-org/hsd/issues/152

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -9,6 +9,7 @@
 const assert = require('bsert');
 const path = require('path');
 const {Server} = require('bweb');
+const bns = require('bns');
 const Validator = require('bval');
 const {base58} = require('bstring');
 const {BloomFilter} = require('bfilter');
@@ -19,6 +20,9 @@ const util = require('../utils/util');
 const TX = require('../primitives/tx');
 const Claim = require('../primitives/claim');
 const Address = require('../primitives/address');
+const Records = require('../dns/records');
+const Resource = require('../dns/resource');
+const NameState = require('../covenants/namestate');
 const Network = require('../protocol/network');
 const pkg = require('../pkg');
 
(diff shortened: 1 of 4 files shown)


lib/node/http.js
```diff
@@ -404,6 +408,59 @@ class HTTP extends Server {
...
+            if (seen === 0)
+              res.write(`"${entry}"`);
+            else
+              res.write(`,"${entry}"`);
+
+            seen++;
+          }
+        }
+      }
+
```
tynes: Log number of records seen here

jacobhaven:
@tynes @chjj Do we know how this will change when RR types are removed for #125 ?

Also, to build a working public resolver, it would be really nice to support iterating every HS name that has changed since a given tree interval (potentially ALL HS names) and dumping them all as zones for a normal DNS NS to serve. We've been working towards this with @mikedamm

tynes:
@jacobhaven 

The `Resource` object should handle the `version` internally and not leak that to the consumer of the `Resource` object. If the `Resource` has a `version` that does not support a particular resource record type, it should not create any resource records in the `toDNS` method. In the case of the code in this PR, `msg.records()` should return an empty iterator: 

```javascript
const msg = resource.toDNS(fqdn, rrtype);

for (const rr of msg.records()) {
  const entry = rr.toString();
}
```

Note that this PR is naive in the sense that it uses brute force, there could be better ways to do it. Also see https://github.com/handshake-org/hsd/issues/125#issuecomment-547652205 as it is impossible to truly remove v0 serialization unless serialization versions become part of consensus.

jacobhaven:
@tynes I don't think doing this in a brute force way is really even that bad, as it's only used every 36 blocks. The problem you also need some way to enumerate all names that have changed since you last synced. It's not clear to me how to do that.

You can "truly remove" and old version of DNS serialization by just no longer resolving it. I doubt that simultaneously supporting multiple types of DNS Zones would be a good idea. Each resolver can just deprecate old ones as it sees fit.

tynes:
> The problem you also need some way to enumerate all names that have changed since you last synced.

This is a very interesting problem to solve, I think a closer inspection of the database would be helpful in solving this:

- https://github.com/handshake-org/urkel

Note that there is also an implementation written in Golang by Oasis Labs

- https://github.com/oasislabs/oasis-core/tree/master/go/storage/mkvs/urkel

----

> You can "truly remove" and old version of DNS serialization by just no longer resolving it.

This is an inherently subjective removal. The only "objective truth" is found through Nakamoto Consensus + Proof of Work, assuming minority byzantine miners (51% attack) and peers (eclipse attack). Whatever is on chain is on chain. Removing the ability to resolve (encode/decode the on chain data) a particular `Resource` version would effectively "brick" the usage of the name. The price of exit is pretty low, assuming that users can run their own full nodes or its relatively easy for new service providers to run a set of full nodes. Users will exit to the resolvers that best serve them.

pinheadmz:
Closing this to focus development on #534 thank you @tynes ! Handsome !

PR #281, opened 2019-10-16
wallet: include network in master key JSON serialization calls

jacobhaven:
Basically, since `toJSON` was incorrectly used instead of `getJSON`, all keys were being serialized with the default mainnet prefix instead of the prefix corresponding to their actual network.

### DIFF
--- a/lib/wallet/masterkey.js
+++ b/lib/wallet/masterkey.js
@@ -615,7 +615,7 @@ class MasterKey extends bio.Struct {
 
     return {
       encrypted: false,
-      key: unsafe ? this.key.toJSON(network) : undefined,
+      key: unsafe ? this.key.getJSON(network) : undefined,
       mnemonic: unsafe && this.mnemonic ? this.mnemonic.toJSON() : undefined
     };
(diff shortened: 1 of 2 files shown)


kilpatty:
Seems like there are quite a few toJSONs used instead of getJSON. Thoughts on expanding this to include all of the `toJSON` -> `getJSON` fixes? Seems like there might be quite a few more scattered throughout

On top of the fix in masterkey.js I see: 
https://github.com/handshake-org/hsd/blob/master/lib/wallet/masterkey.js#L619
https://github.com/handshake-org/hsd/blob/master/lib/wallet/masterkey.js#L632
https://github.com/handshake-org/hsd/blob/master/lib/wallet/masterkey.js#L635

jacobhaven:
@kilpatty I have no way of knowing which are correct and which aren't. These two cases are clearly incorrect and causing problems, though.

boymanjor:
@kilpatty I vote to merge this separately, as a bug fix, and wait for further exploration across the codebase.

Issue #276, opened 2019-10-11
Fee estimator doesn't account for covenant submarkets


rsolari:
The `hsd` [fee estimator](https://github.com/handshake-org/hsd/blob/4b04d9a2f9094b9c4c8e17346e7f24517f0a0336/lib/mempool/fees.js) is bcoin code that has not been updated for hsd, so it doesn't consider fee submarkets at all. 

By submarkets, I mean that there are four limits for how many transactions can go in a block, each of which could fill up separately:
- the main market for block space
- a submarket for [600 "updates"](https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/consensus.js#L200) which are `OPEN`, `UPDATE`, `TRANSFER` and `REVOKE` covenants
- a submarket for [600 "renewals"](https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/consensus.js#L208), which are `FINALIZE`, `REGISTER`, and `RENEW` covenants
- a sub-submarket for [300 opens](https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/consensus.js#L192), which are just `OPEN`s. Note that these 300 opens are a subset of the 600 updates

Note that the other side of the market, the miners who make blocks with the hsd template code, behave rationally. They [pick the highest-paying transactions](https://github.com/handshake-org/hsd/blob/7c172277476ea2a41501413f4f43a4a0a040b100/lib/mining/miner.js#L362) for each block that respect the covenant limits.

I'm planning to write a new `estimateFee()` method that takes covenant type as an argument and estimates fees to account for submarkets. Does that sound like a useful thing to add to fee estimation?

tynes:
Concept ack

Good explanation of the different markets. I agree with your logic.

It would be ideal if an additional class was added that implements the current fee estimator API but with an incremented version:

https://github.com/handshake-org/hsd/blob/a90a45bb9ea2c267e47bbaa173da1f181123f35f/lib/mempool/fees.js#L810

Then it would be possible to run both fee estimators and we can determine how the two perform relative to each other. Also good to show customers that new code saved them a specific amount of money.

rsolari:
Sounds good. I'll use `VERSION`.

> we can determine how the two perform relative to each other

[...]

PR #282, opened 2019-10-17
Covenant fee estimation

rsolari:
Adding covenant fees estimation, as mentioned in https://github.com/handshake-org/hsd/issues/276. 

There are two parts:

- The `ComposedEstimator` class, which keeps track of four estimators, one for each fee limit. Its job is to pass through arguments to the appropriate `PolicyEstimator`.
 - I made an interface change in `node/rpc.js` to show the type of changes required changes to fee estimation. This change is just an example. Ultimately, I'll have to change several of the places where something calls `estimateFee` and `estimatePriority`.

TODOs:

- [x] Pass the optional `type` argument to 'estimateFee()' and 'estimatePriority()' where appropriate.
- [x]  Replace the old `PriorityEstimator` with the new `ComposedEstimator`.
- [x]  Add tests
- [x] Fix build
~Change logging to show which estimator a log line comes from~


lib/node/rpc.js
```diff
@@ -2152,16 +2152,21 @@ class RPC extends RPCBase {
...
     const valid = new Validator(args);
     const blocks = valid.u32(0, 1);
+    const val = valid.str(1);
 
     if (!this.fees)
       throw new RPCError(errs.MISC_ERROR, 'Fee estimation not available.');
 
-    const fee = this.fees.estimateFee(blocks, false);
+    const type = rules.types[val];
+    if (!type)
```
turbomaze: bug: `rules.types: { [string]: number }` and `rules.types.NONE === 0` is falsey.

fix: `rules.types.hasOwnProperty(type)`

lib/node/rpc.js
```diff
@@ -2152,16 +2152,21 @@ class RPC extends RPCBase {
   }
 
   async estimateFee(args, help) {
-    if (help || args.length !== 1)
-      throw new RPCError(errs.MISC_ERROR, 'estimatefee nblocks');
+    if (help || args.length < 1 || args.length > 2)
+      throw new RPCError(errs.MISC_ERROR, 'estimatefee nblocks (covenantType)');
```
tynes: Using `( )` means that it is an optional argument. This means that the args length checker needs to be updated. If you want it to be a non optional argument, then you can simplify into `(help || args.length !== 2)`

See:

https://github.com/handshake-org/hsd/blob/7c4abc2938e37c4db8e8800f52d7386e2ddace38/lib/node/rpc.js#L719

Also can you add `"` around the arguments like: `"nblocks" ( "covenantType" )`

lib/node/rpc.js
```diff
@@ -2152,16 +2152,21 @@ class RPC extends RPCBase {
...
 
     const valid = new Validator(args);
     const blocks = valid.u32(0, 1);
+    const val = valid.str(1);
```
tynes: If its an optional argument, there needs to be a fallback value here

lib/node/rpc.js
tynes: Good catch. Another way to do this would be `if (typeof type !== number)`, so that it also checks on the type of `type`.

rsolari: Good point. Updated to check `typeof type !== number`

lib/node/rpc.js
rsolari: fixed!

lib/node/rpc.js
rsolari: I was relying on the behavior of `estimateFee`, which accepts a similar method signature, but you're right that I should be explicit here.

tynes:
Do you mind squashing your commits in a way that modularizes changes to different files and having commit messages like:

`package name: brief description`

See the commit history for examples.

I'm also on the fence about dropping the old fee estimator completely. It would definitely add some complexity regarding allowing configuration to select different ones. Since there is money involved, I think its a better idea to have to opt in the the new one until its battle tested, since the old one has been used to send a lot of value already. I also think it would be nice to expose both to the RPC, so that an additional optional argument could be passed that can select which fee estimator to use, since they are versioned. This would allow client side applications flexibility. What do you think?

rsolari:
Makes sense. I'll squash the commits down.

I'm hesitant to expand the interface to let users pick their estimator because – in theory – the estimators should behave the same. Would it be more convincing if I wrote some unit tests showing equivalence?

lib/mempool/composedFees.js
```diff
@@ -0,0 +1,290 @@
...
+
+  /**
+   * Estimate a priority.
+   * @param {Number} [target=1] - Confirmation target.
+   * @param {Boolean} [smart=true] - Smart estimation.
+   * @param {String?} - Covenant Type.
+   * @returns {Number}
+   */
+
+  estimatePriority(target, smart, covenantType) {
```
pinheadmz: I think `estimatePriority` is deprecated and removed in Bitcoin Core:

https://github.com/bitcoin/bitcoin/blob/452bb90c718da18a79bfad50ff9b7d1c8f1b4aa3/doc/release-notes/release-notes-0.15.0.md#removal-of-coin-age-priority

In bcoin and hsd, it's still exposed by the API but not used anywhere like `estimateFee` is...

So, maybe this is a bigger discussion, but we can probably forget this for now.

lib/node/rpc.js
```diff
@@ -2152,16 +2152,22 @@ class RPC extends RPCBase {
   }
 
   async estimateFee(args, help) {
-    if (help || args.length !== 1)
-      throw new RPCError(errs.MISC_ERROR, 'estimatefee nblocks');
+    if (help || args.length < 1 || args.length > 2)
```
pinheadmz: We are discussing the return values of `estimateFee` in bcoin right now: https://github.com/bcoin-org/bcoin/pull/889

It's worth noting that Bitcoin Core returns a JSON object for this PRC: https://chainquery.com/bitcoin-cli/estimatesmartfee

And my opinion is that for hsd, we should also be returning a JSON object -- with all the covenant types listed. (And maybe either remove the `type` argument, or add a NEW API or RPC call that specifies `type`?)

I think this may be the most useful to enterprise hsd full nodes, especially given the possibility that covenant types can be combined and batched by large businesses:

```
{
  blocks: 2,
  rate: {
    NONE: 10000,
    OPEN: 15000,
    RENEW: 17589,
    UPDATE: 23000
}
```

pinheadmz:
There is also a `/fee` endpoint in `lib/node/http.js` that should be addressed. 

Note the HTTP module has a RESTful API endpoint like you'd expect:

https://github.com/handshake-org/hsd/blob/master/lib/node/http.js#L408

But ALSO has a websocket call:

https://github.com/handshake-org/hsd/blob/master/lib/node/http.js#L577

...this is used by a wallet node when it is run as a separate process (not a plugin). Although if oyu notice in the bcoin PR #889, we are debating switching that data source from the socket call to the HTTP endpoint.

tynes:
> I'm hesitant to expand the interface to let users pick their estimator because – in theory – the estimators should behave the same. Would it be more convincing if I wrote some unit tests showing equivalence?

In theory is different than in practice, and that difference is important when sending value across the internet. Some unit tests showing equivalence would be helpful.

lib/node/rpc.js
rsolari: That return value looks better. I'm not sure what `hsd`'s policy is on breaking changes, but I'm happy to change the signature in this PR if that change wouldn't make this PR harder to merge.

rsolari:
Ok, great. Thanks for the feedback guys. I'll make some changes:
 - [ ] squash commits into ~one per file
 - [ ] add equivalence tests
 - [ ] update websocket fee call
 - [ ] check for other fee calls
 - [ ] ¿change the API to return JSON?

rsolari:
I haven't forgotten about this PR! I just haven't had time to finish it. I expect to clean it up and make the suggested changes eventually. Perhaps in a month or so.

I haven't prioritized it because it's not a consensus change, and fees don't matter until there's congestion.

nodech:
Closing due to inactivity.

PR #284, opened 2019-10-21
dns/records: add cosmoshub support to Addr record

tynes:
This PR adds native support for Cosmoshub addresses with human readable parts `cosmos`, `cosmosvalcons` and `cosmosvaloper`. With native support, on chain serialization sizes are decreased by 36-40%. 

Unknown currencies are saved on chain in ascii format as currency:address. Adding 'cosmoshub' as a known currency will save space for saving these types of addresses on chain, as the hrp will be encoded in a field instead of saving the ascii to chain. This is particularly helpful for cosmoshub addresses, as the hrp is relatively large compared to other chains.

The sizes below are the on chain serialization sizes in bytes for a single `Addr` record

```

hrp           | size before  | size after |
--------------|--------------|------------|
cosmos        | 59           | 24         |
cosmosvalcons | 66           | 24         |
cosmosvaloper | 66           | 24         |

This reduces the size of the cosmoshub addresses
on chain to 36-40% of the size.

```

These calculations were created using the code snippet below.
I don't think that the address data for cosmoshub is a 20 byte blake2b of a secp256k1 key, but that shouldn't make a difference in the relative space savings.

```javascript
const Records = require('hsd/lib/dns/records');
const {bech32} = require('bstring');
const secp256k1 = require('bcrypto/lib/secp256k1');
const blake2b = require('bcrypto/lib/blake2b');

const hrps = [
  'cosmos',
  'cosmosvalcons',
  'cosmosvaloper'
];

const version = 0;

const priv = secp256k1.privateKeyGenerate();
const pub = secp256k1.publicKeyCreate(priv);
const hash = blake2b.digest(pub, 20);

for (const hrp of hrps) {
  const address = bech32.encode(hrp, version, hash);
  const record = new Records.Addr('cosmoshub.', address);
  console.log(`size: ${record.getSize()}`);
}
```

### DIFF
--- a/lib/dns/records.js
+++ b/lib/dns/records.js
@@ -972,6 +972,24 @@ class PGP extends Struct {
 /**
  * Addr
  * @extends {Struct}
+ *
+ * The binary serialization begins with
+ * a Uint8 which represents the type.
+ * The currently defined types are:
+ *
+ * 1 - Handshake
+ * 2 - Bitcoin bech32
+ * 3 - Ethereum
+ * 4 - Cosmoshub
+ * 0 - Unknown
+ *
+ * For the defined types, some metadata
+ * is used to encode the human readable
+ * prefix for bech32 based addresses
+ * along with the size of the data.
+ *
+ * For the unknown type, it is serialized
+ * in ascii as currency:address.
  */
 
 class Addr extends Struct {
@@ -996,6 +1014,11 @@ class Addr extends Struct {
 
     if (util.equal(this.currency, 'ethereum.'))
       return 1 + 20;
+
+    if (util.equal(this.currency, 'cosmoshub.')) {
+      const {hash} = bech32.decode(this.address);
+      return 1 + 1 + 1 + 1 + hash.length;
+    }
 
     return 1 + sizeName(this.currency) + 1 + this.address.length;
   }
@@ -1043,6 +1066,33 @@ class Addr extends Struct {
       return this;
     }
 
+    if (util.equal(this.currency, 'cosmoshub.')) {
+      const {hrp, version, hash} = bech32.decode(this.address);
+
+      assert(hash.length >= 1);
+      assert(hash.length <= 256);
+
+      let field = 0x00;
+      if (/val/.test(hrp))
+        field |= 0x80;
+      if (/oper/.test(hrp))
+        field |= 0x40;
+      if (/cons/.test(hrp))
+        field |= 0x20;
+      if (/pub/.test(hrp))
+        field |= 0x10;
+
+      const size = hash.length - 1;
+
+      bw.writeU8(4);
+      bw.writeU8(field);
+      bw.writeU8(size);
+      bw.writeU8(version);
+      bw.writeBytes(hash);
+
+      return this;
+    }
+
     bw.writeU8(0);
     writeNameBW(bw, this.currency, c.labels);
     bw.writeU8(this.address.length);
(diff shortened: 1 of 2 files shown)


lib/dns/records.js
```diff
@@ -972,6 +972,25 @@ class PGP extends Struct {
...
+ *
+ * The binary serialization begins with
+ * a Uint8 which represents the type.
+ * The current types are as of follows:
+ *
+ * 1 - Handshake
+ * 2 - Bitcoin bech32
+ * 3 - Ethereum
+ * 4 - Cosmoshub
+ * 0 - Undefinded
```
tynes: Typo: `undefinded`. The word `Other` is more appropriate here

tynes:
There are 3 additional hrps that I did not include as part of this pull request. It would be fairly straightforward to add them using the unused bits in the field, but I was unsure of the differences between these and the ones without the `pub` suffix

- `cosmospub`
- `cosmosvalconspub`
- `cosmosvaloperpub`

Information sourced from:
https://github.com/cosmos/cosmos-sdk/blob/bb6bed4ed39ea52b0a42c81dd5dc4547dafc54e8/docs/spec/addresses/bech32.md

tynes:
I think that its possible to be smarter about encoding the hrp in the field.

Since all of the hrps contain the prefix `cosmos`, you can assume that it will be included. There are 4 other possible suffixes, that can be represented with a single bit.

- `val`
- `oper`
- `cons`
- `pub`

Reading the bits left to right, a 1 could mean append the corresponding word to the prefix `cosmos`. It should be possible to build all possible permutations of the hrp using a single nibble.

This unfortunately doesn't leave enough space to represent the size using just a u8, as the 4 bits can only represent a data length of 16. For the other addr currency types, 7 bits are used to represent the data length, which allows for a data length of up to 128.

lib/dns/records.js
```diff
@@ -1043,6 +1067,32 @@ class Addr extends Struct {
       return this;
     }
 
+    if (util.equal(this.currency, 'cosmoshub.')) {
+      const {hrp, version, hash} = bech32.decode(this.address);
+
+      assert(hash.length >= 1);
+      assert(hash.length <= 128);
+
+      let flag = 0x00;
```
boymanjor: Are there expected to be more than 16 hrps? If not, I think we can fit this encoding into a nibble and save a byte?

boymanjor: Scratch this; my mistake. While we can save 4 bits, @tynes reminded me we still need to encode a between 1 and 128 for the address length. We'll still need 2 bytes.

boymanjor: https://github.com/handshake-org/hsd/pull/284#issuecomment-545093156

tynes:
Implemented https://github.com/handshake-org/hsd/pull/284#issuecomment-545093156 in 86b3b906faba3f4b44d5c6762338f60655f1a52e

Split the field into field and size, both are a u8 instead of making the serialization more complex by using a u16. Added test cases for each of the different types of cosmoshub addresses.

One remaining question is if it should be `cosmoshub` or `cosmos-hub` or `cosmos` as the currency type.

Querying the DNS server would look like this, where `name` is the Handshake name and assuming `hsd` is running locally:
```bash
$ dig @127.0.0.1 -p 15360 +dnssec name URI
```

tynes:
Any changes will also need to be updated here: https://github.com/handshake-org/hnsd/blob/master/src/resource.c#L479

tynes:
Rename from cosmoshub to gaia?

test/records-test.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * records-test.js - Records tests for hsd
+ * Copyright (c) 2019, Mark Tyneway (MIT License).
+ * https://github.com/handshake-org/hsd
+ */
+
+'use strict';
+
```
tynes: @boymanjor If I port this file to another PR without the cosmos additions, do you think we could merge it in? Having the tests in would start the `records-test` file and allow it to be added to without needing this PR to be merged. I could rebase the cosmos additions back on top and then go from there.

tynes:
We need to know the max and min sizes of the bech32 data to be able to move forward with this.

hxrts:
Hey, we were just speaking about a HNS integration into a cosmos hub chain name service. Looks like this could be useful. Any way we can help?

pinheadmz:
@hxrts this particular PR is so old it's actually incompatible with main net HNS ;-)

would you like to join us in HNS developer telegram and talk about alternatives?

https://t.me/hns_tech

If all you need is a way to link cryptocurrency addresses with a HNS domain we have HIP-2 as a proposed standard:

https://hsd-dev.org/HIPs/proposals/0002/

PR #286, opened 2019-10-22
dns/record: fix ssh record object property

### LINKED ISSUE #285: SSH Record Bug
There is a mismatch in object properties for the `SSH`/`SSHFP` record fields between `hsd` and the `bns` (dns protocol) wire specification.

This is the definition of the Handshake `SSH` record:
https://github.com/handshake-org/hsd/blob/7c4abc2938e37c4db8e8800f52d7386e2ddace38/lib/dns/records.js#L870-L876

The codepath for rendering a `SSH` record follows this codepath below. Note that the `rd` variable is an instance of a `SSHFPRecord`, which is defined in `bns/lib/wire`. The constructor is here as well.
 
https://github.com/handshake-org/hsd/blob/7c4abc2938e37c4db8e8800f52d7386e2ddace38/lib/dns/resource.js#L810-L820

```javascript
class SSHFPRecord extends RecordData {
  constructor() {
    super();
    this.algorithm = 0;
    this.digestType = 0;
    this.fingerprint = DUMMY;
  }
```

`rd.keyType` is set in `hsd` (line 819), but it is not a property defined by the `SSHFPRecord`. It has the property `digestType` instead. The `digestType` is what is serialized into its wire format, so this value will always be `0` when returned by `hsd`.

Looking at the wikipedia article for `SSHFP` records, it follows the object defined in `bns` and not `hsd`.
https://en.wikipedia.org/wiki/SSHFP_record

This leads me to believe that the `hsd` `SSH` record needs to alter the `keyType` property to being `digestType` and properly assign it when `Record.toSSHFP`

Link to source:
https://github.com/chjj/bns/blob/6d517b4b4dbdab1fe289347e9b1ebd79e9b0dcb5/lib/wire.js#L3924

tynes:
The Handshake `SSH` record type was defined to have a `keyType`, but the `bns` based `SSHFPRecord` was defined with a `digestType`.

This commit changes the `keyType` property to being called `digestType`, as the data was not being properly assigned in `hsd/lib/resource` when the `Resource` was being rendered into a `sshfp` DNS request.

Closes #285

### DIFF
--- a/lib/dns/records.js
+++ b/lib/dns/records.js
@@ -871,7 +871,7 @@ class SSH extends Struct {
   constructor() {
     super();
     this.algorithm = 0;
-    this.keyType = 0;
+    this.digestType = 0;
     this.fingerprint = DUMMY;
   }
 
@@ -883,7 +883,7 @@ class SSH extends Struct {
 
   write(bw) {
     bw.writeU8(this.algorithm);
-    bw.writeU8(this.keyType);
+    bw.writeU8(this.digestType);
     bw.writeU8(this.fingerprint.length);
     bw.writeBytes(this.fingerprint);
     return this;
@@ -891,7 +891,7 @@ class SSH extends Struct {
 
   read(br) {
     this.algorithm = br.readU8();
-    this.keyType = br.readU8();
+    this.digestType = br.readU8();
     const size = br.readU8();
     assert(size <= 64);
     this.fingerprint = br.readBytes(size);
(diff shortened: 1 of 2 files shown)


kilpatty:
Nice catch, do we have tests for all the record types? I'd be happy to work on some of those if not, since I'm in the process of converting this stuff over to Rust.

tynes:
@kilpatty There are not tests that directly cover the serialization and deserialization of all the Records, but https://github.com/handshake-org/hsd/pull/265 indirectly covers it.

I started a test file in https://github.com/handshake-org/hsd/pull/284, following a similar pattern as in https://github.com/handshake-org/hsd/pull/277

PR #287, opened 2019-10-23
Elligator Squared (WIP)

chjj:
Explanation in the comments. Wondering if the anti-censorship benefits outweigh the performance hit. I have yet to see anyone using elligators in production, so this is sort of unknown territory for most everyone.

This is meant to be used with the master branch of bcrypto. See:

- https://github.com/bcoin-org/bcrypto/blob/master/lib/js/elliptic.js#L966
- https://github.com/bcoin-org/bcrypto/blob/master/src/extra256k1/elligator.h

### DIFF
--- a/lib/net/brontide.js
+++ b/lib/net/brontide.js
@@ -9,6 +9,95 @@
  *   Copyright (C) 2015-2017 The Lightning Network Developers
  *   https://github.com/lightningnetwork/lnd/blob/master/brontide/noise.go
  *   https://github.com/lightningnetwork/lnd/blob/master/brontide/noise_test.go
+ *
+ * We modify the LN handshake in the following way:
+ *
+ *   Definition: "elligator" -- an invertible algebraic
+ *   function which maps a field element to a point. We use
+ *   the _generalized_ definition here. Think tissue, not
+ *   kleenex.
+ *
+ *   For better indistinguishability and censorship mitigation,
+ *   we use an Elligator Squared construction as described by
+ *   Tibouchi[1][2]. As Tibouchi notes, elligators are _more_
+ *   useful on prime order curves. Any curve with torsion
+ *   groups must necessarily be distinguishable, as public
+ *   keys reside in the primary subgroup (though, this may
+ *   not be true of the Ristretto encoding). An adversary
+ *   can simply run the forward map and check whether the
+ *   resulting point is in the primary subgroup.
+ *
+ *   We're in a unique position here, since we're already
+ *   hell bent on using secp256k1 for compat with bitcoin,
+ *   as opposed to something like ed25519. The fact that
+ *   secp256k1 has a cofactor of 1 allows us to use a truly
+ *   indistinguishable encoding.
+ *
+ *   The Elligator Squared construction is such that the
+ *   sender must provide the receiver with two preimages
+ *   (u, v) which map to a point via `f^2(u,v) -> f(u) + f(v)`
+ *   where `f(u)` is the forward map. The inverse map f^2^-1(p)
+ *   is more complex and involves a loop as well as random
+ *   group element generation.
+ *
+ *   Performance may be a concern. Encoding a point requires
+ *   16 inversions and 20 square roots on average (an average
+ *   of 4 iterations of the elligator squared loop). This
+ *   should be on the order of ~2-3 point multiplications.
+ *   In the future, the inversions can likely be optimized
+ *   and reduced if multiple iterations are attempted at
+ *   once.
+ *
+ *   On its face, it seems like this loop should be the
+ *   subject of timing attacks. An adversary may run the
+ *   forward map and gauge their own timing with our's.
+ *   This is not the case for one simple reason: the
+ *   timing is _random_ with elligator squared, as a
+ *   random group element is generated on each iteration.
+ *
+ *   The forward map is cheaper, requiring only 6 square
+ *   roots and 3 inversions. We can visualize this as decoding
+ *   _nine_ public keys instead of one.
+ *
+ *   @sipa proposed what I'm referring to as the "unsquared"
+ *   version of SVDW inversion[3], but this does not
+ *   necessarily save on performance as a Legendre symbol
+ *   has to be computed for every call to the forward map:
+ *   this is roughly as expensive as the square root it
+ *   replaces (furthermore, @sipa's inverse function and
+ *   distribution seem to be incorrect (?)).
+ *
+ *   We can potentially double the performance by moving to
+ *   a curve which is 3-isogenous to secp256k1[4] (idea from [5]).
+ *   This allows us to use the Simplified Shallue-Woestijne-Ulas
+ *   map instead of the Shallue-van de Woestijne map, though,
+ *   the sender may have to modify their private key before
+ *   executing the actual ECDH (as their public key would
+ *   be multiplied by 3 by the receiver).
+ *
+ *   All bets are off if the user is MITM'd, or rather _tricked_
+ *   into connecting to an adversary via their published
+ *   identity key. A honeypot attack like this will no doubt
+ *   prove that this particular exchange of messages is that
+ *   of an end-to-end encryption handshake. However, the
+ *   construction is here to provide a mitigation against
+ *   more passive attacks, like non-interactive packet
+ *   inspection, not MITMs (see also: the great firewall).
+ *
+ *   A future possibility may be to randomly pad out messages
+ *   to avoid suspicion from a tyrannical government
+ *   (todo: research/discuss this more).
+ *
+ *   We have an elligator squared implentation in both C and
+ *   javascript[6][7].
+ *
+ *   [1] https://eprint.iacr.org/2014/043.pdf
+ *   [2] https://www.di.ens.fr/~fouque/pub/latincrypt12.pdf
+ *   [3] https://gist.github.com/sipa/29118d3fcfac69f9930d57433316c039
+ *   [4] https://github.com/bcoin-org/bcrypto/blob/master/scripts/secp256k1.sage
+ *   [5] https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/issues/158
+ *   [6] https://github.com/bcoin-org/bcrypto/blob/master/src/extra256k1/elligator.h
+ *   [7] https://github.com/bcoin-org/bcrypto/blob/master/lib/js/elliptic.js
  */
 
 'use strict';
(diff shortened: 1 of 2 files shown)


lib/net/brontide.js
```diff
@@ -9,6 +9,95 @@
...
+ *   to avoid suspicion from a tyrannical government
+ *   (todo: research/discuss this more).
+ *
+ *   We have an elligator squared implentation in both C and
+ *   javascript[6][7].
+ *
+ *   [1] https://eprint.iacr.org/2014/043.pdf
+ *   [2] https://www.di.ens.fr/~fouque/pub/latincrypt12.pdf
+ *   [3] https://gist.github.com/sipa/29118d3fcfac69f9930d57433316c039
+ *   [4] https://github.com/bcoin-org/bcrypto/blob/master/scripts/secp256k1.sage
```
pinheadmz: `404`... did you mean https://github.com/bcoin-org/bcrypto/blob/master/scripts/elligator.sage ?

pinheadmz:
If an attacker is watching traffic, knows this algorithm and _suspects_ (perhaps by port number) a server is sending EC points, they would be able to `secp256k1.publicKeyFromHash()` the packet and confirm their suspicion?

kilpatty:
@pinheadmz from my understanding I think the answer is yes -> Looks like this is more so that it would be difficulty for them to identify which packet potentially contains an EC point from a mass surveillance deep packet inspection. 

If there are other reasons to suspect a connection is initiating a handshake, calling publicKeyFromHash would identity that public key. So if they are just mass inspecting packets, calling publicKeyFromHash on every single packet of random bytes would be infeasible. 

Basically security through obscurity I think

tynes:
Avoiding censorship via packet analysis is an evolutionary arms race as each side successively gains an advantage through some sort of innovation. I think its doubtful that a single solution will ever work in all cases permanently. All traffic could be blocked on the standard Handshake network port, but a solution to that would be to run `hsd` on port 443, so that it blends in with TLS internet traffic. This wouldn't be perfect because there are statistical patterns to the traffic, ie after a block is found. Randomly padding the packets with a few bytes seems like low hanging fruit, although it is a tradeoff of using more bandwidth. Maybe erlay can be adopted for Handshake in the future.

lib/net/brontide.js
chjj: No, just a file I forgot to commit. Better explanation here now:

- https://gist.github.com/chjj/09c447047d5d0b63afcbbbc484d2c882
- https://github.com/bcoin-org/bcrypto/commit/aac4464c507c823ebcf62501d46d8f427aab4db2

chjj:
> If an attacker is watching traffic, knows this algorithm and suspects (perhaps by port number) a server is sending EC points, they would be able to secp256k1.publicKeyFromHash() the packet and confirm their suspicion?

No. `publicKeyFromHash()` maps _any_ 64 byte chunk of data to a public key. The trick is making it uniformly distributed such that the actual encoded pubkey looks like random data (or maybe it's better to say that the distribution of the forward map should look the same as if random data was passed in). This is where elligator squared's rejection sampling comes in.

> If there are other reasons to suspect a connection is initiating a handshake, calling publicKeyFromHash would identity that public key. So if they are just mass inspecting packets, calling publicKeyFromHash on every single packet of random bytes would be infeasible.

See above.

> Avoiding censorship via packet analysis is an evolutionary arms race as each side successively gains an advantage through some sort of innovation. I think its doubtful that a single solution will ever work in all cases permanently.

Solutions are a red herring. I'm looking for mitigations, not solutions. If we were only willing to settle for a perfect solution, we may as well drop encryption entirely since e2e encryption does nothing for privacy if a user is sybil'd. Even if it's not perfect, I _want_ an attacker to have to put in the extra time and effort. This setup basically requires an adversary to set up nodes in order to catch people in the act of resolving names. It's probably worth making their lives harder.

---

I ran some benchmarks. `secp256k1.publicKeyFromHash` is much faster than a point multiplication, and `secp256k1.publicKeyToHash` is on the order of 4 point multiplications (I'm getting 4k+ ops per second on an i7, so maybe 250us per call).

tynes:
The largest risks of introducing this into the codebase are that it could add a DoS vector and that it makes it more difficult for an additional implementation to be created.

Adding this will make the brontide handshake ~4x more expensive. This handshake must occur every time a new peer is added on the peer to peer network.

There is a difference between inbound and outbound connections, inbound are initiated by the remote node while outbound are initiated by the node itself.

A full node is in control of their inbound and outbound connection count.
See:
https://github.com/handshake-org/hsd/blob/86eb05328c4afd43ad88d8bd4c3f7608068b6ec6/lib/node/fullnode.js#L94-#L95

A user must pass the `--listen` flag to enable inbound connections. It is off by default.

This means that a node is safe from too many nodes trying to connect as inbound peers at the same time. By default, the node tries to maintain 8 outbound connections and will automatically add new peers when one drops off or is banned.

Would it be possible for an entity to spin up a node (or a fleet) and have them connect inbound to an `hsd` node that is known to accept inbound connections, complete the brontide handshake, do something to get banned, switch identity keys, and repeating the process of brontide handshake, getting banned and switching identity keys? I could see this being a potential denial of service attack.

tynes:
@chjj If the elligator implementation in bcrypto is stable, this could use a rebase/version bump of bcrypto so that everybody can see the tests passing.

tynes:
Also wondering about the removal of the version byte in the handshake. Keeping it in plaintext would defeat the purpose of using elligator, but it would make future p2p handshake upgrades more difficult. An upgrade could just look for a different "shape" of data coming across the wire, but that does not feel like the cleanest solution. Would it be useful to still send the version but in an encrypted fashion?

turbomaze:
> Also wondering about the removal of the version byte in the handshake. Keeping it in plaintext would defeat the purpose of using elligator, but it would make future p2p handshake upgrades more difficult. An upgrade could just look for a different "shape" of data coming across the wire, but that does not feel like the cleanest solution. Would it be useful to still send the version but in an encrypted fashion?

We could use some drastically overdetermined linear error correcting code for this. E.g. 1 byte message (the version byte) to 32 byte codewords. On each connection, the node can randomly sample a codeword that maps to their intended version byte. This has the property that the bits used to communicate version information are a) different each time and b) uniformly distributed.

tynes:
> Would it be possible for an entity to spin up a node (or a fleet) and have them connect inbound to an hsd node that is known to accept inbound connections, complete the brontide handshake, do something to get banned, switch identity keys, and repeating the process of brontide handshake, getting banned and switching identity keys? 

This attack is limited by the amount of IP addresses the attacker can control because bans are based on the IP address and not the key.

https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/net/pool.js#L3423-L3428

https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/net/netaddress.js#L46

This security model inherently assumes that an attacker cannot control an arbitrary amount of IPs. Is this a fair assumption to make?

tynes:
> We could use some drastically overdetermined linear error correcting code for this. E.g. 1 byte message (the version byte) to 32 byte codewords

This is a really interesting idea. If I understand correctly, it would map the version number to something that looks like randomness and then can then be mapped back to the same integer correctly.

We would want to be certain that there is no possibility for denial of service attacks in the implementation of the decode function. This would also add an additional 32 bytes of overhead in the handshake, but in return would allow for an easier way to upgrade the p2p handshake over time if need be.

turbomaze:
> > We could use some drastically overdetermined linear error correcting code for this. E.g. 1 byte message (the version byte) to 32 byte codewords
> 
> This is a really interesting idea. If I understand correctly, it would map the version number to something that looks like randomness and then can then be mapped back to the same integer correctly.
> 
> We would want to be certain that there is no possibility for denial of service attacks in the implementation of the decode function. This would also add an additional 32 bytes of overhead in the handshake, but in return would allow for an easier way to upgrade the p2p handshake over time if need be.

I mention LEC because I think it's the purest/most efficient `large space -> small space` mapping. (Alternatively you could do something like el gamal with a publicly-known private key, to similar effect etc etc). Would have to think a little more about sampling a random codeword for a message, but decoding LEC is basically a finite field matrix multiply (could just do GF(2) in which case the matrices are just bits and you use bit tricks to be really fast).

troq:
I'm not convinced we need this before mainnet. At the very least it shouldn't delay mainnet. In addition to treading new ground, there are likely other lower hanging fruit attacks that can be exploited on Handshake before the attack that elligators protect against. It'll likely be a long time before enough people care about Handshake for a state actor to try to censor it. Handshake is much more at risk of failing because no one cares than for failing from a state actor attack.

pinheadmz:
Looks like the size of ACT_ONE_SIZE changes if Elligator is implemented. I wonder if it could be merged in the future based entirely on that?

```
if (actOne.length == ACT_ONE_LEGACY_SIZE)
  recvActOne_legacy(actOne);
else if (actOne.length == ACT_ONE_ELLIGATOR_SIZE)
  recvActOne_elligator(actOne);
else
  throw new Error('Act one: bad size.');
```

kilpatty:
@tynes 
> Would it be possible for an entity to spin up a node (or a fleet) and have them connect inbound to an `hsd` node that is known to accept inbound connections, complete the brontide handshake, do something to get banned, switch identity keys, and repeating the process of brontide handshake, getting banned and switching identity keys? I could see this being a potential denial of service attack.

Two thoughts here:
1. I think it may be worth our time to do some number crunching on just how much of a DoS vector this could be. In some of the papers about Eclipse attacks on Bitcoin, they mention how trivial it was to "rent" botnets. That being said, we don't even need a botnet attack here, but an "infrastructure attack" or onion address rolling would suffice, since we don't have any grouping of IPs (yet).  (https://eprint.iacr.org/2015/263.pdf)

2. I do have a PR for net groups, or at least the beginnings of them, so by implementing that fully, and building out some of the other protections in Address manager that Bitcoin has implemented, I think we can mitigate most of the risk from (1.).

tynes:
## Threat model

Assuming:
- An adversary can inspect network packets
- An adversary can notice key material in network packets
  - An adversary wants to know a user's public key
- Assumes that a user will not connect to any malicious peers
- An adversary wants to know if a user is resolving handshake names

Elligator would require the adversary to run the decode function
over all data it observes, and since it will always map data to
a public key, it gives plausible deniability to the user.

An adversary may be interested in knowing if a particular user
is resolving handshake names. It is unclear if elligator alone
would completely mitigate this, but it does give plausible
deniability and a strong foundation for hardening the p2p
protocol over time.

If the user reuses their p2p identity key, then the adversary can
notice repeated data going over the wire. This can be mitigated by
having users change their identity keys over time.

Assuming:
- An adversary wants to censor HNS name resolution by blocking traffic after detecting a p2p handshake

There are other ways to detect if a Handshake node is running, the protocol would have to be hardened against this security model over time.

## Version byte

Would it be possible to encode the version in the identity key itself? This would require some grinding of keys, as only some identity keys would be valid. Creating an RPC that can create a valid identity key would be useful for this, or having a cli tool to create identity keys.

## Compared to TLS

This website shows what the TLS handshake looks like.
https://tls.ulfheim.net/
Note that it sends the client public key in plaintext, so this is also a problem that TLS has.

It would be ideal if we could create a similar diagram to compare what the p2p handshake would like like with and without elligator.

pinheadmz:
> If the user reuses their p2p identity key, then the adversary can
notice repeated data going over the wire. This can be mitigated by
having users change their identity keys over time.

If I understand Brontode correctly, this already happens: the first message over the wire is a client sending a new ephemeral key to the server. The client already knows the server's static key and so the second message is encrypted by the server with DH(client e, server s)

kilpatty:
Just a quick suggestion here off the top of my head about the version byte. 

A. Do we really _need_ it?

B. Could we use the Protocol Name or Prologue as a Pseudo version byte? 
https://github.com/handshake-org/hsd/blob/master/lib/net/brontide.js#L32-L33

In a world where a second version of Brontide is released, a node could maintain 2 Brontide constructions in memory. The first one being the original, and the second being one created with say `PROLOGUE="hns_1"`. If the initial connection fails, the node can attempt to re-parse the sent acts with the second Brontide construction. 

I don't think this is a very efficient way of doing this, and like I said, I'm not entirely convinced we _need_ a version byte, but hoping that by suggesting this it can stir up some new ideas.

tynes:
> A. Do we really need it?

I would be ok with omitting the version byte, although thinking of Handshake as a 100 year+ protocol, it would be more clean to have some sort of version built in so that it could be upgraded via softfork. There may still be other ways of allowing for upgrade via softfork, but they may not be as clean.

boymanjor:
Merged at e0a2f7b35ccbd9e7406db14b7fa3756941dbd4e3.

PR #288, opened 2019-10-24
dockerfile: update, add docs

tynes:
Update the `Dockerfile` and add documentation on how to use it.

The node HTTP server and wallet HTTP server are both exposed.
The recursive resolver is exposed with both TCP and UDP.
The authoritative resolver is not exposed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,42 @@
-FROM node:alpine AS base
-RUN mkdir -p /code
-WORKDIR /code
-CMD "hsd"
+FROM node:10-alpine AS base
 
-RUN apk upgrade --no-cache && \
-    apk add --no-cache bash unbound-dev gmp-dev
+RUN mkdir /opt/hsd
+WORKDIR /opt/hsd
 
-COPY package.json \
-     #package-lock.json \
-     /code/
+# dynamically linked deps
+RUN apk upgrade --no-cache \
+  && apk add --no-cache unbound-dev
 
-# Install build dependencies and compile
 FROM base AS build
-RUN apk add --no-cache g++ gcc make python2
-RUN npm install --production
+
+# build deps
+RUN apk upgrade --no-cache \
+  && apk add --no-cache bash \
+  git python2 g++ gcc make
+
+COPY . .
+
+RUN npm install
 
 FROM base
-ENV PATH="${PATH}:/code/bin:/code/node_modules/.bin"
-COPY --from=build /code/node_modules /code/node_modules/
-COPY bin /code/bin/
-COPY lib /code/lib/
+
+ENTRYPOINT ["hsd"]
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,22 +1,42 @@
...
 
-RUN apk upgrade --no-cache && \
-    apk add --no-cache bash unbound-dev gmp-dev
+RUN mkdir /opt/hsd
+WORKDIR /opt/hsd
 
-COPY package.json \
-     #package-lock.json \
-     /code/
+# dynamically linked deps
```
kilpatty: Do we want dynamic deps? I thought the idea of keeping package.json here was to cache the builds so that if we don't update any dependencies we can skip the install step.

kilpatty:
This was definitely needed I think as the Dockerfile needed some love. My only other comment is that I've still experienced some slowness using dockerignore, that was apparently "fixed", but I know a handful of people who have experienced it as well. 

Just want to benchmark the build timings between the general copy command (with the larger ignore) and the specific copy command (with the smaller ignore) to see if it makes any noticeable difference.

tynes:
@kilpatty had some good comments in an out of band conversation regarding speeding up the docker build.

Step 1: Copy package.json (could be cached if no changes to deps)
Step 2: Npm Install (Could be cached if the above is cached, so if no changes to deps, we don't have to re-run install again)
Step 3: Copy Src

Dockerfile
tynes: These are native dependencies that have to be there during the build and at runtime. They won't change unless its a forced rebuild and `apk upgrade` finds a new version

afmsavage: @tynes since you are using the default node10 alpine image, best practices recommends using the package versions that the image comes with instead of running apk upgrade.  This allows the maintainer to test the image and the pkg versions before releasing it.  Recommend pinning the version numbers to the below command and also removing the `apk upgrade` command.

Dockerfile
```diff
@@ -1,22 +1,42 @@
...
+RUN apk upgrade --no-cache \
+  && apk add --no-cache unbound-dev
 
-# Install build dependencies and compile
 FROM base AS build
-RUN apk add --no-cache g++ gcc make python2
-RUN npm install --production
+
+# build deps
+RUN apk upgrade --no-cache \
```
afmsavage: Again, would remove the `apk upgrade` command from here as well.  If I'm being nitpicky, the packages being installed below should also have versions pinned to them, but these packages are pretty solid and I wouldn't expect anything out of the ordinary.

Dockerfile
```diff
@@ -1,22 +1,42 @@
...
+  && apk add --no-cache bash \
```
afmsavage: Why are you adding `bash`?  Is it just for easier debugging?

Issue #245, opened 2019-08-27
Proposal: Commit to airdrop tree roots in genesis block.


pinheadmz:
A funny thing about hsd is that there are a few bits of consensus-critical data in peripheral files. The airdrop tree root and faucet tree root are committed to in `lib/primitives/airdropproof.js`, and the ICANN key-signing keys are committed to in `node_modules/bns/lib/constants.js`. 

I understand the reason these are currently where they are, but it feels like since they are esentially part of the parameter set that defines the hsd blockchain, they should maybe also live in `lib/protocol/consensus.js` and I think they should be also be committed to somewhere in the genesis block.

I've been playing around with airdrop proofs and reserved name claims, and modifying these critical values so I can generate valid proofs in regtest. It made me wonder how possible it would be for an attacker to create a malicious hsd client that snuck in different tree root values and include extra airdrop outputs. Even six months into the chain, an updated client could include a modified tree -- it might hard-fork new nodes syncing for the first time -- but could also silently allow unauthorized airdrop proofs in new blocks, or at least hard-fork the network.

In sort of a related issue, it will be basically impossible for independent auditors to replicate and verify that the 83,000-key airdrop tree is honest. So I think anything we can do to prevent malleability of these data would be at least a confidence-builder.

chjj:
Agreed. Putting it in the coinbase seems sufficient. I think we'll regenerate the trees one more time before mainnet (probably in a few days or so).

PR #290, opened 2019-10-28
scripts: genesis commitment

tynes:
Adds a commitment to the airdrop merkle root, the faucet merkle root, the ICANN 2010 KSK DS record and the ICANN 2017 KSK DS record to the genesis block in the format:

```
blake2b_32(AIRDROP_ROOT || FAUCET_ROOT || DS_KSK_2010 || DS_KSK_2017)
```

This hash is used in a null data address (version 31, an opreturn) in the output at index 1. I am open to suggestions for alternative ways to serialize this data if another way makes more sense.

Each of these items are important to Handshake consensus and are now committed from the start.

It would be ideal to import the merkle roots from [hs-airdrop](https://github.com/handshake-org/hs-airdrop/blob/e92f14d00a27aa250d120b4d32786fea5dfc23b2/lib/proof.js#L19) to ensure that they do not fall out of sync, even though they should never change post mainnet.

Good thinking @pinheadmz for suggesting this, see discussion in #245 

Closes https://github.com/handshake-org/hsd/issues/245


pinheadmz:
Concept ACK, thanks for putting this together. Of course, it doesn't really accomplish anything unless there is some kind of verification as well. As mentioned in #245 these important data are scattered about the codebase. I feel like before they are loaded into memory, they should be checked against the commitment. Do you think that's necessary?

PR #292, opened 2019-10-30
dns/resource: onion{ng} dnssec

tynes:
Resource.toA and toAAAA will also add
TXT records to the response. This will
authenticate the Onion and OnionNG records.

Addresses comment https://github.com/handshake-org/hsd/pull/265#discussion_r340285696

### DIFF
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -1203,10 +1203,14 @@ class Resource extends Struct {
       case types.A:
         res.answer = this.toA(name);
         key.signZSK(res.answer, types.A);
+        if (this.hasTOR())
+          key.signZSK(res.answer, types.TXT);
         break;
       case types.AAAA:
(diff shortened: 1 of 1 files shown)


lib/dns/resource.js
```diff
@@ -1203,10 +1203,12 @@ class Resource extends Struct {
       case types.A:
         res.answer = this.toA(name);
         key.signZSK(res.answer, types.A);
+        key.signZSK(res.answer, types.TXT);
```
boymanjor: Without context, this code block may be strange to a reader. What do you think about adding a descriptive guard, e.g.:
```js
case types.A:
  res.answer = this.toA(name);
  key.signZSK(res.answer, types.A);
  if (this.hasTor())
    key.signZSK(res.answer, types.TXT);
  break;
```

PR #295, opened 2019-11-04
utils: fix amount to unit.

nodech:
Added small test vector.

Updated!

So changes:
 - Removed toBits/fromBits.
 - Apparently `uhns` was broken, it should be same as `doo`.
 - Added `handshake` as amount unit type as well.

Notes about types:
 - Added `@typedef` for unit enum: `AmountUnitType`

So we get currencies:
| Unit   | Decimal (HNS) | Exponent | Note              |
| ---    | ---           | ---      | ---               |
| `doo`  | 0.000001      | 0        | Base unit         |
| `uhns` | 0.000001      | 0        | micro-hns or μhns |
| `mhns` | 0.001         | 3        | milli-hns or mhns |
| `hns`  | 1             | 6        | or `handshake`    |

There's an issue with conflicting types: `lib/types` defines global typedefs for the project.
There's typedef: https://github.com/handshake-org/hsd/blob/2d1cbe9c17b0ad4e8858c06a8f85625dbee35ba9/lib/types.js#L66-L74

Which has same name as `Amount` class, so neither jsdoc nor tools can determine what is Amount.
So I can't actually define `@return {Amount}` meaning the class instead of typedef.

One solution would be to change name for either of them:
 e.g. call DooInteger or something like that to that typedef.

### DIFF
--- a/lib/ui/amount.js
+++ b/lib/ui/amount.js
@@ -15,7 +15,7 @@ const pkg = require('../pkg');
  * Amount
  * Represents a currency amount (base unit internally).
  * @alias module:currency.Amount
- * @property {Amount} value
+ * @property {Number} value
  */
 
 class Amount {
@@ -23,7 +23,7 @@ class Amount {
    * Create an amount.
    * @constructor
    * @param {(String|Number)?} value
-   * @param {String?} unit
+   * @param {AmountUnitType} [unit=doo]
    */
 
   constructor(value, unit) {
@@ -36,8 +36,8 @@ class Amount {
   /**
    * Inject properties from options.
    * @private
-   * @param {(String|Number)?} value
-   * @param {String?} unit
+   * @param {String|Number} value
+   * @param {AmountUnitType} [unit=doo]
    * @returns {Amount}
    */
 
@@ -62,7 +62,7 @@ class Amount {
 
   /**
    * Get base unit string or value.
-   * @param {Boolean?} num
+   * @param {Boolean} [num=false] - Return a number.
    * @returns {String|Amount}
    */
 
@@ -74,18 +74,8 @@ class Amount {
   }
 
   /**
-   * Get bits string or value.
-   * @param {Boolean?} num
-   * @returns {String|Amount}
-   */
-
-  toBits(num) {
-    return Amount.encode(this.value, 2, num);
-  }
-
-  /**
    * Get mhns string or value.
-   * @param {Boolean?} num
+   * @param {Boolean} [num=false] - Return a number.
    * @returns {String|Amount}
    */
 
@@ -95,7 +85,7 @@ class Amount {
 
   /**
    * Get currency string or value.
-   * @param {Boolean?} num
+   * @param {Boolean} [num=false] - Return a number.
    * @returns {String|Amount}
    */
 
@@ -105,20 +95,20 @@ class Amount {
 
   /**
    * Get unit string or value.
-   * @param {String} unit
-   * @param {Boolean?} num
+   * @param {AmountUnitType} unit
+   * @param {Boolean} [num=false] - Return a number.
    * @returns {String|Amount}
+   * @throws on incorrect unit type.
    */
 
   to(unit, num) {
     switch (unit) {
       case pkg.base:
-        return this.toBase(num);
       case `u${pkg.unit}`:
-      case 'bits':
-        return this.toBits(num);
+        return this.toBase(num);
       case `m${pkg.unit}`:
         return this.toMilli(num);
+      case pkg.unit:
       case pkg.currency:
         return this.toCoins(num);
     }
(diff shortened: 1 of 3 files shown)


test/utils-test.js
```diff
@@ -55,4 +78,43 @@ describe('Utils', function() {
     assert.strictEqual(fixed.fromFloat(0.15645647, 8), 15645647);
     assert.strictEqual(fixed.toFloat(15645647, 8), 0.15645647);
   });
+
+  it('should Amount from units', () => {
```
boymanjor: nit: "should _convert_ Amount from units"

tynes:
> I also wonder if we should remove the concepts of micro hns and bits.

Maybe we convert `bits` to `doos`.

boymanjor:
> Maybe we convert `bits` to `doos`.

@tynes to clarify, are you implying we should have `bits`, `micro hns`, and `doos` represent the same unit? I should add that technically micro hns isn't Bitcoin specific, it is a metric unit. Since we already have a unit for 1/1000000 of an HNS it seems redundant to keep micro hns. It is also worth mentioning that the micro hns unit is currently incorrect.

tynes:
Sorry if that was unclear. I was suggesting that the term `bits` is removed in favor of the term `doos`, since Bitcoin already uses `bits`. 
https://github.com/bitcoin/bips/blob/master/bip-0176.mediawiki

I'm not particularly opinionated on micro hns, but it should represent the correct value if it is included.

lib/ui/amount.js
```diff
@@ -119,7 +119,7 @@ class Amount {
         return this.toBits(num);
       case `m${pkg.unit}`:
         return this.toMilli(num);
-      case pkg.currency:
+      case pkg.unit:
```
pinheadmz: This could maybe accept both, they are synonymous ones just an abbreviation:

```
case pkg.unit:
case pkg.currency:
  return this.toCoins(num)
```

pinheadmz:
I agree that we should remove "bits" entirely from hsd, including:

https://github.com/handshake-org/hsd/blob/f9d00cec0dfb64fcafec01a5107dd6725823dcad/lib/ui/amount.js#L76-L84

"1 bit" in Bitcoin is this value: `0.00000100` (100 satoshis) the motivation I think being that "bits" would express Bitcoin units in a world-currency standard way (with two decimal places):

At this time, these values are approximately equal:
`$12.45`
`b1,531.35`

SO, the fact that Handshake uses 6 decimal places instead of 8 is going to make this even more confusing for people familiar with both (note: I always hated the "bits" unit 😬 )

If we want to express a name for something equal to 100 dollarydoos... may I suggest... something like...

`fingers` ?

`1 hns` = `1,000,000 doos` = `10,000.00 fingers`

boymanjor:
> Sorry if that was unclear. I was suggesting that the term bits is removed in favor of the term doos, since Bitcoin already uses bits.

@tynes I understand and agree with this.

nodech:
Updated description.

nodech:
Oh, actually my tools resolve `Amount` as the class when I am in the file, but jsdoc generated docs will resolve to global typedef )

pinheadmz (approved):
Thanks! Looks good. Built the jsdocs, ran the tests, played with `to` / `from` in REPL

ACK 61e278799c59916ce8b02c0d81eb3de021df41b2

```
Hash: SHA256

ACK 61e278799c59916ce8b02c0d81eb3de021df41b2
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmBsisMACgkQ5+KYS2KJ
yToHFhAApRNPpn29455o6lY96XOLTT4reKGn0ihgX/lR+uEZU6CcMFy38+ehtkN1
yPH5pINbpxXU1QPA+KuW8wnn30+zJYe670eQJA2Hvkrad1y9+ey6pDc/Aoh4mqUI
ldukT6WtTKG6N3l85sEPoMSGwjq0Vm42q1A1l/XHX70J8y1srxDV/qZfH81gzmOw
xXVukFk7seE/pjIaBC2QA9+ywoap2U+5+fuScyOBW8dcgl4In/lTfMIhIf1IwrYk
/oWdiWAhyYfQB/+FwH7KsUHfSAOTCICn6wKsn2N2niPsG3WLIIV6ruRNuf20ltp7
jw7rIv2ZqcimoV9sX/V10qqhcuFCyJLacn3+OjvfL9886ILEUZvhYHSMUmbRvu8/
kGCqjRPPEkFnMedMxOqCiITwRScvPxZP+43YdZdRUsgpTQMk3QFbdzcPIXE8ChVL
SIGDY7P4Xm0zaUuVKF2AQDhuH/lxGoqALfRe9MPCBVv9tG9h0VMSzMsaOP2lvgfh
74ymYNO6YvK8gP3Kf4kJ1EyyG/MxFt9jkoHmiDqJxU+zCCMc6J0yBJ4rWFa502te
5uRcTQODpuF3ZkwrQqONa4894avsNtqflCt4foi4BSJealMpB7xtFPhl8kGv3BTt
A/qGtX5psDFKzojokmLZ+N4sy1Sn8KvtS+e3qv6FDdn/mBwnP68=
=Mfp7
-----END PGP SIGNATURE-----
```

</details>

PR #297, opened 2019-11-04
test: block and memblock serialization

pinheadmz:
Cleans up `block-test.js` and ports the test from https://github.com/handshake-org/hsd/pull/249

### DIFF
--- a/test/block-test.js
+++ b/test/block-test.js
@@ -4,19 +4,72 @@
 
 'use strict';
 
-const {BloomFilter} = require('bfilter');
 const assert = require('bsert');
-const common = require('./util/common');
 const Block = require('../lib/primitives/block');
-const MerkleBlock = require('../lib/primitives/merkleblock');
-const consensus = require('../lib/protocol/consensus');
-const Network = require('../lib/protocol/network');
-const Script = require('../lib/script/script');
-const bip152 = require('../lib/net/bip152');
-const CompactBlock = bip152.CompactBlock;
-const TXRequest = bip152.TXRequest;
-const TXResponse = bip152.TXResponse;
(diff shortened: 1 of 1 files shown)


test/block-test.js
```diff
@@ -4,19 +4,43 @@
...
+        bits: 0,
+        nonce: 0,
+        extraNonce: Buffer.alloc(24, 5),
+        mask: Buffer.alloc(32, 6)
+      });
+      assert(block);
+    });
+
+    it('should deserialize and reserialze block', async () => {
+      raw = block.toRaw();
```
tynes: Nit: use `encode`/`decode` over `toRaw` and `fromRaw`

pinheadmz: thanks updated

pinheadmz:
Notice my dummy block has 0 transactions (not even a coinbase), looks like `Memblock` doesn't parse those anyway, and leaving them out makes this a more acute unit test (not having to include the `TX` module...) Does that seem ok?

pinheadmz:
Added a TX to the block and switched to `fromJSON()`.

test/block-test.js
```diff
@@ -4,19 +4,72 @@
...
+            outputs: [
+              {
+                value: 1234567,
+                address: 'rs1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqn6kda'
+              }
+            ]
+          }
+        ]
+      });
+      assert(block);
```
boymanjor: How do you feel about asserting the individual attributes on the block primitive?

pinheadmz: Good idea, done at 3419d7cfece14db95cc899f996a351c8527f8387

PR #302, opened 2019-11-18
ui: decouple consensus from ui

tynes:
This pull request defines constants for the min/max address version and the min/max address data size, allowing for more address versions and larger address data size.

The constants defined in `lib/protocol/consensus` are:

- `MAX_WITNESS_PROGRAM_VERSION = 255`
- `MIN_WITNESS_PROGRAM_VERSION = 0`
- `MAX_WITNESS_PROGRAM_SIZE = 64`
- `MIN_WITNESS_PROGRAM_SIZE = 2`

Previously, the witness program size was a consensus rule that lived in the `bech32` library.  [BIP 173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) defines the min and max values for Bitcoin's use of bech32, Lightning uses much longer bech32 encoded payment invoices.

From `lnd` source code:
```
// NOTE: This method it a slight modification of the method bech32.Decode found
// btcutil, allowing strings to be more than 90 characters.
```

To allow for this, I had to make a similar change to `bcrypto/lib/js/bech32` and vendor it in `hsd/lib/ui`. I am not sure if that is the best place for it to live, maybe it could live in `bcrypto` as a generalized bech32 implementation without any size restrictions.

See:
https://github.com/lightningnetwork/lnd/blob/4d7a151b4763f5893c91c4f6b49316391db17f73/zpay32/bech32.go#L12

Some considerations:

- 256 total address versions should be plenty although it could fracture the anonymity set
- 64 bytes in the address data should be enough for a stealth addressing scheme
- BLS pubkeys are 48 bytes
  - https://github.com/ethereum/eth2.0-specs/blob/dev/specs/bls_signature.md#g1-points
- Large enough to commit to two 32 byte hashes for `OP_RETURN` outputs (version=31)
- Should we allow a larger address data amount and then introduce a policy limit similar to bitcoin?
- Could remove `MIN_WITNESS_PROGRAM_VERSION` and just assume `0`

TODO:
- [x] Make sure all instances of min and max address size are accounted for, ie https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/covenants/ownership.js#L145

This PR also fixes a bug in the `tests/tx-test` where the incorrect bytes were being altered, the test still passed because the test asserts on a throw and it did throw but for the wrong reason. It was supposed to change the output value but it actually changed the address version. Previously it went unseen because it made the address version too large (which throws) but now it falls into the valid range, which no longer caused it to throw.

### DIFF
--- a/lib/primitives/address.js
+++ b/lib/primitives/address.js
@@ -199,16 +199,9 @@ class Address extends bio.Struct {
    * @returns {Address}
    */
 
-  fromPubkey(key, version) {
-    if (version == null)
-      version = 0;
-
-    if (version === 0) {
-      assert(Buffer.isBuffer(key) && key.length === 33);
-      return this.fromHash(blake2b.digest(key, 20), 0);
-    }
-
-    return null;
+  fromPubkey(key) {
+    assert(Buffer.isBuffer(key) && key.length === 33);
+    return this.fromHash(blake2b.digest(key, 20), 0);
   }
 
   /**
@@ -217,16 +210,9 @@ class Address extends bio.Struct {
    * @returns {Address}
    */
 
-  fromScript(script, version) {
-    if (version == null)
-      version = 0;
-
-    if (version === 0) {
-      assert(script && typeof script.encode === 'function');
-      return this.fromHash(sha3.digest(script.encode()), 0);
-    }
-
-    return null;
+  fromScript(script) {
+    assert(script && typeof script.encode === 'function');
+    return this.fromHash(sha3.digest(script.encode()), 0);
   }
 
   /**
@@ -251,15 +237,14 @@ class Address extends bio.Struct {
    * Inject properties from witness.
    * @private
    * @param {Witness} witness
-   * @param {Number} version
    */
 
-  fromWitness(witness, version) {
+  fromWitness(witness) {
     const [, pk] = witness.getPubkeyhashInput();
 
     if (pk) {
       this.hash = blake2b.digest(pk, 20);
-      this.version = version ? version : 0;
+      this.version = 0;
       return this;
     }
 
@@ -267,7 +252,7 @@ class Address extends bio.Struct {
 
     if (redeem) {
       this.hash = sha3.digest(redeem);
-      this.version = version ? version : 0;
+      this.version = 0;
       return this;
     }
 
@@ -314,16 +299,9 @@ class Address extends bio.Struct {
    * @returns {Address}
    */
 
-  fromPubkeyhash(hash, version) {
-    if (version == null)
-      version = 0;
-
-    if (version === 0) {
-      assert(hash && hash.length === 20, 'P2WPKH must be 20 bytes.');
-      return this.fromHash(hash, 0);
-    }
-
-    return null;
+  fromPubkeyhash(hash) {
+    assert(hash && hash.length === 20, 'P2WPKH must be 20 bytes.');
+    return this.fromHash(hash, 0);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


tynes:
Note that there is a comment in the code that says:

https://github.com/handshake-org/hsd/blob/f9d00cec0dfb64fcafec01a5107dd6725823dcad/lib/covenants/rules.js#L962

tynes:
@chjj The benefit to still using bech32 is that the addresses that have been generated will not need to be altered. This PR simply drops the size assertions in the bech32 library itself and moves them into `lib/protocol/consensus` along with allowing for a max witness program version of 255 and a max witness program size of 64 bytes.

I do not have strong feelings on what the max size should be, but 40 felt a bit restrictive when I was trying to develop the non interactive atomic swaps. We came up with an alternative protocol that did not need to take advantage of op_return, which was much more ideal. If people are particularly opinionated, I can drop it back down to 40 or increase it to something larger, my main thought is that it should be flexible enough to allow for interesting new schemes to be adopted, I imagine a particular witness program version marking an output as being part of a coinjoin liquidity pool or a more modern stealth addressing scheme and these outputs may need larger witness programs.

lib/primitives/address.js
```diff
@@ -225,17 +226,18 @@ class Address extends bio.Struct {
...
 
-    return this.fromHash(addr.hash, addr.version);
+    return this.fromHash(hash, version);
   }
 
   /**
    * Inject properties from witness.
    * @private
    * @param {Witness} witness
+   * @param {Number} version
```
pinheadmz: Did you mean to add `version` as an argument to `fromWitness()` ? Or is is the JSDoc comment from an outdated commit?

lib/primitives/address.js
```diff
@@ -320,7 +325,9 @@ class Address extends bio.Struct {
    */
 
   fromProgram(version, hash) {
-    assert(version >= 0, 'Bad version for witness program.');
+    assert(version >= consensus.MIN_WITNESS_PROGRAM_VERSION
```
pinheadmz: Interesting that a max version wasn't enforced here already - do you know why? Might need to add an extra check to bcoin...

lib/primitives/address.js
tynes: I reverted some changes that I was playing with and accidentally left this in. The intention was to prepare the codebase for new versions, as methods like `fromPubkeyhash` and `fromPubkey` assume version 0. I thought it would be too large of a change for this PR.

lib/primitives/address.js
tynes: `fromProgram` is not part of the consensus code path in either `hsd` or `bcoin`

test/tx-test.js
```diff
@@ -242,7 +242,7 @@ describe('TX', function() {
 
     let raw = tx.encode();
     assert.strictEqual(encoding.readU64(raw, 46), 0xdeadbeef);
-    raw[54] = 0x7f;
+    raw[52] = 0x7f;
```
pinheadmz: Why did this change?

pinheadmz: Oh nm, just saw the commit msg.

pinheadmz:
Would it be possible to see a diff of bech32.js before and after? Or commit in a way that it can be reviewed? I'm just curious what had to be changed in that file.

pinheadmz:
nm: I gotcha 

https://gist.github.com/pinheadmz/52853f865cdd530fff8421443fe40cab

tynes:
> nm: I gotcha
> https://gist.github.com/pinheadmz/52853f865cdd530fff8421443fe40cab

Thanks for posting this @pinheadmz.

I'm realizing that using a buffer pool (`POOL128`) will not work, as there is no upper bound on the size of the data that can be encoded. All of those checks happen outside of `lib/ui/bech32.js`.

If there is consensus on this PR moving forward, I will fix that problem using `Buffer.allocUnsafe`.

PR #305, opened 2019-11-20
chain: disable goosig after 1 year + 1 month

tynes:
Edited 1/24/20

Adds a new hard coded deployment that disables GooSig after 1 year and 1 month of blocktime. This is to prevent the cryptography being broken in an unforseen way or a social attack where somebody claims it is broken when it actually isn't.

~~Two~~ One new field is added to `lib/protocol/networks`

- ~~`[network].airdrop.disableGoosig`~~
- ~~`[network].airdrop.disableGoosigHeight`~~
- `[network].goosigStop`

~~I used 2 variables to be very explicit, `disableGoosig` is a boolean and `disableGoosigHeight` is an integer. We could combine them into a single variable using `-1` as `disableGoosig: false` if that is preferred.~~

~~The `DeploymentState` is updated with a new property `goobye` and a getter `hasGooBye`. I wasn't sure what the best name for this deployment is, so I just went with that. Open to suggestions.~~

The functionality is as of follows:

If the network allows for GooSig to be disabled and the height of the previous block is greater than or equal to the hardcoded network height, then disable GooSig.

TODO:

- [x] Figure out correct testnet values

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -550,6 +550,24 @@ class Chain extends AsyncEmitter {
             100);
         }
 
+        if (this.height >= this.network.goosigStop) {
+          const key = proof.getKey();
+
+          if (!key) {
+            throw new VerifyError(block,
+              'invalid',
+              'bad-airdrop-format',
+              100);
+          }
+
+          if (key.isGoo()) {
+            throw new VerifyError(block,
+              'invalid',
+              'bad-goosig-disabled',
+              100);
+          }
+        }
+
         // Note: GooSig RSA 1024 is possible to
         // crack as well, but in order to make
         // it safe we would need to include a
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -556,7 +556,7 @@ class Chain extends AsyncEmitter {
           if (!key) {
             throw new VerifyError(block,
               'invalid',
-              'bad-airdrop-format',
+              'bad-airdrop-proof',
               100);
           }
 
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -282,6 +282,23 @@ class Mempool extends EventEmitter {
         this.untrackClaim(entry);
     }
 
+    // Remove all GooSig based airdrops from the mempool
+    // when the block one before the height that disables
+    // GooSig is added to the mempool to prevent the
+    // mining of invalid blocks.
+    if (block.height + 1 === this.network.goosigStop) {
+      for (const [hash, entry] of this.airdrops.entries()) {
+        const airdrop = this.getAirdrop(hash);
+        const key = airdrop.getKey();
+
+        if (!key)
+          continue;
+
+        if (key.isGoo())
+          this.untrackAirdrop(entry);
+      }
+    }
+
     this.cache.sync(block.hash);
 
     await this.cache.flush();
@@ -1232,6 +1249,16 @@ class Mempool extends EventEmitter {
 
     if (!proof.isSane())
       throw new VerifyError(proof, 'invalid', 'bad-aidrop-proof', 100);
+
+    if (this.chain.height + 1 >= this.network.goosigStop) {
+      const key = proof.getKey();
+
+      if (!key)
+        throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 100);
+
+      if (key.isGoo())
+        throw new VerifyError(proof, 'invalid', 'bad-goosig-disabled', 100);
+    }
 
     if (this.chain.db.field.isSpent(proof.position())) {
       throw new VerifyError(proof,
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -1254,10 +1254,10 @@ class Mempool extends EventEmitter {
       const key = proof.getKey();
 
       if (!key)
-        throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 100);
+        throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 0);
 
       if (key.isGoo())
-        throw new VerifyError(proof, 'invalid', 'bad-goosig-disabled', 100);
+        throw new VerifyError(proof, 'invalid', 'bad-goosig-disabled', 0);
     }
 
     if (this.chain.db.field.isSpent(proof.position())) {
--- a/lib/node/spvnode.js
+++ b/lib/node/spvnode.js
@@ -341,6 +341,20 @@ class SPVNode extends Node {
    */
 
   sendAirdrop(proof) {
+    const key = proof.getKey();
+
+    if (!key) {
+      this.emit('error', new Error('Invalid Airdrop.'));
+      return Promise.resolve();
+    }
+
+    if (this.chain.tip.height + 1 >= this.network.goosigStop) {
+      if (key.isGoo()) {
+        this.emit('error', new Error('GooSig disabled.'));
+        return Promise.resolve();
+      }
+    }
+
     return this.broadcast(proof);
   }
 
--- a/lib/protocol/networks.js
+++ b/lib/protocol/networks.js
@@ -367,6 +367,19 @@ main.block = {
 };
 
 /**
+ * Block height at which GooSig claims are
+ * disabled. This limits risk associated
+ * with newly discovered cryptography
+ * attacks or social engineering attacks.
+ *
+ * Estimated to be disabled at 1 year +
+ * 1 month from the start of the network
+ * on mainnet.
+ */
+
+main.goosigStop = (365 + 30) * main.pow.blocksPerDay;
+
+/**
  * For versionbits.
  * @const {Number}
  * @default
@@ -627,6 +640,8 @@ testnet.block = {
   slowHeight: 0
 };
 
+testnet.goosigStop = 20 * testnet.pow.blocksPerDay;
+
 testnet.activationThreshold = 1512;
 
 testnet.minerWindow = 2016;
@@ -774,6 +789,8 @@ regtest.block = {
   maxTipAge: 0xffffffff,
   slowHeight: 0
 };
+
+regtest.goosigStop = Number.MAX_SAFE_INTEGER;
 
 regtest.activationThreshold = 108;
 
@@ -926,6 +943,8 @@ simnet.block = {
   maxTipAge: 0xffffffff,
   slowHeight: 0
 };
+
+simnet.goosigStop = Number.MAX_SAFE_INTEGER;
 
 simnet.activationThreshold = 75;
 
--- a/test/disable-goosig-test.js
+++ b/test/disable-goosig-test.js
@@ -1,0 +1,232 @@
+/**
+ * test/disable-goosig-test.js - Test disabling GooSig
+ * Copyright (c) 2020, The Handshake Developers (MIT License).
+ * https://github.com/handshake-org/hsd
+ */
+
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const fs = require('fs');
+const {resolve} = require('path');
+const random = require('bcrypto/lib/random');
+const FullNode = require('../lib/node/fullnode');
+const SPVNode = require('../lib/node/spvnode');
+const AirdropProof = require('../lib/primitives/airdropproof');
+const BlockTemplate = require('../lib/mining/template');
+const Block = require('../lib/primitives/block');
+const consensus = require('../lib/protocol/consensus');
+const Network = require('../lib/protocol/network');
+const common = require('../lib/blockchain/common');
+const VERIFY_NONE = common.flags.VERIFY_NONE;
+
+const network = Network.get('regtest');
+const PROOF_FILE = resolve(__dirname, 'data', 'airdrop-proof.base64');
+const raw = Buffer.from(fs.readFileSync(PROOF_FILE, 'binary'), 'base64');
+const proof = AirdropProof.decode(raw);
+
+/**
+ * Create a new FullNode with GooSig
+ * disabled at a specific height.
+ * @param {String} type
+ * @returns {FullNode|SPVNode}
+ */
+
+function createNode(type) {
+  if (type === 'spv')
+    return new SPVNode({
+      memory: true,
+      network: 'regtest'
+    });
+
+  return new FullNode({
+    memory: true,
+    network: 'regtest'
+  });
+}
+
+/**
+ * Create a mock block with a
+ * known previous blockhash.
+ * @param {Buffer} prev
+ * @returns {Block}
+ */
+
+function mockBlock(prev) {
+  assert(Buffer.isBuffer(prev));
+  assert(prev.length === 32);
+
+  return new Block({
+    version: 0,
+    prevBlock: prev,
+    merkleRoot: random.randomBytes(32),
+    witnessRoot: random.randomBytes(32),
+    treeRoot: random.randomBytes(32),
+    reservedRoot: random.randomBytes(32),
+    time: 0,
+    bits: network.pow.bits,
+    nonce: 0,
+    extraNonce: random.randomBytes(24),
+    mask: random.randomBytes(32)
+  });
+}
+
+describe('Disable GooSig', function() {
+  let goosigStop;
+
+  before(() => {
+    goosigStop = network.goosigStop;
+    network.goosigStop = 10;
+  });
+
+  after(() => {
+    network.goosigStop = goosigStop;
+  });
+
+  describe('Before Disable', () => {
+    let node;
+
+    before(async () => {
+      node = createNode();
+      await node.open();
+    });
+
+    after(async () => {
+      await node.close();
+    });
+
+    it('should accept GooSig based airdrop in mempool', async () => {
+      assert(node.chain.height < node.network.goosigStop);
+      await node.mempool.addAirdrop(proof);
+      assert.strictEqual(node.mempool.airdrops.size, 1);
+      assert(node.mempool.has(proof.hash()));
+    });
+
+    it('should accept GooSig based airdrop in block', async () => {
+      const block = await node.miner.mineBlock();
+      assert.strictEqual(block.txs[0].inputs.length, 2);
+      assert(await node.chain.add(block));
+    });
+  });
+
+  describe('After Disable', () => {
+    let node;
+
+    before(async () => {
+      node = createNode();
+      await node.open();
+    });
+
+    after(async () => {
+      await node.close();
+    });
+
+    it('should reject GooSig based airdrop in mempool', async () => {
+      while (node.chain.height - 1 < node.network.goosigStop) {
+        const block = await node.miner.mineBlock();
+        assert(await node.chain.add(block));
+      }
+
+      assert.strictEqual(node.chain.height - 1, node.network.goosigStop);
+      await assert.rejects(node.mempool.addAirdrop(proof),
+        {reason: 'bad-goosig-disabled'});
+    });
+
+    it('should reject GooSig based airdrop in block', async () => {
+      const tip = node.chain.tip;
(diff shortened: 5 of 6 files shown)


lib/protocol/networks.js
```diff
@@ -627,6 +653,13 @@ testnet.block = {
   slowHeight: 0
 };
 
+testnet.airdrop = {
+  disableGoosig: true,
+  // TODO: see what this should be relative
+  // to the rest of the network.
+  disableGoosigHeight: (365 - 30) * testnet.pow.blocksPerDay
```
kilpatty: Just taking a look at main `auctionstart=10` vs testnet `auctionstart=.25` the correlation is main*.025 = testnet. 

So I think if we apply the same logic here (365+30)*.025 = ~10 days. That seems a bit too early, so I'd suggest maybe 20 days for this on testnet?

kilpatty:
If I'm understanding this correctly, does this essentially put a time limit on the airdrops claims of 1 year and 1 month? I instinctively feel that this is a bit too short, but definitely open to setting *some time limit on Goosig.

tynes:
> If I'm understanding this correctly, does this essentially put a time limit on the airdrops claims of 1 year and 1 month? I instinctively feel that this is a bit too short, but definitely open to setting *some time limit on Goosig.

Not airdrop claims, just GooSig based claims. People will still be able to claim without using GooSig by using a normal RSA signature after the time period has expired.

lib/protocol/networks.js
tynes: I agree that it does feel early, 20 days sounds good to me. If there was a really easy way for people to do their claims in regtest, that would be most ideal.

kilpatty:
> Not airdrop claims, just GooSig based claims. People will still be able to claim without using GooSig by using a normal RSA signature after the time period has expired.

If Goosig is broken, doesn't this result in the same outcome privacy wise?

Goosig not broken && lasts forever -> Goosig claims are fine.
Goosig broken && lasts forever -> Goosig claims are de-anonymized.
Goosig not broken && removed after 1 year -> Early Airdrop claimers are fine, new ones are de-anonymized.
Goosig broken && removed after 1 year -> Everyone is de-anonymized. 

Does removing Goosig stop the crypto being broken retroactively?

tynes:
> If Goosig is broken, doesn't this result in the same outcome privacy wise?

You also need to consider unforgeability, the event by which somebody without the private key can create a signature that is valid.

tynes:
Updated the mempool code as well to not allow goosig based airdrops into the mempool after the network defined number of blocks has passed. A peer will be banned if they gossip a goosig based airdrop.

kilpatty:
@tynes since regtest essentially allows for an arbitrary number of blocks to be mined at any point, should we consider enabling this on it as well with similar values to testnet?

tynes:
@kilpatty I thought purposefully not disabling GooSig on regtest would be a feature, as people may want to be able to test over time without needing to roll over their chaindb (same with simnet). If there is consensus on also disabling on regtest, I will push up a change for that.

kilpatty:
Good point, and I suppose if someone really wants to test the reverse behavior, they can just adjust the values for their own local regtest network. 

Quick question: What happens in the case of a node that sends Goosig airdrop proofs 1 block away from the cutoff. Those proofs are then not mined in the cutoff block, and so they become invalid. Does that peer get banned?

tynes:
> Quick question: What happens in the case of a node that sends Goosig airdrop proofs 1 block away from the cutoff. Those proofs are then not mined in the cutoff block, and so they become invalid. Does that peer get banned?

Good thinking. The peer would not get banned, but that transaction would be stuck in the mempool. It seems like it would be ideal to evict any goosig based airdrops from the mempool after goosig has been disabled. That way there is no accidental banning of peers.

tynes:
Thinking about how to drop GooSig based airdrops out of the mempool correctly, it feels wasteful to check for the GooSig disabled deployment and then attempt to remove all GooSig airdrops every block that is added to the mempool, since no more GooSig based airdrops will be added after a certain height. Another approach would be to only drop GooSig based airdrops out of the mempool if the block being added is at the height of deployment being activated. This seems like its subject to weird race conditions with reorgs.

My main concern is that somebody will end up with a GooSig based airdrop in the mempool during the activation and will not be able to broadcast their normal RSA based airdrop.

tynes:
I opted for removing all GooSig based airdrops from the mempool at the height that GooSig is disabled instead of checking for the deployment. This saves some extra checks for all future blocks that come after GooSig is disabled.

lib/mempool/mempool.js
```diff
@@ -282,6 +282,27 @@ class Mempool extends EventEmitter {
         this.untrackClaim(entry);
     }
 
+    // Remove all GooSig based airdrops from the mempool
+    // when the block at the height that disables GooSig
+    // is added to the mempool. It would be wasteful to do
+    // this check on every block after GooSig is disabled,
+    // so only do it once on the block height that disables
+    // GooSig.
```
tynes: By directly accessing `this.network.airdrop`, it bypasses the whole deployment state abstraction, which no other parts of the codebase do that. I'm considering removing the checks below on `if (... this.network)` and just using `if (await this.hasGooBye())` instead. This would mean that all future blocks past GooSig being disabled would cause all of the airdrops in the mempool to be checked. While it is wasteful, its less code and more simple. The check could then technically be deleted from the codebase far after GooSig is disabled, or we could add another deployment that is used by the mempool to stop it from checking. It may also be possible to do something like:

```
if (!this.nextState.hasGooBye() && (await this.getNextState()).hasGooBye())
```

This logic would be similar to https://github.com/handshake-org/hsd/pull/305/files#diff-17fec30dbf2ab21611334d7aaa3aec43R668, basically comparing the current state and the next state.

lib/mempool/mempool.js
```diff
@@ -282,6 +282,27 @@ class Mempool extends EventEmitter {
...
+    if (this.network.airdrop.disableGoosig) {
+      if (block.height === this.network.airdrop.disableGoosigHeight) {
+        for (const [hash, entry] of this.airdrops.values()) {
```
pinheadmz: By specifying `.values()` here you will only get the entries back (not the hashes) and the `for` loop will throw as non-iterable

pinheadmz:
Just thinking about the UX of a user sending a goosig airdrop after the expiration time. With a Full Node, it won't go out (`err.score === 100` here)

https://github.com/handshake-org/hsd/blob/c554a6fc15594024b96a4ddc544ff54a0a7cb4f1/lib/node/fullnode.js#L416-L420

But SPV has no such protection:

https://github.com/handshake-org/hsd/blob/c554a6fc15594024b96a4ddc544ff54a0a7cb4f1/lib/node/spvnode.js#L343-L345

In this case, an SPV node _could_ be aware of the rules it is breaking since its just a chain height. I wonder if we should protect such users from being banned this way.

pinheadmz:
I think since this is consensus-critical, it'd be nice to see tests. Could be similar to https://github.com/handshake-org/hsd/pull/330 for the flag-height activation, and similar to https://github.com/handshake-org/hsd/pull/319 for clearing out the mempool.

lib/protocol/networks.js
```diff
@@ -366,6 +366,32 @@ main.block = {
...
+  /**
+   * Allow GooSig to be disabled.
+   */
+
+  disableGoosig: true,
+
+  /**
+   * Block height at which GooSig claims are
+   * disabled. This limits risk associated
+   * with newly discoveredy cryptography
```
pinheadmz: `discoveredy` typo

lib/mempool/mempool.js
pinheadmz: Actually you bring up a good point here. `GooBye` is NOT a deployment -- which is a term that usually refers to soft-forks. It's a hard consensus rule, set before the project launches. It's more akin to `network.auctionStart` and for that reason I think you can pull out all the `hasGoobye()` stuff and just compare to the constant flag-height.

boymanjor: I agree with @pinheadmz. I think we should move all goosig related logic out of the deployment state and reference the network parameters where necessary. However, I like that the network params are split into a boolean representing whether goosig should be disabled and a height specifying when (if necessary).

lib/protocol/networks.js
```diff
@@ -775,6 +806,11 @@ regtest.block = {
   slowHeight: 0
 };
 
+regtest.airdrop = {
+  disableGoosig: false,
+  disableGoosigHeight: 0
```
boymanjor: Maybe we should make this default to `-1`. The codebase typicallys uses `-1` for default, unset heights. Also, a height of `0` implies goosig is always disabled. I understand the boolean parameter should protect against this false interpretation of the height, but it seems safer to set it to `-1`.

lib/protocol/networks.js
tynes: Good catch, fixed.

lib/protocol/networks.js
tynes: Updated to set it as `-1`. If we want to get rid of the boolean and just have `-1` mean never disable, we could do that as well. I wanted to be explicit when writing this.

lib/mempool/mempool.js
tynes: Good catch, there is test coverage for this codepath now.

lib/mempool/mempool.js
tynes: This has been updated with hardcoded constants. I also removed the logline `GooSig has been disabled`, I think that logline would still be nice to have but no blocker.

tynes:
This now has test coverage and is ready for another round of review. Will squash when it looks good.

tynes:
> But SPV has no such protection:

This is a great catch.

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+ * disabled at a specific height.
+ */
+
+function newNode() {
+  const node = new FullNode({
+    memory: true,
+    network: 'regtest'
+  });
+
+  networkAirdrop = node.network.airdrop;
```
boymanjor: This line is not resulting in its intended behavior. Mutating `node.network.airdrop` below will mutate `networkAirdrop`. I think we should remove this line and reset the `network.airdrop` properties within the `after` blocks. I'd also like to suggest renaming this function `createNode` (to follow convention across the test suite) and move it to the top of the file.

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+const BlockTemplate = require('../lib/mining/template');
+const consensus = require('../lib/protocol/consensus');
+
+// Use a valid, working airdrop proof
+const AIRDROP_PROOF_FILE = resolve(__dirname, 'data', 'airdrop-proof.base64');
+const rawAirdropProof =
+  Buffer.from(fs.readFileSync(AIRDROP_PROOF_FILE, 'binary'), 'base64');
+const proof = AirdropProof.decode(rawAirdropProof);
+
+let node, networkAirdrop;
```
boymanjor: nit: because we intend to create a new `node` every internal `describe` block, perhaps we should declare a new node for each `describe` block. Declaring the variable globally implies that we will be using the same instance across the tests, but this PR does a great job of isolating test scenarios.

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+      node = newNode();
+      await node.open();
+    });
+
+    after(async () => {
+      await node.close();
+      node.network.airdrop = networkAirdrop;
+    });
+
+    it('should accept GooSig based airdrop in mempool before', async () => {
```
boymanjor: nit: do we need "before" at the end of this description? The wrapping describe block implies this. This nit applies across all `it` blocks.

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+      await assert.rejects(node.mempool.addAirdrop(proof),
+        {reason: 'bad-airdrop-disabled'});
+    });
+
+    it('should reject GooSig based airdrop in block after', async () => {
+      const tip = node.chain.tip;
+      const version = await node.chain.computeBlockVersion(tip);
+      const mtp = await node.chain.getMedianTime(tip);
+      const time = Math.max(node.network.now(), mtp + 1);
+
```
boymanjor: nit: do we need this newline?

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+    });
+
+    after(async () => {
+      await node.close();
+      node.network.airdrop = networkAirdrop;
+    });
+
+    it('should accept GooSig based airdrop in mempool before', async () => {
+      assert(node.chain.height < node.network.airdrop.disableGoosigHeight);
+
```
boymanjor: nit: do we need this newline?

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+      await node.mempool.addAirdrop(proof);
+      assert.strictEqual(node.mempool.airdrops.size, 1);
+      assert(node.mempool.has(proof.hash()));
+    });
+
+    it('should accept GooSig based airdrop in block before', async () => {
+      const block = await node.miner.mineBlock();
+
```
boymanjor: nit: do we need this newline?

test/disable-goosig-test.js
```diff
@@ -0,0 +1,159 @@
...
+        time: time,
+        bits: target,
+        mtp: mtp
+      });
+
+      template.addAirdrop(proof);
+      template.refresh();
+
+      let block = template.toBlock();
+
```
boymanjor: nit: do we need this newline?

test/disable-goosig-test.js
boymanjor: Zippy's test handles this situation in a clear way: https://github.com/handshake-org/hsd/pull/332/files#diff-0bc7309930b0155fa695605f7786d072R13

lib/mempool/mempool.js
```diff
@@ -1233,6 +1254,17 @@ class Mempool extends EventEmitter {
     if (!proof.isSane())
       throw new VerifyError(proof, 'invalid', 'bad-aidrop-proof', 100);
 
+    const key = proof.getKey();
+
+    if (!key)
+      throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 100);
+
+    if (this.network.airdrop.disableGoosig) {
+      if (this.chain.height >= this.network.airdrop.disableGoosigHeight
```
pinheadmz: I just realized in the mempool all the height checks are `chain.height + 1` -- as in, the mempool checks what is valid in the _next_ block. I gotta fix this in #330 as well

tynes:
Going to update to remove the bool and just use `-1` to mean to never disable.

lib/mempool/mempool.js
tynes: Good looks

tynes:
Squashed into 2 commits

lib/node/spvnode.js
```diff
@@ -341,6 +341,20 @@ class SPVNode extends Node {
...
 
   sendAirdrop(proof) {
+    const key = proof.getKey();
+
+    if (!key) {
+      this.emit('error', new Error('Invalid Airdrop.'));
+      return Promise.resolve();
+    }
+
+    if (this.chain.tip.height >= this.network.goosigStop) {
```
tynes: This should check `this.chain.tip.height + 1`

lib/mempool/mempool.js
```diff
@@ -282,6 +282,25 @@ class Mempool extends EventEmitter {
         this.untrackClaim(entry);
     }
 
+    // Remove all GooSig based airdrops from the mempool
+    // when the block at the height that disables GooSig
+    // is added to the mempool. It would be wasteful to do
+    // this check on every block after GooSig is disabled,
+    // so only do it once on the block height that disables
+    // GooSig.
+    if (block.height === this.network.goosigStop) {
```
tynes: This should be `block.height` + 1

test/disable-goosig-test.js
```diff
@@ -0,0 +1,225 @@
...
+    });
+  });
+});
+
+/**
+ * Create a new FullNode with GooSig
+ * disabled at a specific height.
+ */
+
+function newNode(type = 'full') {
```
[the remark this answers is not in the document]
boymanjor: Also, how do you feel about a name change to `createNode` and moving these helpers to the top of the file? It would keep the style consistent with other test files.

test/disable-goosig-test.js
```diff
@@ -0,0 +1,225 @@
...
+  if (type === 'spv')
+    return new SPVNode({
+      memory: true,
+      network: 'regtest'
+    });
+
+  return null;
+}
+
+function mockBlock(prev) {
```
boymanjor: Do we have a test utility module where this function can live? If not, I think we should just inline since it is only used once.

lib/mempool/mempool.js
boymanjor: We should update the comment as well.

lib/blockchain/chain.js
```diff
@@ -550,6 +550,24 @@ class Chain extends AsyncEmitter {
...
+            throw new VerifyError(block,
+              'invalid',
+              'bad-airdrop-sanity',
+              100);
+          }
+
+          if (key.isGoo()) {
+            throw new VerifyError(block,
+              'invalid',
+              'bad-airdrop-disabled',
```
boymanjor: Should the reason be 'bad-goosig-disabled'?

tynes: Updated to this

test/disable-goosig-test.js
tynes: We spoke out of band about this, but keeping it as a function was done purposely to make the test logic easier to read. Moved the function up to the top of the file.

tynes:
@boymanjor Any more comments on this one? I think its good to be merged.

lib/node/spvnode.js
```diff
@@ -341,6 +341,20 @@ class SPVNode extends Node {
...
 
   sendAirdrop(proof) {
+    const key = proof.getKey();
+
+    if (!key) {
+      this.emit('error', new Error('Invalid Airdrop.'));
+      return Promise.resolve();
+    }
+
+    if (this.chain.tip.height + 1 >= this.network.goosigStop) {
```
tynes: This should be minus one, not plus one.

lib/mempool/mempool.js
```diff
@@ -1233,6 +1250,16 @@ class Mempool extends EventEmitter {
...
       throw new VerifyError(proof, 'invalid', 'bad-aidrop-proof', 100);
 
+    if (this.chain.height + 1 >= this.network.goosigStop) {
+      const key = proof.getKey();
+
+      if (!key)
+        throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 100);
+
+      if (key.isGoo())
+        throw new VerifyError(proof, 'invalid', 'bad-goosig-disabled', 100);
```
chjj: I think @pinheadmz raises a good point. This should probably be a ban score of zero as people might make mistakes, and this particular sanity check doesn't require much processing power.

tynes: Updated for these errors to have a banscore of 0 and added test coverage of those banscores.

lib/blockchain/chain.js
```diff
@@ -550,6 +550,24 @@ class Chain extends AsyncEmitter {
             100);
         }
 
+        if (this.height >= this.network.goosigStop) {
+          const key = proof.getKey();
+
+          if (!key) {
+            throw new VerifyError(block,
+              'invalid',
+              'bad-airdrop-format',
```
boymanjor: nit: should this reason read 'bad-airdrop-proof'?

tynes: I used `bad-airdrop-format` since the same error message is being used earlier in the validation.

https://github.com/handshake-org/hsd/blob/e2039e895484b17a43ac025a87a033aed2fe20ba/lib/blockchain/chain.js#L540-L544

tynes: Updated since `bad-xxx-format` is used when decoding specifically

PR #307, opened 2019-12-12
genesis: regenerate

tynes:
Regenerate the genesis block for each network after the `NAME_ROOT`, `AIRDROP_ROOT` and `FAUCET_ROOT` are committed to in the genesis block.

The `NAME_ROOT` is a merkle root that commits to all of the reserved names.

This was created by running `node scripts/genesis.js` in the root of `handshake-org/hsd`

See commit https://github.com/handshake-org/hsd/commit/a1a2231a60654c6375699b4ce9c2120c25d723bf


tynes:
https://github.com/tynes/hsd/pull/1

PR #308, opened 2019-12-12
Clear simnet identity key for experimental testnet

mslipper:


tynes:
Testnet sets its identity key to `null`, just like what this PR is doing to regtest.

https://github.com/handshake-org/hsd/blob/f47b860ffbb583b5e050ea446b5a629e860349b9/lib/protocol/networks.js#L700

Looks good to me, will merge after CI finishes and passes

PR #309, opened 2019-12-12
experimental testnet network values v0

tynes:
Updating the simnet network values to be more similar to the testnet. Some goals for this were to speed up the bidding process and lower the amount of time before names expire. This will help to more easily test out infrastructure regarding the auction process. It will be more annoying to use the network as names will roll over faster, but that is ok for this experimental testnet.


lib/protocol/networks.js
```diff
@@ -899,23 +897,23 @@ simnet.pow.targetTimespan =
...
-  renewalWindow: 2500,
+  renewalWindow: simnet.pow.blocksPerDay,
   renewalPeriod: 1250,
-  renewalMaturity: 25,
+  renewalMaturity: 1,
   claimPeriod: 75000,
-  biddingPeriod: 25,
-  revealPeriod: 50,
+  biddingPeriod: (simnet.pow.blocksPerDay / 2) | 0,
+  revealPeriod: (simnet.pow.blocksPerDay / 2) | 0,
```
kilpatty: Thoughts about making this a day long rather than 12 hours? Not that the names on this network matter all that much, but it may make for some awkward times needed to reveal a name (e.g. Bid at 10am, renewal is overnight from 10PM to 10am).

lib/protocol/networks.js
```diff
@@ -864,9 +864,7 @@ const simnet = {};
 
 simnet.type = 'simnet';
 
-simnet.seeds = [
-  'aorsxa4ylaacshipyjkfbvzfkh3jhh4yowtoqdt64nzemqtiw2whk@127.0.0.1'
-];
+simnet.seeds = [];
```
tynes: I can add seeder nodes here if anybody wants to start up a node and post their key:ip. You will need to restart and pull the commits that add in the keys.

lib/protocol/networks.js
```diff
@@ -884,7 +882,7 @@ simnet.genesisBlock = genesis.simnetData;
 
 simnet.pow = {};
 simnet.pow.limit = new BN(
-  '7fffff0000000000000000000000000000000000000000000000000000000000',
+  '0000003000000000000000000000000000000000000000000000000000000000',
   'hex'
```
tynes: This is to match the testnet value.

lib/protocol/networks.js
```diff
@@ -899,23 +897,23 @@ simnet.pow.targetTimespan =
   simnet.pow.targetWindow * simnet.pow.targetSpacing;
 simnet.pow.minActual = (simnet.pow.targetTimespan / 4) >>> 0;
 simnet.pow.maxActual = simnet.pow.targetTimespan * 4;
-simnet.pow.targetReset = false;
+simnet.pow.targetReset = true;
 simnet.pow.noRetargeting = false;
```
tynes: I opted into `targetReset` to allow the pow difficulty to reset if there were no blocks produced after a very long period. This is useful since there is a GPU miner out there and there was not always consistent mining on `testnet4-stable`.

lib/protocol/networks.js
```diff
@@ -864,9 +864,7 @@ const simnet = {};
 
 simnet.type = 'simnet';
 
-simnet.seeds = [
-  'aorsxa4ylaacshipyjkfbvzfkh3jhh4yowtoqdt64nzemqtiw2whk@127.0.0.1'
-];
+simnet.seeds = [];
 
 simnet.magic = 0x473bd012;
```
kilpatty: Should we change the magic number? Prevents any weirdness for anyone who may have a leftover simnet directory in their hsd folder. - OR, thinking out loud here, we write a small script to just nuke any current ~./hsd/simnet folder to prevent any collisions.

lib/protocol/networks.js
```diff
@@ -899,23 +897,23 @@ simnet.pow.targetTimespan =
...
   claimPeriod: 75000,
-  biddingPeriod: 25,
-  revealPeriod: 50,
+  biddingPeriod: (simnet.pow.blocksPerDay / 2) | 0,
+  revealPeriod: (simnet.pow.blocksPerDay / 2) | 0,
   treeInterval: 2,
   transferLockup: 5,
   revocationDelay: 25,
-  auctionMaturity: 25 + 50 + 25,
+  auctionMaturity: 25,
```
kilpatty: Does this need to be 25 + (simnet.pow.blocksPerDay / 2) + (simnet.pow.blocksPerDay / 2)?

lib/protocol/networks.js
kilpatty: `al5pxjw6a6xirdzok3unbatjlsrszwx4bgumteoexdmihdsvpdphy@138.68.61.31`

lib/protocol/networks.js
```diff
@@ -899,23 +897,23 @@ simnet.pow.targetTimespan =
...
-  biddingPeriod: 25,
-  revealPeriod: 50,
+  biddingPeriod: simnet.pow.blocksPerDay,
+  revealPeriod: simnet.pow.blocksPerDay,
   treeInterval: 2,
   transferLockup: 5,
   revocationDelay: 25,
-  auctionMaturity: 25 + 50 + 25,
+  auctionMaturity: 2 + 5 + 25,
   noRollout: false,
```
tynes: Following the same pattern here as in regtest.

`auctionMaturity` = `treeInterval` + `trasferLockup` + `revocationDelay`

https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/protocol/networks.js#L763-L766

lib/protocol/networks.js
tynes: I don't think we should change the magic number but I'm open to it if you are strongly opinionated

lib/protocol/networks.js
mslipper: Kyokan seed nodes:

- `apmv724pysb6ifiqsqzrixg23dybnwehinvmzdv6v6n4h7w2da5rq@45.55.111.156`
- `anhaoy6kcptjt25qeas3zgmsrt5twwjlvqqa4pwnt6elsceazsez2@174.138.111.111`

turbomaze: ajd6iidvt7p22eomfkzxtslgsrp2wllef3j4xqemvbknjfquppdmk@44.229.138.206

lib/protocol/networks.js
tynes: We won't have to make any changes for `hnsd` if we don't change the magic

PR #317, opened 2019-12-19
wallet: mitigate watch only wallet bug for getKey

boymanjor:
This PR mitigates a bug which potentially stores an incorrect
account index in the db of watch only wallets. This is only a
patch of a specific API endpoint. It is necessary for recent
work on ledger-app-hns. Specifically, we have implemented
on-device output verification and need to be able to retrieve
a reliable path to verify and skip change outputs.

Future work involves fixing the source of this issue. For more
details see: https://github.com/bcoin-org/bcoin/issues/698.

### DIFF
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -3483,9 +3483,6 @@ class Wallet extends EventEmitter {
     if (account.accountIndex & HD.common.HARDENED)
       account.accountIndex ^= HD.common.HARDENED;
 
-    // Update the path's account index as well.
-    path.account = account.accountIndex;
-
     return account.derivePath(path, this.master);
   }
 
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -3468,6 +3468,24 @@ class Wallet extends EventEmitter {
     if (!account)
       return null;
 
+    // The account index in the db may be wrong.
+    // We must read it from the stored xpub to be
+    // sure of its correctness.
+    //
+    // For more details see:
+    // https://github.com/bcoin-org/bcoin/issues/698.
+    //
+    // TODO(boymanjor): remove index manipulation
+    // once the watch-only wallet bug is fixed.
+    account.accountIndex = account.accountKey.childIndex;
+
+    // Unharden the account index, if necessary.
+    if (account.accountIndex & HD.common.HARDENED)
+      account.accountIndex ^= HD.common.HARDENED;
+
+    // Update the path's account index as well.
+    path.account = account.accountIndex;
+
     return account.derivePath(path, this.master);
   }
 
@@ -3777,6 +3795,9 @@ class Wallet extends EventEmitter {
       //
       // For more details see:
       // https://github.com/bcoin-org/bcoin/issues/698.
+      //
+      // TODO(boymanjor): remove index manipulation
+      // once the watch-only wallet bug is fixed.
       path.account = account.accountKey.childIndex;
 
       // Unharden the account index, if necessary.
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -18,6 +18,7 @@ const {isSignatureEncoding, isKeyEncoding} = require('../lib/script/common');
 const Resource = require('../lib/dns/resource');
 const Address = require('../lib/primitives/address');
 const Output = require('../lib/primitives/output');
+const HD = require('../lib/hd/hd');
 const rules = require('../lib/covenants/rules');
 const {types} = rules;
 const secp256k1 = require('bcrypto/lib/secp256k1');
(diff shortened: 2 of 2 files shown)


lib/wallet/wallet.js
```diff
@@ -3777,6 +3795,9 @@ class Wallet extends EventEmitter {
       //
       // For more details see:
       // https://github.com/bcoin-org/bcoin/issues/698.
+      //
```
boymanjor: This note demarcates the other location we've implemented this patch.

lib/wallet/wallet.js
```diff
@@ -3468,6 +3468,24 @@ class Wallet extends EventEmitter {
...
+    // TODO(boymanjor): remove index manipulation
+    // once the watch-only wallet bug is fixed.
+    account.accountIndex = account.accountKey.childIndex;
+
+    // Unharden the account index, if necessary.
+    if (account.accountIndex & HD.common.HARDENED)
+      account.accountIndex ^= HD.common.HARDENED;
+
+    // Update the path's account index as well.
+    path.account = account.accountIndex;
```
pinheadmz: Is this necessary? It looks like `Account.derivePath()` only reads `branch` and `index` from the `path` argument (in this case `index` is the final bip32 field, not the account index).

https://github.com/handshake-org/hsd/blob/fc4623d8750399ff086e6de8c023e6dd8c5aa39d/lib/wallet/account.js#L388-L392

and then `Account.deriveKey()` doesn't need anything else from `path`...

pinheadmz:
Tested this out by generating an xpub like so:

mnemonic: `suggest make thumb suggest name approve rally tornado cash boy old couch`
path: `m/44'/5355'/5'` (hsd regtest cointype, child index 5)
account key: 
```
rpubKBAuNsciNsgGoLXLqHegYLDMERvPDouCGkctrSpF41pAQud2prmT6tRUNEtZEjYVsfWHonrZnJ9QxEfkoRBd3BMRKiVPkwv4xokQYKpZwo8j
```

...and importing the key into a new watchonly wallet.
After importing the wallet, testing the output of receive address zero against master and the PR using the CLI:

#### `path-path`

```
 hsw-cli --id=aa key rs1qf93zdnurg5p5ls3gf2yryee68wfd6hsxcjcysm
{
  "name": "default",
  "account": 5,
  "branch": 0,
  "index": 0,
  "publicKey": "022918377b57fb440e4bfa2473facfd053b96b2478708a45cca7c9942ce84823cf",
  "script": null,
  "address": "rs1qf93zdnurg5p5ls3gf2yryee68wfd6hsxcjcysm"
}
```

#### `master`

```
hsw-cli --id=aa key rs1qf93zdnurg5p5ls3gf2yryee68wfd6hsxcjcysm
{
  "name": "default",
  "account": 0,
  "branch": 0,
  "index": 0,
  "publicKey": "022918377b57fb440e4bfa2473facfd053b96b2478708a45cca7c9942ce84823cf",
  "script": null,
  "address": "rs1qf93zdnurg5p5ls3gf2yryee68wfd6hsxcjcysm"
}
```

So that's cool - but I wonder if this should be fixed for other API use cases as well, for example getting wallet history, a TX to this address does _not_ show the child index in the path (on either branch). Although the path indicated here has [other issues](https://github.com/bcoin-org/bcoin/issues/484).

```
    "outputs": [
      {
        "value": 2000000000,
        "address": "rs1qf93zdnurg5p5ls3gf2yryee68wfd6hsxcjcysm",
        "covenant": {
          "type": 0,
          "action": "NONE",
          "items": []
        },
        "path": {
          "name": "default",
          "account": 0,
          "change": false,
          "derivation": "m/0'/0/0"
        }
      }
    ],
```

lib/wallet/wallet.js
boymanjor: We discussed this comment out of band. TL;DR - it is necessary because the wallet uses its current `accountDepth` for the account index of newly created accounts. This behavior breaks if an xpub is imported, which encodes an account index different from the wallet's current `accountDepth`.

boymanjor:
@pinheadmz thanks for the review! I will add a similar test to the test suite.

> So that's cool - but I wonder if this should be fixed for other API use cases as well, for example getting wallet history, a TX to this address does not show the child index in the path (on either branch).

I totally agree. I chose to patch this instead of administering a general fix because we are trying to keep changes to a minimum pre mainnet launch. I have an idea of how to fix this across the board. The approach involves separating the account depth field from the account index field in the walletdb & txdb. I started an exploratory branch with TODOs here: https://github.com/handshake-org/hsd/compare/watchonly

lib/wallet/wallet.js
pinheadmz: I understand that but why update `path.account`? I don't see where that value is ever checked and if I comment out that line, my ad hoc test still passes, as does the one you added in 5611feb18687b5b047621804b51554c19eff6e0e

test/wallet-http-test.js
```diff
@@ -86,6 +88,20 @@ describe('Wallet HTTP', function() {
...
+      watchOnly: true,
+      accountKey: accountInfo.accountKey
+    });
+    const key = await wallet3.getKey(accountInfo.receiveAddress);
+    const xpub = HD.HDPublicKey.fromBase58(accountInfo.accountKey, 'regtest');
+    assert.equal(accountInfo.accountIndex, key.account);
+    assert.equal(xpub.childIndex ^ HD.common.HARDENED, key.account);
+    assert.equal(0, key.branch);
+    assert.equal(0, key.index);
+  });
```
pinheadmz: I feel like this test should explicitly assert that `account.accountIndex != key.childIndex` since that's what we're really trying to solve. Since you're using an account generated by a previous test, regression is otherwise possible. You can use `wallet.getAccountIndex()`

boymanjor: We discussed out of band and decided to derive an account key and address within the test. This allows the reader to be sure of the xpub's account index.

lib/wallet/wallet.js
boymanjor: Ah, I see. I've removed the unnecessary update.

Issue #315, opened 2019-12-19
Node ID keys are malleable: preventing connections possible


pinheadmz:
h/t @kilpatty for discussion.

### Background

1. hsd nodes have public, static ID keys
2. These keys, along with a node's IP and port, are gossiped between peers in `addr` packets `key@IP:port` (`fullname`)
3. In `HostList` and other `net` modules, peers are only identified by their `IP:port` (`hostname`)
4. When a node has received a `fullname` for a potential peer, it will create a record indexed by the `hostname`, and never update it.

### Issues

1. The `key` provided in the `fullname` could be malleated by an attacker. An attacker could send `addr` packets to the entire network with incorrect `key`'s, and if those messages get to a victim before honest `addr` packets, that victim will never be able to establish outbound connections to the nodes in the attacker's packet.

2. A long-running honest full node operator my update their ID key for some reason. Even though they broadcast their own `key` in `addr` packets they send to their own outbound peers, the issue persists. The old deprecated key will remain in `HostList`'s all over the network, and all inbound connections to this honest full node will fail.

#### 1. Key by `fullname` (potential eclipse attack)

The simplest thing to do here would be just modify anything like `this.map.get(hostname)` in `HostList` and related modules so that a `fullname` is used instead. This means that there will be separate entries for each combination of `key` and `hostname`. 

So these entries would be saved as two separate nodes:
`aaaaa@10.20.30.40:5000`
`bbbbb@10.20.30.40:5000`

If one entry fails, the other might still succeed. They do not cancel each other out and new keys do not replace old keys. Old keys persist, but if they are bad keys, we just don't connect. The behavior would be the same as if we just had a bad IP address.

Drawback: An attacker could spin up one single full node and just modify it to accept connections with a huge battery of keys. An eclipse attack is then possible if a node thinks it's connecting to 8 different entities when really its all the same IP address and port, but different ID keys.

#### 2. Store ID keys in a separate data structure

We could perhaps add a vector to each potential host that stores more than one key. This would be a change to `NetAddress` and might look like this:

```
NetAddress {
  // Store array of keys we have learned about for this address
  this.keys = ['aaaaa', 'bbbbb', ...];

  // Only set this value once a connection has been established.
  // If a connection fails using this key, set it to null and try
  // the next key from the array.
  this.key = 'ccccc';
}
```

#### 3. Switch noise protocol

Handshake currently uses the `XK` variant of noise:

```
XK:
  <- s
  ...
  -> e, es
[...]

PR #318, opened 2019-12-20
Use fullname (key@IP:port) for keys in HostList map

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/315

This essentially implements Solution 1 as described in the issue. There still may be other attack vectors open, but this should mitigate the black-hole effects of gossiping bad ID keys.

We switch from indexing by `hostname` (`IP:port`) to `fullname` (`key@IP:port`) in some data structures:

#### `HostList`

This is the manager of all available hosts. Internally it has several maps and lists. This PR changes the main map functions `has`, `get` and `set` to index by `fullname` instead of `hostname`. This also required changing `HostEntry.key()` to return `fullname` instead of `hostname`. 

Banned peers are still indexed by IP _only_ -- new ID keys or ports will not allow an attacker to circumvent a ban.

#### `Peer.addrFilter`

This is a Bloom Filter used when sending and receiving `addr` messages to ensure that we don't send a peer an address it already knows about (they sent it to us, or we've already sent it to them in a previous message).

This change means that we will broadcast new ID keys for an address even if the IP and port for that address has already been sent or received. 

#### `PeerList` (`Pool.peers`)

Ok, so - this object DOES NOT change. It will continue to index peers by `hostname`. This will prevent us from connecting more than once to the same `hostname` (`IP:port`), even if we have multiple ID keys for that host.

https://github.com/handshake-org/hsd/blob/c554a6fc15594024b96a4ddc544ff54a0a7cb4f1/lib/net/pool.js#L3265-L3271

#### Other

- A few rpc calls and inspect functions have been modified to print out `fullname` instead of `hostname`.

- `NonceList` (`Pool.nonces`) will remain unchanged - these are the nonces we use in `version` packets to prevent self-connection. Indexing by `hostname` should be fine.

- `TimeData.known` (`Network.time`) will remain unchanged - this is a map of clock-time offsets we use to stay in sync with peers. We don't need ID keys for this since we won't be connecting to more than one peer at the same `IP:port`.

- `HostList.used` and `HostList.fresh` remain unchanged - these are arrays of maps/lists that serve as the "buckets" of hosts. The `used` bucket for each host is still determined ONLY by it's IP address (the murmur3 hash of the IP address or `NetAddress.raw`). The `fresh` bucket also salts that hash with the IP of the peer that _sent us the address_. However, within a single bucket, addresses are indexed by `fullname` (or `HostEntry.key()`). This means that two keys for the same IP address will end up in the same bucket as two distinct hosts.

### DIFF
--- a/lib/net/hostlist.js
+++ b/lib/net/hostlist.js
@@ -1361,7 +1361,7 @@ class HostEntry {
    */
 
   key() {
-    return this.addr.hostname;
+    return this.addr.fullname();
   }
 
   /**
--- a/lib/net/hostlist.js
+++ b/lib/net/hostlist.js
@@ -911,7 +911,7 @@ class HostList {
    */
 
   pushLocal(addr, score) {
-    if (!addr.isRoutable())
+    if (!addr.isRoutable() && this.network.type !== 'regtest')
       return false;
 
     if (this.local.has(addr.hostname))
--- a/lib/net/hostlist.js
+++ b/lib/net/hostlist.js
@@ -454,7 +454,7 @@ class HostList {
   add(addr, src) {
     assert(addr.port !== 0);
 
-    let entry = this.map.get(addr.hostname);
+    let entry = this.map.get(addr.fullname());
 
     if (entry) {
       let penalty = 2 * 60 * 60;
@@ -610,12 +610,12 @@ class HostList {
 
   /**
    * Remove host from host list.
-   * @param {String} hostname
+   * @param {String} fullname
    * @returns {NetAddress}
    */
 
-  remove(hostname) {
-    const entry = this.map.get(hostname);
+  remove(fullname) {
+    const entry = this.map.get(fullname);
 
     if (!entry)
       return null;
@@ -655,11 +655,11 @@ class HostList {
 
   /**
    * Mark host as failed.
-   * @param {String} hostname
+   * @param {String} fullname
    */
 
-  markAttempt(hostname) {
-    const entry = this.map.get(hostname);
+  markAttempt(fullname) {
+    const entry = this.map.get(fullname);
     const now = this.network.now();
 
     if (!entry)
@@ -671,11 +671,11 @@ class HostList {
 
   /**
    * Mark host as successfully connected.
-   * @param {String} hostname
+   * @param {String} fullname
    */
 
-  markSuccess(hostname) {
-    const entry = this.map.get(hostname);
+  markSuccess(fullname) {
+    const entry = this.map.get(fullname);
     const now = this.network.now();
 
     if (!entry)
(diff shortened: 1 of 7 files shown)


pinheadmz:
Marking this ready for review to get the ball rolling on concept ACK's. I think we still need unit tests for `Pool` to check how `HostList` and `PeerList` are used together.

pinheadmz:
New tests added to cover `Pool`, specifically to verify that even though we may ingest multiple keys for a single IP:port, we will not connect twice to the same IP:port

@kilpatty

kilpatty:
A couple of thoughts: 

I think this is probably one of the better solutions to the key@id issue we are seeing, without massively rewriting brontide. Although I have a few hesitations:

It makes eclipse attackers significantly easier to execute and less expensive. Previously, the cost of executing a successful eclipse attack was the amount of time needed to monopolize the hostlist of a node times the cost of renting the number of IPs needed to do that. With this change, the number of IPs needed to successfully eclipse a node is cut by a large majority. 

The eclipse attack can essentially be broken down into a few pieces. 

A. Start connecting to a victim with "attack node IPs" - Addresses that connect as an inbound are put into the "tried" bucket in the hostlist.

B. Fill victim's new bucket with "trash" IPs. When a node restarts, it goes through a process of selecting new addresses from either tried or new. By filling new with unconnectable trash addresses, it increases the changes of picking from "tried" (where the attack addresses live). 

C. Wait for the victim to restart and only be using the Attack addresses. 

With this change, we could execute the eclipse attack much more efficiently. 

A. Start connecting to the node with "attack node keys" - This could be a select of key/ip pairs that are owned by the attacker. Because key/IP pairs are stored in the same bucket, to ensure maximum success, one would only need as many IPs as the number of tried buckets (~30? right now in hsd, ~64? in Bitcoind). 

B. Instead of filling the the new buckets with strictly trash, fill them with a combination of attack key/ip pairs and trash. The more valid attack key/ip pairs inserted into new, the higher the probability of success, and since generating a new key is trivial this is a cost-free increase of success. 

C. Wait for victim to restart 

To execute a successful eclipse attack now, someone could manage to do it with roughly 30-40 IPs. This could be quite trivial to accomplish using the cloud hosting platforms - whereas without storing key@ip the amount of IPs needed was in the thousands - something likely not accessible to the average attacker. 

Even after implementing some of the mitigations that Bitcoin did from the eclipse paper: https://www.usenix.org/system/files/conference/usenixsecurity15/sec15-paper-heilman.pdf
I believe that Eclipse attacks would still be trivially easy using this attack vector. 

Even further, I could see an attack where every single node only connects to a single node. If I'm an attacker, I could connect to the network and start sending 1000s of Addr packets with key/ip pairs that I accept. While a single node will only accept those packets and put them into a single bucket, if other nodes also gossip those addrs then they could be placed into different buckets bc the bucket selection process takes the originating IP into account. Therefore, it's possible that after being gossiped thoroughly by the network (or by a select group of attack nodes), that nodes have their entire hostlist filled with a single IP but multiple keys. 

It may be useful for us to test out this kind of attack with this patch on a real simnet and see if it feasible.

pinheadmz:
This will be mitigated by making brontide optional and not gossiping node public keys

Issue #306, opened 2019-12-01
TXs do not re-enter mempool when a block is disconnected


pinheadmz:
This came to my attention when a user opened https://github.com/bcoin-org/bcoin/issues/915, as an amazing coincidence I was also studying the [Bitcoin Core BIP68 test](https://github.com/bitcoin/bitcoin/blob/master/test/functional/feature_bip68_sequence.py) which shed some important light on the subject.

The issue arises when a block is disconnected (either by a reorg, or locally with `rpc invalidateblock`) and there are two relevant places in the code.

First: when a block is removed, if the mempool is _empty_, all the TXs in the block are discarded:

https://github.com/handshake-org/hsd/blob/27c99adda0ff7f91d3ccfa0af5806cc4871d1b00/lib/mempool/mempool.js#L333-L339

Second, if a `'reorganize'` event is caught by mempool, the mempool is "sterilized" by [`_handleReorg()`](https://github.com/handshake-org/hsd/blob/27c99adda0ff7f91d3ccfa0af5806cc4871d1b00/lib/mempool/mempool.js#L408) but this function is a little brutal, it removes all CSV transactions and all coinbase-spends.

The effect of these two snippets is to simplify mempool processing after a reorg (the BIP68 rules get stupid-complicated when transactions are re-entering the mempool, especially if the _parent of a CSV TX_  re-enters the mempool)

I don't think this is a much an issue with bcoin but it will be with Handshake, since this is the code all miners will be running, and there are more likely to be re-orgs in the early stages of the chain. The effect will be lots of transactions disappearing completely from mempools and blocks if a block is removed, the sender (or recipient) will have to re-broadcast. It is also very non-optimal for miners because they are throwing away fees.

Using the Bitcoin Core test as a guide, I'm going to try to implement deeper reorg-mempool behavior behavior in bcoin and after review and merge, port that back to hsd.

PR #319, opened 2019-12-22
Improve re-insertion of TXs from disconnected blocks into mempool

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/306

Port of https://github.com/bcoin-org/bcoin/pull/917, although it hasn't been reviewed or merged yet over there. I think this is a bigger concern for Handshake since this codebase will represent 100% of the network, and chain reorgs are more likely on hsd in the early days.

The contextual checks for TXs with covenants re-entering mempool after a block disconnect has already been written in `_handleReorg()`, so for that case I just added a test.

The other cases ported over from bcoin are premature coinbase spends, and premature BIP68 sequence lock spends.

With these contextual checks in place, the "don't remove block if mempool empty" check can be safely removed.

### DIFF
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -416,6 +416,8 @@ class Mempool extends EventEmitter {
         continue;
       }
 
+      const view = await this.getCoinView(tx);
+
       let hasLocks = false;
 
       for (const {sequence} of tx.inputs) {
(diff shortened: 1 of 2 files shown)


pinheadmz:
Also thinking we can deal with Claims and AirdropProofs here too:

https://github.com/handshake-org/hsd/blob/c554a6fc15594024b96a4ddc544ff54a0a7cb4f1/lib/mempool/mempool.js#L414-L416

They can be re-evaluated on reorgs. `Claim`s for example commit to a block hash so that needs to be re-checked.

pinheadmz:
Ok this latest commit adds the Claims and AirDrops back into the mempool after a reorg. This will require diligent review if we want to keep it. From what I can tell, there are no restrictions on AirDrops if the chain is rewound (unlike premature coinbase spends). The only restriction on Claims would be the inception time (`RRSIG` contains a timestamp which is compared to the chain tip) so I added that exception.

pinheadmz:
Forgot to include checks for block commitment of Claims re-entering mempool in last commit, that has been updated. Everything should be covered now.

pinheadmz:
Bonus TODO: this is getting pretty advanced, but Bitcoin Core will even hold on to the evicted TXs until after a reorg is complete, and then see if any are still ok (for example, a coinbase spend might still be valid after a reorg, just not _during_):


https://github.com/bitcoin/bitcoin/blob/631df3ee87ec93e1fc748715671cdb5cff7308e6/src/txmempool.h#L766-L776

chjj:
After looking at this for a while, I can't find anything wrong with it. Merging for now.

PR #320, opened 2019-12-29
Cleaning up Dockerfile with Best Practices

afmsavage:
Removed the `apt update` command as the best practice way to use a 3rd party Docker image as base layer is to keep the package versions that come with it.  Also might consider pinning the versions of each package added in line 14 to known working versions in the case there's any breaking changes (unlikely with those specific packages so I left it alone.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,13 @@
 FROM node:alpine AS base
 RUN mkdir -p /code
 WORKDIR /code
-CMD "hsd"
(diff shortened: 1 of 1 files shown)


tynes:
There is another PR that updates the `Dockerfile` https://github.com/handshake-org/hsd/pull/288

Good tip regarding updating the package versions

afmsavage:
@tynes perfect, I'll close this PR and add some comments over there.

Issue #312, opened 2019-12-17
RPC getblock confirmations inaccurate


tynes:
The node RPC `getblock` should return `confirmations: -1` for any blocks that are not in the main chain. This is helpful for applications that need to test if blocks are in the main chain or if they are actually an orphan. `bitcoind` has this functionality, but it was not implemented in `bcoin`.

There is currently a PR open in `bcoin`, this needs to be ported to `hsd` eventually.

See:
https://github.com/bcoin-org/bcoin/issues/918
https://github.com/bcoin-org/bcoin/pull/919

PR #321, opened 2020-01-01
node/rpc: getblock confirmation fixes

tynes:
Update the node RPC methods `getblock` and `getblockheader` for returning the `confirmations` as `-1` if the block is not in the main chain. There is no way to tell if a block is not in the main chain from the API without these changes.

Closes https://github.com/handshake-org/hsd/issues/312

Port of:
- https://github.com/bcoin-org/bcoin/commit/b9ad073b9bca527b0e6faca8a9f1781d7c68d175
- https://github.com/bcoin-org/bcoin/commit/54b31293ced0f3652651156ea5b285619371ef17
- Additional tests in `test/node-rpc-test` for `signmessagewithprivkey` and `verifymessage`

These changes have already been merged into `bcoin`

### DIFF
--- a/test/node-rpc-test.js
+++ b/test/node-rpc-test.js
@@ -1,0 +1,134 @@
+/**
+ * test/node-rpc-test.js - Node RPC tests for hsd
+ * Copyright (c) 2020, The Handshake Developers (MIT Licence)
+ * https://github.com/handshake-org/hsd
+ */
+
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const FullNode = require('../lib/node/fullnode');
+const {NodeClient} = require('hs-client');
+
+const ports = {
+  p2p: 49331,
+  node: 49332,
+  wallet: 49333
+};
+const node = new FullNode({
+  network: 'regtest',
+  apiKey: 'foo',
+  walletAuth: true,
+  memory: true,
+  workers: true,
+  workersSize: 2,
+  port: ports.p2p,
+  httpPort: ports.node
+});
+
+const nclient = new NodeClient({
+  port: ports.node,
+  apiKey: 'foo',
+  timeout: 15000
+});
+
+describe('RPC', function() {
+  this.timeout(15000);
+
+  before(async () => {
+    await node.open();
+    await nclient.open();
+  });
+
+  after(async () => {
+    await nclient.close();
+    await node.close();
+  });
+
+  describe('getblock', function () {
+    it('should rpc getblock', async () => {
+      const {chain} = await nclient.getInfo();
+      const info = await nclient.execute('getblock', [chain.tip]);
+
+      const properties = [
+        'hash', 'confirmations', 'strippedsize',
+        'size', 'weight', 'height', 'version',
+        'versionHex', 'merkleroot', 'witnessroot',
+        'treeroot', 'reservedroot', 'mask',
+        'coinbase', 'tx', 'time', 'mediantime',
+        'nonce', 'bits', 'difficulty', 'chainwork',
+        'nTx', 'previousblockhash', 'nextblockhash'
+      ];
+
+      for (const property of properties)
+        assert(property in info);
+
+      assert.deepEqual(chain.height, info.height);
+      assert.deepEqual(chain.tip, info.hash);
+      assert.deepEqual(chain.treeRoot, info.treeroot);
+    });
+
+    it('should return correct height', async () => {
+      const address = 'rs1qjjpnmnrzfvxgqlqf5j48j50jmq9pyqjz0a7ytz';
+
+      // Mine two blocks.
+      await nclient.execute('generatetoaddress', [2, address]);
+
+      const {chain} = await nclient.getInfo();
+      const info = await nclient.execute('getblock', [chain.tip]);
+
+      // Assert the heights match.
+      assert.deepEqual(chain.height, info.height);
+    });
+
+    it('should return confirmations (main chain)', async () => {
+      const {chain} = await nclient.getInfo();
+
+      const {genesis} = node.network;
+      const hash = genesis.hash.toString('hex');
+
+      const info = await nclient.execute('getblock', [hash]);
+      assert.deepEqual(chain.height, info.confirmations - 1);
+    });
+
+    it('should return confirmations (post reorg)', async () => {
+      // Get the current chain state
+      const {chain} = await nclient.getInfo();
+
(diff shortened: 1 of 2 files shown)


test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+  it('should rpc help', async () => {
+    assert(await nclient.execute('help', []));
+    await assert.rejects(async () => {
+      await nclient.execute('help', ['getinfo']);
+    }, {
+      name: 'Error',
+      message: /^getinfo/
+    });
+  });
+  it('should rpc getinfo', async () => {
```
pinheadmz: nit: newline

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+      const info = await nclient.execute('getblock', [chain.tip]);
+
+      const properties = [
+        'hash', 'confirmations', 'strippedsize',
+        'size', 'weight', 'height', 'version',
+        'versionHex', 'merkleroot', 'coinbase',
+        'tx', 'time', 'mediantime', 'nonce',
+        'bits', 'difficulty', 'chainwork',
+        'nTx', 'previousblockhash', 'nextblockhash'
+      ];
```
pinheadmz: could also include some of the hsd-exclusive stuff `witnessRoot`, `treeRoot`, etc, and compare against the `chain.tip` entry in the assertion tests below

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+
+      const {genesis} = node.network;
+      const hash = genesis.hash.toString('hex');
+
+      const info = await nclient.execute('getblock', [hash]);
+
+      assert.deepEqual(chain.height, info.confirmations - 1);
+    });
+
+    it('should return confirmations (orphan)', async () => {
```
pinheadmz: Technically it is not an orphan block since it has a parent. I think `stale` or `alternate` is the better term.

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+      // the genesis block.
+      const {genesis} = node.network;
+      let entry = await node.chain.getEntry(genesis.hash);
+
+      // Reorg from the genesis block.
+      for (let i = 0; i < chain.height + 1; i++) {
+        const block = await node.miner.mineBlock(entry);
+        await node.chain.add(block);
+        entry = await node.chain.getEntry(block.hash());
+      }
```
pinheadmz: I know this is how the test was merged into bcoin, but you don't need a reorg for this: I think you can just mine a single alternate block off of any block in the current chain that isn't the tip. It will be saved as an alternate and should not have a main chain height.

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+    before(async () => {
+      const addr = await wclient.execute('getnewaddress');
+      privKeyWIF = await wclient.execute('dumpprivkey', [addr]);
+      ring = KeyRing.fromSecret(privKeyWIF, 'regtest');
+    });
+
+    it('should sign message', async () => {
+      const msg = Buffer.from(MAGIC_STRING + message, 'utf8');
+      const digest = blake2b.digest(msg);
+      const expectedSignature = ring.sign(digest);
```
pinheadmz: I feel like hard coded values for key WIF and expected sig would be more robust tests against future regression, especially since the rpc call uses `KeyRing.fromSecret(raw, network)`, this test is comparing the same output from `KeyRing.sign()`.

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+      const result = await nclient.execute('verifymessage', [
+        address,
+        base64.encode(signature),
+        message
+      ]);
+
+      assert.equal(result, true);
+    });
+
+    it('should verify invalid signature', async () => {
```
pinheadmz: nit: should "reject" invalid signature

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+  wallet: 49333
+};
+const node = new FullNode({
+  network: 'regtest',
+  apiKey: 'foo',
+  walletAuth: true,
+  memory: true,
+  workers: true,
+  workersSize: 2,
+  plugins: [require('../lib/wallet/plugin')],
```
pinheadmz: Including a wallet here seems excessive to me, since it's just used to derive a coinbase address, which could be hard-coded and reduce test complexity, but it doesn't bother me that much.

pinheadmz:
Besides some nits in the test, this looks good to me. I compared against bitcoin core `getblock` and `getblockheader` to make sure that confirmations weren't off by one (for example, the chain tip has "1 confirmation").

Got expected results from SPV mode as well: `getblock` returns "block not found" but `getblockheader` works fine.

test/node-rpc-test.js
```diff
@@ -0,0 +1,230 @@
...
+/* eslint prefer-arrow-callback: "off" */
+'use strict';
+
+const assert = require('bsert');
+const FullNode = require('../lib/node/fullnode');
+const {NodeClient, WalletClient} = require('hs-client');
+const KeyRing = require('../lib/primitives/keyring');
+const pkg = require('../lib/pkg');
+
+// will this work?
```
tynes: Delete this dead code

test/node-rpc-test.js
tynes: Good point. I would like to add in a call to `getblock` with the tip before it is reorged out to show that the same block shows confirmations and then is -1 after the reorg. What you are describing could be another test case.

test/node-rpc-test.js
tynes: This is a good point, will push up a fix.

tynes:
@pinheadmz Thank you for the review, pushed up fixes to address your comments.

test/node-rpc-test.js
```diff
@@ -0,0 +1,236 @@
...
+      const block = await node.miner.mineBlock(entry);
+      assert(await node.chain.add(block));
+
+      const hash = block.hash().toString('hex');
+      const info = await nclient.execute('getblock', [hash]);
+      assert.deepEqual(info.confirmations, -1);
+    });
+  });
+
+  describe('signmessagewithprivkey', function () {
```
boymanjor: @tynes can you separate these sign message tests into a separate PR? They are off topic here.

tynes:
@boymanjor I removed the tests unrelated to the RPC `getblock`, will add them back in an additional PR after this is merged since they depend on a file added by this PR (`node-rpc-test.js`) existing.

PR #324, opened 2020-01-10
protocol/networks: update simnet parameters to prevent premature expiry

turbomaze:
The `renewalWindow` must be `> biddingPeriod + revealPeriod + treeInterval` because otherwise, the name will expire before it is possible to send a register/redeem.


pinheadmz:
LGTM, untested ACK

PR #326, opened 2020-01-10
test: abstract some mining heights

tynes:
It is much more useful to mine based
on the auction lengths instead of using
magic numbers. This allows the parameters
to be tweaked and then ran through the
test suite because it is possible to choose
parameters that do not allow the auctions
to work at all. The `Network.get` call
at the top of the file can be altered
to select a different network, such as
simnet.

### DIFF
--- a/test/auction-test.js
+++ b/test/auction-test.js
@@ -129,8 +129,8 @@ describe('Auction', function() {
       assert(await chain.add(block));
     });
 
-    it('should mine 10 blocks', async () => {
-      for (let i = 0; i < 10; i++) {
+    it('should mine through the bidding period', async () => {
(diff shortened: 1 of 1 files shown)


kilpatty:
LGTM!

turbomaze:
Good idea mining based on auction stage lengths

tynes:
Note that there are other places in this file that still uses magic numbers, but this was the minimum changes needed to verify that our new simnet params work here https://github.com/handshake-org/hsd/pull/324

pinheadmz:
Concept ACK but I think we should cover all the tests before merge:
`auction-reorg-test.js` I think may be the only other test that needs updates.

Seems to have this in place already:
`wallet-http-test.js` 
`auction-rpc-test.js`

PR #327, opened 2020-01-10
node/rpc: fix logger statement

tynes:
Previously this log statement would log a `Buffer`, but it should actually log a `string`. ~~This pull request switches the `block.hash` method to using the `block.toHex` method.~~ This pull request uses the correct string formatting symbol in the logger statement so that a string is logged instead of a stringified `Buffer`.

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2608,7 +2608,7 @@ class RPC extends RPCBase {
       entry = await this.chain._add(block);
     } catch (err) {
       if (err.type === 'VerifyError') {
-        this.logger.warning('RPC block rejected: %s (%s).',
+        this.logger.warning('RPC block rejected: %x (%s).',
           block.hash(), err.reason);
         return `rejected: ${err.reason}`;
       }


lib/node/rpc.js
```diff
@@ -2609,7 +2609,7 @@ class RPC extends RPCBase {
     } catch (err) {
       if (err.type === 'VerifyError') {
         this.logger.warning('RPC block rejected: %s (%s).',
-          block.hash(), err.reason);
+          block.toHex(), err.reason);
```
pinheadmz: I think the convention around the code is to use a `%x` placeholder for hash->string. That would make this port to bcoin as well (where block does not extend bufio)

lib/node/rpc.js
```diff
@@ -2608,8 +2608,8 @@ class RPC extends RPCBase {
       entry = await this.chain._add(block);
     } catch (err) {
       if (err.type === 'VerifyError') {
-        this.logger.warning('RPC block rejected: %s (%s).',
-          block.hash(), err.reason);
+        this.logger.warning('RPC block rejected: %x (%s).',
+          block(), err.reason);
```
tynes: brb gotta push a fix for this quick, `hash` got deleted accidentaly

Issue #159, opened 2019-04-29
Miner `createBlock` with Arbitrary Tip Causes Bad Template


tynes:
If the `miner.createBlock` function is called without the `tip` argument, it will mine on top of the chain's tip. In the case where a `tip` is passed in and its a `ChainEntry` that was created during a different Tree Root period (an epoch), it will template the block incorrectly because it uses the current tree root, `this.chain.db.treeRoot()` (scroll down in the snippet)

https://github.com/handshake-org/hsd/blob/bd1ba3dfca841655ada5ab96e70619112721c049/lib/mining/miner.js#L121-L138

I would expect the miner to roll back the state appropriately so that the correct `treeRoot` is calculated correctly irrespective of the height of the `ChainEntry` passed in or just use the previous block header's `treeRoot` field if its not a block height that starts the next epoch (which needs to calculate the epoch's root).

Its still possible to create reorgs but it would require the miner's chain to be at its local tip which is at a lesser height than the global tip. The function `miner.createBlock` can accept an arbitrary tip to mine on top of, which is why I think this function needs to be updated.

You can argue that its not worth time working on code that makes it easier to create reorgs, but then its security by obscurity.

chjj:
I think this is left-over code from bcoin. It's actually possible to do this on bcoin, but HSD adds a _lot_ more state. In theory we could get a snapshot of the previous tree state, but maybe better to just remove this.

pinheadmz:
@tynes didn't you patch this? I remebber you working on it.

pinheadmz:
> @tynes didn't you patch this? I remember you working on it.

pingy ping ping! @tynes

PR #330, opened 2020-01-14
[consensus] No transactions for the first two weeks

pinheadmz:
Before headers-first sync is implemented, full nodes are vulnerable to disk-fill attacks. The simplest solution to this is to reject forks prior to the latest checkpoint. See: https://github.com/bcoin-org/bcoin/pull/764.

This could be especially problematic for a low-work chain without any checkpoints.

In order to build up a decent amount of chainwork and give the developers a chance to release a patch with at least one buried block checkpoint, we've discussed eliminating all transactions for the first two weeks of mainnet.

This PR adds the consensus rule that rejects transactions from the mempool and blockchain for the first 14 days worth of blocks. This also means the original value for `auctionStart` had to be bumped from 10 to 14 days. I set the ban score for relaying transactions to 0, but including in a block to 100 (ban peer immediately).

Tests cover regular transactions as well as name Claims and Airdrop proofs.

The no-forks-prior-to-last-checkpoint rule is also implemented in this PR.

We should probably expect to release a patch for hsd before the 14 day window expires with a checkpoint from day 12 or something...

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2006,6 +2006,36 @@ class Chain extends AsyncEmitter {
       }
     }
 
+    // Transactions are not allowed in any block before
+    // a certain amount of chainwork has been accumulated.
+    // This a non-malleated, permanently invalid block
+    // and whoever sent it should be banned.
+    if (this.height < this.network.txStart) {
+      let invalid = false;
+
+      // No transactions allowed besides coinbase
+      if (block.txs.length > 1)
+        invalid = true;
+
+      // No claims or airdrops allowed in coinbase yet.
+      const cb = block.txs[0];
+      if (cb.outputs.length > 1)
+        invalid = true;
+
+      // Sanity check
+      if (!cb.outputs[0].covenant.isNone())
+        invalid = true;
+
+      if (invalid) {
+        this.setInvalid(block.hash());
+        throw new VerifyError(block,
+          'invalid',
+          'no-tx-allowed-yet',
+          100,
+          false);
+      }
+    }
+
     // Create a new chain entry.
     const entry = ChainEntry.fromBlock(block, prev);
 
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2018,7 +2018,7 @@ class Chain extends AsyncEmitter {
 
     // Transactions are not allowed in any block before
     // a certain amount of chainwork has been accumulated.
-    // This a non-malleated, permanently invalid block
+    // This is a non-malleated, permanently invalid block
     // and whoever sent it should be banned.
     if (prev.height + 1 < this.network.txStart) {
       let invalid = false;
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2020,7 +2020,7 @@ class Chain extends AsyncEmitter {
     // a certain amount of chainwork has been accumulated.
     // This a non-malleated, permanently invalid block
     // and whoever sent it should be banned.
-    if (this.height < this.network.txStart) {
+    if (prev.height + 1 < this.network.txStart) {
       let invalid = false;
 
       // No transactions allowed besides coinbase
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -1686,6 +1686,16 @@ class Chain extends AsyncEmitter {
    */
 
   async saveAlternate(entry, block, prev, flags) {
+    // Do not accept forked chain older than the
+    // last checkpoint.
+    if (this.options.checkpoints) {
+      if (prev.height + 1 < this.network.lastCheckpoint)
+        throw new VerifyError(block,
+          'checkpoint',
+          'bad-fork-prior-to-checkpoint',
+          100);
+    }
+
     try {
       // Do as much verification
       // as we can before saving.
--- a/test/chain-test.js
+++ b/test/chain-test.js
@@ -768,7 +768,8 @@ describe('Chain', function() {
 
   describe('Checkpoints', function() {
     before(async () => {
-      const entry = await chain.getEntry(chain.tip.height - 5);
+      const CHECKPOINT = chain.tip.height - 5;
+      const entry = await chain.getEntry(CHECKPOINT);
       assert(Buffer.isBuffer(entry.hash));
       assert(Number.isInteger(entry.height));
 
@@ -782,7 +783,8 @@ describe('Chain', function() {
     });
 
     it('will reject blocks before last checkpoint', async () => {
-      const entry = await chain.getEntry(chain.tip.height - 10);
+      const BEFORE_CHECKPOINT = chain.tip.height - 10;
+      const entry = await chain.getEntry(BEFORE_CHECKPOINT);
       const block = await cpu.mineBlock(entry);
 
       let err = null;
@@ -800,9 +802,9 @@ describe('Chain', function() {
     });
 
     it('will accept blocks after last checkpoint', async () => {
-      const entry = await chain.getEntry(chain.tip.height - 4);
+      const AFTER_CHECKPOINT = chain.tip.height - 4;
+      const entry = await chain.getEntry(AFTER_CHECKPOINT);
       const block = await cpu.mineBlock(entry);
-
       assert(await chain.add(block));
     });
   });
--- a/test/chain-test.js
+++ b/test/chain-test.js
@@ -112,9 +112,14 @@ chain.on('disconnect', (entry, block) => {
 describe('Chain', function() {
   this.timeout(45000);
 
-  it('should open chain and miner', async () => {
+  before(async () => {
     await chain.open();
     await miner.open();
+  });
+
+  after(async () => {
+    await miner.close();
+    await chain.close();
   });
 
   it('should add addrs to miner', async () => {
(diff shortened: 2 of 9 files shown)


tynes:
I think it would be a nice to have for a logline that says `Transactions are enabled` similar to:

https://github.com/handshake-org/hsd/blob/733e11166cc78b6987bb89dd52f14524343ff942/lib/blockchain/chain.js#L638-L642

pinheadmz:
@tynes logging is a bit awkward, and this applies to #305 as well.

Since we're not storing these rules in the deployment state (which I still think is correct) we don't have the ability to check if "now this rule is active, but it wasn't before".

We cant just say `if (this.height >= this.network.txStart)` because that would log on every single block after activation.

We could log exactly on `=== txStart` but then that would only happen once, when that block is reached. With the soft forks, you see those log messages when the chain loads (because chain starts with a null state, and then loads the real state from the db). I'm ok with this behavior if you are.

And finally, `setDeploymentState()` has extra logic that bypasses all the logging altogether once we past the last network checkpoint. I'm not sure why that is there, except that maybe buried deployments are not worth logging for the user.

tynes:
> Since we're not storing these rules in the deployment state (which I still think is correct)

Let me make sure I understand your reasoning: its because if they were in deployment state, they would be _optional_ if miners decided to run different software. Since they are consensus, they shouldn't exist in deployments.

> We could log exactly on === txStart but then that would only happen once, when that block is reached. With the soft forks, you see those log messages when the chain loads (because chain starts with a null state, and then loads the real state from the db). I'm ok with this behavior if you are.

I don't think the logline is necessary, it just felt like a nice to have. We don't need to add extra complexity just for something simple like that.

test/txstart-test.js
```diff
@@ -0,0 +1,226 @@
...
+const RESET_TXSTART = node.network.txStart;
+
+// Ease up that mempool
+node.mempool.options.minRelay = 0;
+node.mempool.options.rejectAbsurdFees = false;
+node.miner.options.minWeight = 1000000;
+
+// Use a valid, working airdrop proof
+const FAUCET_PROOF_FILE = resolve(__dirname, 'data', 'faucet-proof.base64');
+const rawFaucetProof =
```
tynes: Echoing boyma's comment on the goosig PR, this could fit on one line if you call it `raw` and change `FAUCET_PROOF_FILE` to `PROOF_FILE`

test/txstart-test.js
```diff
@@ -0,0 +1,226 @@
...
+  Buffer.from(fs.readFileSync(FAUCET_PROOF_FILE, 'binary'), 'base64');
+const proof = AirdropProof.decode(rawFaucetProof);
+
+// We only need this for the fakeClaim
+const wallet = new MemWallet({network: 'regtest'});
+node.chain.on('connect', (entry, block) => {
+  wallet.addBlock(entry, block.txs);
+});
```
tynes: Nit: space?

pinheadmz: I was hoping to indicate that all this crap (lines 40-51) is just to setup the MemWallet to create a fakeclaim. The `getNameStatus` function isn't called in the test suite, it's just a stub.

test/txstart-test.js
pinheadmz: thanks, force pushed

pinheadmz:
@tynes 

> Let me make sure I understand your reasoning: its because if they were in deployment state, they would be optional if miners decided to run different software. Since they are consensus, they shouldn't exist in deployments.

Yup. They aren't being "deployed" on an active network, they're part of the design like any other consensus rule.

> I don't think the logline is necessary, it just felt like a nice to have. We don't need to add extra complexity just for something simple like that.

Ok cool we'll leave it thanks!

lib/blockchain/chain.js
```diff
@@ -1686,6 +1686,16 @@ class Chain extends AsyncEmitter {
    */
```
tynes: Checkpoints default to `true` which is particularly important for the start of this chain to prevent the chain width expansion attack. :+1:

lib/blockchain/chain.js
```diff
@@ -2006,6 +2016,36 @@ class Chain extends AsyncEmitter {
...
+      const cb = block.txs[0];
+      if (cb.outputs.length > 1)
+        invalid = true;
+
+      // Sanity check
+      if (!cb.outputs[0].covenant.isNone())
+        invalid = true;
+
+      if (invalid) {
+        this.setInvalid(block.hash());
```
tynes: Good call on setting the block as invalid here, because it is normally called further down the call stack in a try/catch around `chain.setInvalid` :+1:

tynes:
One thought that comes to mind is adding the same `SPVNode` protections as in https://github.com/handshake-org/hsd/pull/305

This would protect users running spv and using cold storage for the first two weeks. I don't think there will be many users that use that configuration (if any) but it would save them from being banned.

test/chain-test.js
```diff
@@ -761,8 +766,44 @@ describe('Chain', function() {
...
+      network.checkpointMap[entry.height] = entry.hash;
+      network.lastCheckpoint = entry.height;
+    });
+
+    after(async () => {
+      network.checkpointMap = {};
+      network.lastCheckpoint = 0;
+    });
+
+    it('will reject blocks before last checkpoint', async () => {
```
tynes: This test is fragile to tree root updates breaking the call to `cpu.mineBlock` if the tree interval is crossed until https://github.com/handshake-org/hsd/issues/159 is fixed, although I don't expect that problem to ever be ran into. Tests in https://github.com/handshake-org/hsd/pull/305 are as well.

test/chain-test.js
```diff
@@ -761,8 +766,44 @@ describe('Chain', function() {
...
+      }
+
+      assert(err);
+      assert.equal(err.type, 'VerifyError');
+      assert.equal(err.reason, 'bad-fork-prior-to-checkpoint');
+      assert.equal(err.score, 100);
+    });
+
+    it('will accept blocks after last checkpoint', async () => {
+      const entry = await chain.getEntry(chain.tip.height - 4);
```
tynes: A constant that specifies the checkpoint height that is used here and when creating the checkpoint could make it more obvious that these blocks come after the last checkpoint. A comment could also suffice. This is good, just commenting on a way that I think we could make the test take less cognitive overhead to be sure of its correctness.

test/txstart-test.js
```diff
@@ -0,0 +1,225 @@
...
+
+    const block = new Block();
+    block.txs.push(tx1);
+    block.txs.push(tx2);
+    block.prevBlock = node.chain.tip.hash;
+    block.time = node.chain.tip.time + 2;
+    block.bits = await node.chain.getTarget(block.time, node.chain.tip);
+
+    await assert.rejects(node.chain.add(block, VERIFY_NONE),
+      {reason: 'no-tx-allowed-yet'});
```
tynes: An assertion that the block was added to the invalids cache could be good here:

```js
assert(node.chain.hasInvalid(block));
```

test/txstart-test.js
```diff
@@ -0,0 +1,225 @@
...
+    tx1.locktime = node.chain.height + 1;
+
+    const block = new Block();
+    block.txs.push(tx1);
+    block.prevBlock = node.chain.tip.hash;
+    block.time = node.chain.tip.time + 1;
+    block.bits = await node.chain.getTarget(block.time, node.chain.tip);
+
+    await assert.rejects(node.chain.add(block, VERIFY_NONE),
+      {reason: 'no-tx-allowed-yet'});
```
tynes: Another place where an assertion that the block was added to the invalids cache could be nice here.

test/txstart-test.js
```diff
@@ -0,0 +1,225 @@
...
+    tx1.outputs[0].address = addr;
+    tx1.locktime = node.chain.height + 1;
+
+    const block = new Block();
+    block.txs.push(tx1);
+    block.prevBlock = node.chain.tip.hash;
+    block.time = node.chain.tip.time + 3;
+    block.bits = await node.chain.getTarget(block.time, node.chain.tip);
+
+    await node.chain.add(block, VERIFY_NONE);
```
tynes: Wrapping this statement in an assert could be more robust here.

test/txstart-test.js
```diff
@@ -0,0 +1,225 @@
...
+    try {
+      ownership.ignore = true;
+      assert(await node.chain.add(block));
+    } finally {
+      ownership.ignore = false;
+    }
+
+    // This is the first block where transactions are allowed
+    assert.strictEqual(node.chain.height, node.network.txStart);
+
```
tynes: A comment summarizing the contents of the block would be helpful here, in particular the outputs in the coinbase.

tynes:
Overall great work, my recent comments are nice to haves. I think the invalid cache assertion is probably the most important one of the bunch.

pinheadmz:
@boymanjor wow thanks for picking all those nits! All the updates look good to me.

test/txstart-test.js
```diff
@@ -0,0 +1,228 @@
...
+      ownership.ignore = true;
+      await node.mempool.addClaim(claim);
+    } finally {
+      ownership.ignore = false;
+    }
+    assert.strictEqual(node.mempool.claims.size, 1);
+    assert(node.mempool.hasClaim(claim.hash()));
+  });
+
+  it('should allow airdrop in mempool at txStart height', async () => {
```
tynes: Since this test is specifically testing in context to a blockheight, it could use an assertion that the height is as expected. It requires context from previous tests to be certain that it is the correct height. This will also prevent regressions if any tests are changed in this file.

lib/blockchain/chain.js
chjj: Yeah, this rejection below should be the default in the future.

PR #331, opened 2020-01-18
p2p: implement two servers, plaintext + encrypted

tynes:
This PR adds a new p2p server that operates using plaintext, just like how Bitcoin works. The brontide p2p server still exists, but is not used by default. The brontide server is only used when the user specifically wants to use encryption, through configuration options `--nodes`, `--only` or the RPC method `addnode`.

This is to prevent nodes from gossiping around fake keys. It would be possible to fill up another node's hostlist with fake keys, essentially denial of servicing their ability to connect to the node. There is also the concern of a reflection attack where an adversary orchestrates many nodes in the network attempting to make an inbound connection to a single node at the same time. The brontide handshake adds additional overhead that could denial of service the node.

~~This PR still needs tests.~~
Tests have been implemented for both the `--only` option (which is a subset of `--nodes`) and the RPC `addnode`.
Closes https://github.com/handshake-org/hsd/issues/315


lib/net/pool.js
```diff
@@ -3318,6 +3364,10 @@ class Pool extends EventEmitter {
     if (!this.opened)
       return;
 
+    // TODO: this saves compute by exiting early
+    // Now there are two different settings. Without
+    // an argument to `addOutbound`, there is no way to
+    // exit early like this.
```
tynes: I think this problem could be solved with something like:

```js
if (this.peers.outbound >= this.options.maxOutbound
    && this.peers.outboundBrontide >= this.options.maxOutboundBrontide)
  return;
```

Then it would involve determining which type (brontide or plaintext) could be added and getting that type from `getHost`

lib/net/hostlist.js
```diff
@@ -787,6 +787,24 @@ class HostList {
...
   }
 
+  /**
+   * Convert address list to a filtered
+   * array. Only addresses without keys
+   * are included.
+   *
+   * @returns {NetAddress[]}
+   */
+
```
tynes: I don't love the name of this method. Would like to rename it to something better.

kilpatty:
Looking over everything now - quick question that I am wondering. Do we maintain separate ban lists?

I can just envision a scenario where my Node (Node 1) has 2 servers open (1a non-brontide p2p, 1b brontide p2p). Someone gossips my 1b IP:Port combination without the identity key. Nodes now start to attempt to connect to me, but because I'm expecting a Brontide Handshake and then an encrypted message - I can't interpret the packets and therefore ban that peer. If that ban persists across brontide and non-brontide connections, I could see that potentially becoming a problem of targeted bans against peers.

lib/net/hostlist.js
pinheadmz: You could add a boolean to `toArray()` to filter and keep it as one function?

lib/protocol/networks.js
```diff
@@ -59,13 +59,23 @@ main.seeds = [];
...
 main.port = 12038;
 
+/**
+ * Default brontide network port.
+ * Used for encrypted communications.
+ * @const {Number}
+ * @default
+ */
+
+main.brontidePort = 12036;
```
pinheadmz: Don't forget to add `this.brontidePort = options.brontidePort` in `network.js`

lib/net/peer.js
```diff
@@ -242,10 +258,12 @@ class Peer extends EventEmitter {
...
-    this.brontide.once('error', (err) => {
-      this.error(err);
-      this.destroy();
-    });
+    if (this.encrypt) {
+      this.brontide.once('error', (err) => {
+        this.error(err);
+        this.destroy();
+      });
+    }
```
pinheadmz: I'm not sure you need to `if` this - looking at the constructor, every peer still has a `this.brontide = BrontideStream()` whether or not it gets used.

lib/net/peer.js
```diff
@@ -264,7 +282,11 @@ class Peer extends EventEmitter {
       this.handleDrain();
     });
 
-    this.brontide.on('data', (chunk) => {
+    // Listen to the data event on brontide if
+    // encryption is enabled, otherwise just listen
+    // to the socket itself.
+    const listener = this.encrypt ? this.brontide : this.socket;
```
pinheadmz: You have this logic twice, I wonder if part of `init()` should be setting a single `listener` for the instance?

lib/net/pool.js
```diff
@@ -4214,6 +4264,10 @@ class PoolOptions {
     this.port = this.network.port;
     this.publicHost = '0.0.0.0';
     this.publicPort = this.network.port;
+    this.brontideHost = '0.0.0.0';
+    this.brontidePort = this.network.brontidePort;
+    this.publicBrontideHost = '0.0.0.0';
+    this.publicBrontidePort = this.network.brontidePort;
```
pinheadmz: These options may need to be added to `FullNode` and `SPVNode` as well so they can be set from the config file, etc.

pinheadmz:
Great work dude! Have you explored adding an argument to `getHost()` to select hosts with/without key? In a more generic `createOutbound` case, the method could "prefer" brontide peers.

pinheadmz:
@kilpatty the ban list is IP-only (all ports, all keys), I think it makes sense to keep it that way. I also don't think we ban if Brontide handshake fails. So I don't think this is "any worse" than the kind DDoS we are already vulnerable to. It would be nice if we could configure nodes to recognize the Brontide port and not even try to connect without an ID key, but users can set their ports to whatever they want so, it would just be a guess.

kilpatty:
@pinheadmz Ah good point, but doesn't not banning brontide handshake failures essentially allow for anyone to hold the inbound slots of another node hostage? As soon as they receive a disconnect they just reconnect since there is no back-off or ban. I suppose that is another good reason to make Brontide optional

lib/protocol/networks.js
tynes: Thanks for the reminder, this was just addressed in https://github.com/handshake-org/hsd/pull/331/commits/532f4848036e424de514010a47abbd103292dcfe

lib/net/peer.js
tynes: I don't think there is a need to set up listeners when they will never be used

lib/net/pool.js
```diff
@@ -597,9 +628,9 @@ class Pool extends EventEmitter {
 
     const host = IP.toHostname(ip, socket.remotePort);
 
-    assert(!this.peers.map.has(host), 'Port collision.');
+    assert(!this.peers.map.has(host), 'Port collision (clear).');
```
tynes: Nit: no need to have `clear` here, this is an artifact of when two peers lists were being maintained. I decided it would be more simple to just maintain a single peers list.

lib/net/pool.js
tynes: These options were added to `FullNode` but there is no need to add them to `SPVNode`, as it is configured with `listen: false` so it will not accept incoming connections.

https://github.com/handshake-org/hsd/blob/master/lib/node/spvnode.js#L76
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/pool.js#L366-L367

tynes: See 0e3da111539d67283c529b781310524db2480920

tynes:
We still need to properly expose the number of inbound/outbound peers in the HTTP/RPC API as the count of encrypted/plaintext are maintained separately. I don't want to blow up `GET /` more since it is already huge, maybe a `GET /peers` endpoint would be useful or adding the info to the RPC `getnetworkinfo`

lib/net/peer.js
tynes: Could you expand on this comment if its still relevant please?

tynes:
> Have you explored adding an argument to getHost() to select hosts with/without key? In a more generic createOutbound case, the method could "prefer" brontide peers.

This is a good thought. I haven't implemented filling up outgoing brontide peers yet, mostly because I haven't fully decided what the intended behavior should be.

If it loops constantly trying to fill up the outbound (like in Bitcoin), then it will keep going through `BrontideStream.connect`:
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/brontide.js#L492

I don't think its a huge amount of overhead, but we should also try to be as lightweight as possible.

tynes:
I was able to effectively develop this using this combination of a patch and bash script.

```diff
diff --git a/bin/node b/bin/node
index 3a3aa158..94a2909a 100755
--- a/bin/node
+++ b/bin/node
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env -S node debug
 
 'use strict';
 

```

```bash
#!/bin/bash

NETWORK="$1"
CMD=./bin/hsd

if [ "$2" = "debug" ]; then
    CMD=./bin/node
fi

$CMD --memory true \
    --network regtest \
    --host "$1" \
    --brontide-host "$1" \
    --http-host "$1" \
    --rs-host "$1" \
    --ns-host "$1" \
    --wallet-http-host "$1" \
    --api-key foo \
    --listen true \
    --memory true \
    --max-inbound 8
```

test/peering-test.js
```diff
@@ -0,0 +1,176 @@
...
+    if (one)
+      await one.close();
+
+    if (two)
+      await two.close();
+
+    one = null, two = null;
+  });
+
+  it('should create peers (plaintext)', async () => {
```
tynes: I'd like to deduplicate these two tests using a `for` loop since they share a lot of the same logic. The main difference is what is asserted in the `'peer connect'` event callback function.

tynes:
@kilpatty 

> I can't interpret the packets and therefore ban that peer

If a cleartext connection attempt occurs on the brontide server ip:port, this error will be thrown:
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/brontide.js#L266-L267

Which will be caught here:
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/brontide.js#L633-L642

The emitted error will be caught here:
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/pool.js#L111-L113

The codepath never makes it to the `'connect'` event being emitted:
https://github.com/handshake-org/hsd/blob/6afb18b134f97ec702b268a52f7673b8aa4e132b/lib/net/brontide.js#L674

Which causes the peer connection to timeout and the peer is then removed, not banned.

lib/net/peer.js
pinheadmz: The variable name changed but still relevant. I was thinking its a bit more on-style for us to set an instance-level property during `init()` or `constructor()`:

`this.channel = this.encrypt ? this.brontide : this.socket;`

and then throughout the rest of the class just use `this.channel...` instead of having the switching logic in multiple places.

lib/net/peer.js
pinheadmz: I guess it just seems cleaner, but I don't care enough to push it. I just think the fact that the instance still has a `BrontideStream` means that it is in use, and if it's ever opened by some external caller, it's not going to _hurt_ to have an error listener.

lib/net/pool.js
pinheadmz: re: SPV -- oh yeah! 👍

lib/net/peer.js
tynes: Event listeners do add some resource usage and can result in memory leaks if they are not removed

lib/net/netaddress.js
```diff
@@ -170,7 +170,7 @@ class NetAddress extends bio.Struct {
    */
 
   hasKey() {
-    return this.key.equals(ZERO_KEY) !== 0;
+    return !this.key.equals(ZERO_KEY);
   }
```
tynes: This is a bugfix, it would return the opposite value that it was supposed to.

lib/net/peer.js
tynes: I do think this is a good idea

tynes:
This code path hard codes in the `port` and `publicPort` and has not been updated to work with the brontide ports.

https://github.com/handshake-org/hsd/blob/733e11166cc78b6987bb89dd52f14524343ff942/lib/net/pool.js#L437-L439

Do we want 1 to 1 parity between the two listening servers or are there things that we can do to make the brontide port more secretive?

I don't think we need to have this codepath work with the brontide port since `UPNP.discover` sends a request to `239.255.255.250:1900`. This just alerts that server that something is trying to traverse a NAT.

See:
https://github.com/bcoin-org/bupnp/blob/master/lib/upnp.js#L29

```js
    const msg = ''
      + 'M-SEARCH * HTTP/1.1\r\n'
      + `HOST: ${this.host}:${this.port}\r\n`
      + 'MAN: ssdp:discover\r\n'
      + 'MX: 10\r\n'
      + 'ST: ssdp:all\r\n';
```

https://github.com/bcoin-org/bupnp/blob/c44fa7356aa297c9de96e8ad094a6816939cd688/lib/upnp.js#L129

Note the comment:
```
 // UPNP is always optional, since
 // it's likely to not work anyway.
```

lib/net/peer.js
tynes: > I guess it just seems cleaner

I do agree that it looks way cleaner and adds less mental overhead when reasoning about the codepath.

lib/net/peer.js
tynes: We would need to do some refactoring to be able to set `this.channel` in the `constructor` or `init`. Right now, I am not passing the `encrypt` property in the peer options object as to not mutate or need to copy the Pool's `this.options`

https://github.com/handshake-org/hsd/blob/733e11166cc78b6987bb89dd52f14524343ff942/lib/net/pool.js#L1112

It is set externally from the constructor:

```js
  static fromOutbound(options, addr) {
    const peer = new this(options);
    if (addr.hasKey())
      peer.encrypt = true;
```

lib/net/hostlist.js
```diff
@@ -787,23 +787,41 @@ class HostList {
...
+  }
+
   /**
    * Add a preferred seed.
    * @param {String} host
    */
 
   addSeed(host) {
-    const ip = IP.fromHostname(host, this.network.port);
+    let ip = IP.fromHostname(host, this.network.port);
```
tynes: Some work could be saved here since `IP.fromHostname` is potentially called twice which duplicates work. I did this for simplicity's sake, since the `ip.key` branch should be the minority. A `hasKey` function could be helpful here for getting the correct port and then calling `IP.fromHostname` a single time.

lib/net/peer.js
```diff
@@ -285,7 +314,13 @@ class Peer extends EventEmitter {
...
     this.outbound = false;
     this.connected = false;
-    this.brontide.accept(socket, this.identityKey);
+
+    if (this.encrypt) {
+      this.brontide.accept(socket, this.identityKey);
+    } else {
+      this.connected = true;
+      this.time = Date.now();
+    }
```
tynes: This `else` block comes directly from the `bcoin` codepath

lib/node/fullnode.js
```diff
@@ -87,6 +87,7 @@ class FullNode extends Node {
       chain: this.chain,
       mempool: this.mempool,
       prefix: this.config.prefix,
+      agent: this.config.str('agent'),
```
tynes: I'm using this `agent` string in the tests to exactly identity peers. I'm doing this to prevent the need to create custom identity keys.

pinheadmz:
Just had a thought: should we update the seeds list as well? I think it makes sense if the official seeds have public ID keys, but imagine the seed nodes' brontide inbound connections is full -- they may still have open slots on the clearnet server. So perhaps one easy thing to do for the seeds list would be to duplicate all the seed nodes, without the ID key and with the other network port. Bootstrapping nodes would try the encrypted connections first, then fall back to clearnet.

tynes:
Needs rebase after 536da2d3190151dbe6d67e462859656e9cda8f7b was merged.

tynes:
> Just had a thought: should we update the seeds list as well? I think it makes sense if the official seeds have public ID keys, but imagine the seed nodes' brontide inbound connections is full -- they may still have open slots on the clearnet server. So perhaps one easy thing to do for the seeds list would be to duplicate all the seed nodes, without the ID key and with the other network port. Bootstrapping nodes would try the encrypted connections first, then fall back to clearnet.

Good thinking @pinheadmz, there is a similar open issue here https://github.com/handshake-org/hsd/issues/335

Maybe whoever is operating the seed can decide if they want to accept inbound encrypted connections? Your suggestion works for sure but would require some additional code in the hostlist. We just don't want to end up in a situation where the a node is in an infinite loop trying to connect to a seed node over brontide when it doesn't accept inbound brontide connections.

tynes:
Note to reviewers: we should make sure that all event listeners are being cleaned up properly as to prevent memory leaks.

test/peering-test.js
```diff
@@ -0,0 +1,249 @@
...
+        assert.equal(two.node.pool.peers.outboundBrontide, 0);
+
+        // The peer is not encrypted.
+        assert.equal(peer.encrypt, false);
+        // The peer's address does not have a key.
+        assert.equal(peer.address.hasKey(), false);
+
+        // The expected peer has connected.
+        assert.deepEqual(peer.realname(), addr);
+        assert.deepEqual(peer.hostname(), addr);
```
pinheadmz: These are both strings, `assert.equal()` should be sufficient

test/peering-test.js
```diff
@@ -0,0 +1,249 @@
...
+
+      const two = await factory.new({
+        only: [addr]
+      });
+
+      nodes.push(one, two);
+
+      two.node.pool.on('peer connect', (peer) => {
+        // Peer one has a single inbound connection.
+        assert.equal(one.node.pool.peers.inbound, 1);
```
pinheadmz: this assertion fails intermittently on repeat testing... I wonder if you should listen for an event directly from node `one` and test it separately to avoid race condition.

test/peering-test.js
```diff
@@ -0,0 +1,249 @@
...
+      nodes.push(one, two, three);
+
+      await common.event(three.node.pool, 'peer connect');
+
+      const peers = await one.nclient.execute('getpeerinfo');
+
+      assert.equal(peers.length, 2);
+
+      const encrypted = peers.find(p => p.encrypted);
+      const plaintext = peers.find(p => !p.encrypted);
```
pinheadmz: Clever! 👍

lib/node/rpc.js
```diff
@@ -485,11 +485,13 @@ class RPC extends RPCBase {
...
       peers.push({
         id: peer.id,
-        addr: peer.fullname(),
+        addr: peer.realname(),
         addrlocal: !peer.local.isNull()
           ? peer.local.fullname()
           : undefined,
         name: peer.name || undefined,
+        encrypted: peer.encrypt,
+        identitykey: peer.address.getKey('base32'),
```
pinheadmz: For clearnet peers, this will display the always-hilarious bech32 string:
` identitykey: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',`

tynes: Do you think it should return `null` instead?

test/peering-test.js
tynes: Good catch, I haven't noticed this on my machine at all. Will fix the race condition

lib/node/rpc.js
pinheadmz: Maybe, or not display anything. I guess `null` is better. Could this be written in the actual `NetAddress.getKey()` function to return null if it equals `ZERO_KEY` ?

lib/net/peer.js
```diff
@@ -205,13 +214,27 @@ class Peer extends EventEmitter {
...
     return this.address.fullname();
   }
 
+  /**
+   * Getter to retrieve realname. Returns
+   * the fullname when encrypted, otherwise
+   * the hostname.
+   * @returns {String}
+   */
+
```
tynes: We could opt to just using `fullname` instead of adding `realname`, `fullname` will never render the key. `realname` is more useful if we ever want to choose to create an outbound connection in plaintext when we also have the key.

pinheadmz: oh yeah, you've got this already - so maybe we dont need to add identity key again in `getpeerinfo`

pinheadmz: this already looks ok:

```   
{
    addr: 'aorsxa4ylaacshipyjkfbvzfkh3jhh4yowtoqdt64nzemqtiw2whk@127.0.0.1:51363',
    addrlocal: 'aorsxa4ylaacshipyjkfbvzfkh3jhh4yowtoqdt64nzemqtiw2whk@127.0.0.1:10006',
    encrypted: true,
    ...
  },
  {
    id: 2,
    addr: '127.0.0.1:51368',
    addrlocal: '127.0.0.1:10005',
    encrypted: false,
    ...
```

lib/net/hostlist.js
pinheadmz: `IP.fromHostname()` returns a [dumb object](https://github.com/bcoin-org/binet/blob/224bbfac6056e0480351f430486f200ca6bc2c25/lib/ip.js#L544-L551), I wonder if you could just set it quickly after the first call:

```
if (ip.key && ip.port === this.network.port)
  ip.port == this.network.brontidePort;
```

The intent would be "if the port is the fallback port for clearnet, assume no custom port was provided. Since we have a key, we should use the brontide fallback port"

This would only fail if someone specifically had a hostname WITH a key, but also used the default clearnet port. Which would be a stupid edge case... so maybe the best way is what you have... or an update to `binet`

lib/net/peer.js
```diff
@@ -62,6 +62,7 @@ class Peer extends EventEmitter {
     this.socket = null;
     this.brontide = new BrontideStream();
     this.identityKey = this.options.identityKey;
```
tynes: The options above are not actually receiving `PeerOptions` but `PoolOptions`. Passing `PeerOptions` will cause it to break in the cleartext case.

lib/net/peer.js
```diff
@@ -137,11 +138,15 @@ class Peer extends EventEmitter {
    * Create inbound peer from socket.
    * @param {PeerOptions} options
    * @param {net.Socket} socket
+   * @param {Boolean?} brontide
    * @returns {Peer}
    */
 
-  static fromInbound(options, socket) {
```
tynes: Maybe expose `PeerOptions` and construct in `Pool`, actually pass in `PeerOptions` to these functions. This will eliminate the need to have the `brontide` arg since it can exist in `PeerOptions`.

lib/net/peer.js
tynes: If trying to connect to the brontide server when using the plaintext protocol, this should definitely keep the listener. Need to verify that.

lib/net/peer.js
tynes: Would it make sense to build an interface that matches the `BrontideStream` for the regular `socket`?

lib/net/pool.js
```diff
@@ -1126,11 +1177,12 @@ class Pool extends EventEmitter {
    * Accept an inbound socket.
    * @private
    * @param {net.Socket} socket
+   * @param {Boolean?} brontide
    * @returns {Peer}
    */
 
-  createInbound(socket) {
-    const peer = Peer.fromInbound(this.options, socket);
+  createInbound(socket, brontide) {
```
tynes: Create `PeerOptions` here?

lib/net/peer.js
tynes: Verified that we should set the listener in case somebody tries to connect in clear on the brontide port. See https://github.com/handshake-org/hsd/pull/331#issuecomment-577457042

lib/net/hostlist.js
tynes: The reason why I don't love mutating the `ip` object returned from `IP.fromHostName` is because it builds additional properties on that object that would fall out of sync with the change we made, see https://github.com/bcoin-org/binet/blob/master/lib/ip.js#L542.

Its possible to create a `hasKey` function, I feel like that would be better off in `binet` so that it can be consumed internal to the library as well. Maybe we can update `binet` later and go with this for now? Or just write the `hasKey` function and keep it as a helper in this file until we can move it into `binet.

lib/net/peer.js
tynes: Resolved by using `PeerOptions`

lib/net/peer.js
```diff
@@ -137,11 +139,13 @@ class Peer extends EventEmitter {
    * Create inbound peer from socket.
    * @param {PeerOptions} options
    * @param {net.Socket} socket
+   * @param {Boolean?} brontide
```
tynes: This docstring param can be deleted.

lib/net/pool.js
```diff
@@ -3235,24 +3301,25 @@ class Pool extends EventEmitter {
...
       'Unknown packet: %d (%s).',
-      packet.type, peer.hostname());
+      packet.type, peer.fullname());
   }
 
   /**
    * Create an inbound peer from an existing socket.
    * @private
    * @param {net.Socket} socket
+   * @param {Boolean?} brontide
```
tynes: This is no longer optional

lib/node/rpc.js
tynes: Updated this functionality to return `null` if there is no key

lib/net/peer.js
tynes: Updating the above comment. The `STANDARD` case is directly from the `bcoin` codepath.

lib/net/peer.js
```diff
@@ -391,6 +420,11 @@ class Peer extends EventEmitter {
    */
 
   initConnect() {
```
tynes: This code is directly from the `bcoin` codepath and is used to prevent waiting on `socket.once` forever.

lib/net/pool.js
```diff
@@ -4214,8 +4322,14 @@ class PoolOptions {
...
     this.publicPort = this.network.port;
+    this.brontideHost = '0.0.0.0';
+    this.brontidePort = this.network.brontidePort;
+    this.brontidePublicHost = '0.0.0.0';
+    this.brontidePublicPort = this.network.brontidePort;
     this.maxOutbound = 8;
     this.maxInbound = 8;
+    this.brontideMaxInbound = 2;
+    this.brontideMaxOutbound = 0;
     this.createSocket = this._createSocket.bind(this);
```
tynes: The default values for inbound and outbound connections are here. I set the max brontide inbound to 2 as to not add a lot of extra bandwidth usage. The max outbound is set to 0 as to not infinitely loop on trying to connect to outbound peers using brontide. If `hnsd` only supports brontide at first, then bumping the `brontideMaxInbound` could be helpful.

tynes:
A really nice to have would be to prefer brontide based peers for spv nodes that need to pick a prover. See `Pool.pickProver`

tynes:
This is ready for another round of review.

tynes:
> Just had a thought: should we update the seeds list as well? I think it makes sense if the official seeds have public ID keys, but imagine the seed nodes' brontide inbound connections is full -- they may still have open slots on the clearnet server. So perhaps one easy thing to do for the seeds list would be to duplicate all the seed nodes, without the ID key and with the other network port. Bootstrapping nodes would try the encrypted connections first, then fall back to clearnet.

Thinking about this more, I think its more simple to just have the seed node publish two seed entries, one with a key and one without a key. This makes the bookkeeping way more simple. Then the seeds have complete control over the types of connections they advertise.

lib/net/pool.js
```diff
@@ -490,6 +531,7 @@ class Pool extends EventEmitter {
    * @returns {Promise}
    */
 
+  // TODO: does this need brontide solution?
```
tynes: Now that I think about it, this doesn't need any work to accommodate for brontide unless we were using bolt10 because the identity keys would not fit in `A` or `AAAA` records. If we did write a bolt10 server/client, this is where it could be used.

boymanjor: nit: `TODO(tynes):`

lib/net/pool.js
```diff
@@ -1126,11 +1186,16 @@ class Pool extends EventEmitter {
...
    * @param {net.Socket} socket
+   * @param {Number} type
    * @returns {Peer}
    */
 
-  createInbound(socket) {
-    const peer = Peer.fromInbound(this.options, socket);
+  createInbound(socket, type) {
+    assert(typeof type === 'number');
+    const options = new PeerOptions(this.options);
```
boymanjor: Should we pass type to `PeerOptions`? Or create `StandardPeerOptions` & `BrontidePeerOptions`? Maybe this is overkill, but I don't like that a caller could miss this crucial property.

lib/net/hostlist.js
boymanjor: `toStandardArray()` or `toStandardList()`?

lib/net/pool.js
boymanjor: We should add a `TODO` about bumping `brontideMaxInbound` if `hnsd` is brontide-only.

lib/net/pool.js
```diff
@@ -4342,6 +4461,31 @@ class PoolOptions {
...
+      assert(typeof options.host === 'string');
+      const raw = IP.toBuffer(options.brontideHost);
+      this.brontideHost = IP.toString(raw);
+      if (IP.isRoutable(raw))
+        this.brontidePublicHost = this.host;
+    }
+
+    if (options.brontidePort != null) {
+      assert((options.brontidePort & 0xffff) === options.brontidePort);
+      this.brontidePort  = options.brontidePort;
```
boymanjor: nit: double space before assignment operator.

lib/net/pool.js
```diff
@@ -4342,6 +4461,31 @@ class PoolOptions {
...
+    }
+
+    if (options.brontidePublicPort != null) {
+      assert((options.brontidePublicPort & 0xffff)
+        === options.brontidePublicPort);
+      this.brontidePublicPort = options.brontidePublicPort;
+    }
+
     if (options.maxOutbound != null) {
       assert(typeof options.maxOutbound === 'number');
```
boymanjor: We should assert that inbound and outbound maximums are integers.

lib/net/peer.js
```diff
@@ -62,6 +62,9 @@ class Peer extends EventEmitter {
     this.socket = null;
     this.brontide = new BrontideStream();
     this.identityKey = this.options.identityKey;
+    this.type = this.options.type;
+
```
boymanjor: nit: is this necessary?

lib/net/peer.js
```diff
@@ -142,6 +145,7 @@ class Peer extends EventEmitter {
 
   static fromInbound(options, socket) {
     const peer = new this(options);
+
```
boymanjor: nit: is this necessary?

lib/net/peer.js
```diff
@@ -155,6 +159,7 @@ class Peer extends EventEmitter {
 
   static fromOutbound(options, addr) {
     const peer = new this(options);
+
```
boymanjor: nit: ditto above.

lib/net/peer.js
```diff
@@ -2145,6 +2221,29 @@ class PeerOptions {
...
+      this.peerTypes = options.peerTypes;
+    }
+
+    if (options.peerTypesByVal != null) {
+      assert(typeof options.peerTypesByVal === 'object');
+      this.peerTypesByVal = options.peerTypesByVal;
+    }
+
+    if (options.type != null) {
+      assert(typeof options.type === 'number');
```
boymanjor: We should assert this is an integer.

boymanjor: Better yet, we should assert that this is one of our peer types.

lib/net/peer.js
```diff
@@ -2235,6 +2344,28 @@ class PeerOptions {
...
+
+  /**
+   * Get the peer type by address.
+   * See PoolOptions.peerTypes.
+   * @private
+   * @param {NetAddress}
+   * @returns {Number}
+   */
+
+  static getTypeByAddr(addr) {
```
boymanjor: Why do this and `static getPeerType()` exist?

lib/net/pool.js
```diff
@@ -4436,7 +4590,9 @@ class PoolOptions {
       assert(Array.isArray(options.only));
       if (options.only.length > 0) {
         this.nodes = options.only;
-        this.maxOutbound = options.only.length;
+        const [standard, brontide] = this.countHostnames(this.nodes);
```
boymanjor: I think we should allow for a smaller max outbound passed through the config to take precedence over the number of addrs passed in `only`. E.g. I pass 4 standard peers in `only` but I only want a max of two of them to be connected at a time.

lib/net/peer.js
tynes: It follows the same pattern of the other `static` functions, to set the initial values of the class methods. These class methods are then overwritten by the functions that are passed in.

lib/net/peer.js
tynes: I think asserting that it is an integer would fragment the style of the codebase.

```bash
$ grep -rin "=== 'number'" lib | wc -l
97
```

```bash
$ grep -rin "Number.isInteger" lib | wc -l
0
```

lib/net/pool.js
tynes: I think this functionality is achievable with a combination of `--nodes` and `--max-outbound`. If not, maybe we can focus on this in the future.

lib/net/pool.js
```diff
@@ -3291,6 +3370,14 @@ class Pool extends EventEmitter {
       if (!addr.isValid())
         continue;
 
+      if (type === this.options.peerTypes.STANDARD)
```
chjj: nit: multi-line if statements should have curly braces.

lib/net/pool.js
```diff
@@ -3291,6 +3370,14 @@ class Pool extends EventEmitter {
...
+        if (addr.hasKey())
+          continue;
+
+      if (type === this.options.peerTypes.BRONTIDE)
```
chjj: ditto

lib/net/peer.js
```diff
@@ -285,7 +296,19 @@ class Peer extends EventEmitter {
     this.address.services = 0;
     this.outbound = false;
     this.connected = false;
-    this.brontide.accept(socket, this.identityKey);
+
+    switch (this.type) {
+      case this.options.peerTypes.STANDARD:
```
chjj: Is there a reason these constants reside on the options object? I feel like they should just be global constants.

tynes: It was so that they could be shared between the `Pool` and `Peer` files easily but that could also be done by exposing them as a static `Peer.peerTypes`

lib/net/pool.js
tynes: I added checks for each of the inbound/outbound integers

tynes:
@chjj @boymanjor @pinheadmz 

I've addressed many of the comments, adding some additional conversation here.

> We should add a TODO about bumping brontideMaxInbound if hnsd is brontide-only.

The default values are:
- `brontideMaxInbound = 2`
- `brontideMaxOutbound = 0`

Since most users will not alter the default values, do these seem sane to start? They can always be tweaked in newer releases. For SPV, the Pool's prover Peer isn't yet prioritized for selecting brontide based peers. `hnsd` is currently brontide only, so there may not be many fullnodes out there that are accepting public inbound brontide peers with these numbers. We definitely want name resolution to happen in an encrypted manner.

----

For `Pool.addOutbound`:

```
    // Try to connect to a brontide peer 20% of
    // the time when a the outbound brontide peers
    // are not full.
```

I picked that number sort of arbitrarily, but with the intention of having it loop less on trying to create new brontide outbound peers when its `brontideMaxOutbound` isn't full.

chjj:
I think that's fine to start. Merging.

PR #332, opened 2020-01-19
Never activate soft fork deployments (for now)

pinheadmz:
`hsd` ships on day 1 with two soft forks defined:
- `hardening`: a rule to reject 1024-bit RSA keys from Airdrop proofs
- `rollover`: a rule to require all DNSSEC Claims be signed by the 2017 ICANN key-signing-key, rejecting the 2010 key

Both of these soft forks are designed to activate based on [BIP9 (versionbits)](https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki)signaling by miners. BIP9 defines the miner-signaling window with `startTime` and `timeout` as UNIX timestamps.

Currently, the start times for main net are set for June 1, 2019. This means that when mainnet launches, both soft forks will activate after three 2-week (aka `minerWindow`) periods:

`DEFINED`->`STARTED`->`LOCKED_IN`->`ACTIVE`

...and so both rules will be enforced at height `6049` (I might be off by one)

At this time we do not know when these soft forks will be necessary to enforce. From the whitepaper:

> In the case that 1024 bit RSA is compromised, we turn to miner consensus to resolve the issue. This allows the blockchain to support RSA-1024 until a practical attack is demonstrated against it.
> ...
> Our primary concern is that ICANN may decide to revoke KSK-2010 at some point by publishing its corresponding private key. To deal with this issue, a final consensus rule can be added to disable KSK-2010 once a separate proof is published which demonstrates that KSK-2017 is now active. 

So this PR is pretty simple: we postpone the `startTime` of our two soft forks until `0xffffffff` (02/07/2106 @ 6:28am UTC) which gives us ample time to accept the initial chain rules. When a soft fork is necessary to deploy, the community can set the `startTime` to a practical value and hopefully miners will upgrade, signal support, and activate the new rules in a timely manner.

Generic BIP9 tests are added, adopted from https://github.com/bcoin-org/bcoin/pull/724 which adds BIP9 signaling statistics to `rpc getblockchaininfo`. I didn't add the new rpc functionality to this PR to keep review simple.

### DIFF
--- a/lib/protocol/networks.js
+++ b/lib/protocol/networks.js
@@ -392,8 +392,8 @@ main.deployments = {
   hardening: {
     name: 'hardening',
     bit: 0,
-    startTime: 1559347200, // June 1st, 2019
-    timeout: 1654041600, // June 1st, 2022
+    startTime: 0xffffffff,
+    timeout: 0,
     threshold: -1,
     window: -1,
     required: false,
@@ -402,8 +402,8 @@ main.deployments = {
   rollover: {
     name: 'rollover',
     bit: 1,
-    startTime: 1559347200, // June 1st, 2019
-    timeout: 1654041600, // June 1st, 2022
+    startTime: 0xffffffff,
+    timeout: 0,
     threshold: -1,
     window: -1,
     required: false,
@@ -635,8 +635,8 @@ testnet.deployments = {
   hardening: {
     name: 'hardening',
     bit: 0,
-    startTime: 1559347200, // June 1st, 2019
-    timeout: 1654041600, // June 1st, 2022
+    startTime: 0xffffffff,
+    timeout: 0,
     threshold: -1,
     window: -1,
     required: false,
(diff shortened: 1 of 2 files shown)


pinheadmz:
The effect of the current `startTime`'s being in the past are observable on regtest:

`hsd-rpc --network=regtest generatetoaddress 440 ...` 

From the logs:

```
[info] (chain) Block 05a82fe8cf678c45d3a3541feaee1aa7636e6d99a39d155dfcef64ec97438535 (431) added to chain (size=351 txs=1 time=0.751243).
[debug] (miner) Created block tmpl (height=432, weight=4000, fees=0, txs=1, diff=0, bits=545259519).
[info] (chaindb) Saving 3 state cache updates.
[warning] (chain) RSA hardening has been activated.
[warning] (chain) KSK-2017 rollover has been activated.
```

pinheadmz:
@boymanjor added comments

boymanjor:
Closing in favor #337. Discussion is still active re: how long to delay the soft fork start times.

PR #333, opened 2020-01-23
wallet: port various fixes and improvements from bcoin

pinheadmz:
This is a port of the following bcoin PRs:

https://github.com/bcoin-org/bcoin/pull/931 wallet: set tip after txs have been added

https://github.com/bcoin-org/bcoin/pull/932 Prevent out-of-memory crash during sync with wallet as plugin 

https://github.com/bcoin-org/bcoin/pull/934 wallet: remove checkpoints that skip transactions

and also I applied:

https://github.com/handshake-org/hsd/pull/181 wallet: remove dead option

to one other place (`lib/wallet/plugin.js').

What we're doing here is removing dead code, keeping walletDB in sync with the chain without filling up memory, and preventing possible walletDB corruption after a crash.

### DIFF
--- a/lib/node/spvnode.js
+++ b/lib/node/spvnode.js
@@ -7,7 +7,6 @@
 'use strict';
 
 const assert = require('bsert');
-const {Lock} = require('bmutex');
 const NameState = require('../covenants/namestate');
 const Chain = require('../blockchain/chain');
 const Pool = require('../net/pool');
@@ -112,10 +111,6 @@ class SPVNode extends Node {
       noUnbound: this.config.bool('rs-no-unbound')
     });
 
-    this.rescanJob = null;
-    this.scanLock = new Lock();
-    this.watchLock = new Lock();
-
     this.init();
   }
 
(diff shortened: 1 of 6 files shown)


lib/node/spvnode.js
```diff
@@ -220,64 +203,11 @@ class SPVNode extends Node {
    * Scan for any missed transactions.
    * Note that this will replay the blockchain sync.
    * @param {Number|Hash} start - Start block.
-   * @param {Bloom} filter
-   * @param {Function} iter - Iterator.
-   * @returns {Promise}
-   */
-
```
tynes: Would the removal of this code prevent any wallet rescans?

pinheadmz: SPV never really rescanned since it doesn't have block data. All it can do is reset and redownload the chain. The reset logic remains and will be run in wallet as a plug-in mode. The only thing actually killed by this PR is if the wallet is run as a separate process against an spv node (which is problematic for other reasons) and requests a rescan.

tynes: Thank you for clarifying @pinheadmz, makes sense.

pinheadmz:
This needs an update, closing until it can be fixed.

pinheadmz:
This PR now also ports over https://github.com/bcoin-org/bcoin/pull/939 from bcoin which fixes a bug introduced in https://github.com/bcoin-org/bcoin/pull/931

PR #334, opened 2020-01-24
getproof - dos prevention using sliding window counter

tuxcanfly:
An alternative approach to solving the dynamic DOS prevention for `getproof` p2p messages, using the Sliding Window Counter algorithim:

https://blog.cloudflare.com/counting-things-a-lot-of-different-things/

https://github.com/RussellLuo/slidingwindow

It is much simpler to reason about because you only need to keep track of current and previous windows.

Assume that we need to limit to 100 requests per second, in the simplest case:

< 1s >
-------
| 100 |  
-------

all 100 requests fall in the same window, so no more can be allowed.

In the second case:

<1s> <0.5s>
------------
| 50 | 50 |
-----------

50 requests fall in the previous window, 50 in the first half of the current window (0.5s). Since the current window is not yet done, we need to appromixate a score with the current knowledge.

We calculate an approximate score as 50 * (1-0.5) + 50 = 75, so in the next 0.5s we allow 100-75 = 25 requests and no more. If the previous window were empty, we could have allowed 50 more, where as if it was completely full, we can not allow any more. This is close to the our goal of only allowing 100 requests per second on average.

### DIFF
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -29,6 +29,7 @@ const NetAddress = require('./netaddress');
 const Network = require('../protocol/network');
 const {BrontideStream} = require('./brontide');
 const AirdropProof = require('../primitives/airdropproof');
+const SlidingWindow = require('./slidingwindow');
 const services = common.services;
 const invTypes = InvItem.types;
 const packetTypes = packets.types;
@@ -152,6 +153,7 @@ class Peer extends EventEmitter {
     this.nameMap = new BufferMap();
     this.totalProofs = 0;
 
+    this.proofWindow = null;
     this.init();
   }
 
@@ -550,6 +552,11 @@ class Peer extends EventEmitter {
     this.invTimer = setInterval(() => {
       this.flushInv();
     }, Peer.INV_INTERVAL);
+
+    this.proofWindow = new SlidingWindow({
+      limit: this.options.maxProofRPS
+    });
+    this.proofWindow.start();
   }
 
   /**
@@ -962,6 +969,10 @@ class Peer extends EventEmitter {
     if (this.invTimer != null) {
       clearInterval(this.invTimer);
       this.invTimer = null;
+    }
+
+    if (this.proofWindow != null) {
+      this.proofWindow.stop();
     }
 
     if (this.stallTimer != null) {
@@ -1960,6 +1971,13 @@ class Peer extends EventEmitter {
       'Sending proof (%s).',
       this.fullname());
 
+    this.proofWindow.increase(1);
+    if (!this.proofWindow.allow()) {
+      this.logger.debug('proof: rate limit exceeded (%s).', this.hostname());
+      this.ban();
+      return;
+    }
+
     this.send(new packets.ProofPacket(root, key, proof));
   }
 
@@ -2147,6 +2165,7 @@ class PeerOptions {
     this.compact = false;
     this.headers = false;
     this.banScore = common.BAN_SCORE;
+    this.proofPRS = 100;
 
     this.nonces = null;
     this.type = null;
@@ -2232,6 +2251,11 @@ class PeerOptions {
     if (options.banScore != null) {
       assert(typeof options.banScore === 'number');
       this.banScore = options.banScore;
+    }
+
+    if (options.maxProofRPS != null) {
+      assert(typeof options.maxProofRPS === 'number');
+      this.maxProofRPS = options.maxProofRPS;
     }
 
     if (options.nonces != null) {
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -1973,7 +1973,7 @@ class Peer extends EventEmitter {
 
     this.proofWindow.increase(1);
     if (!this.proofWindow.allow()) {
-      this.logger.debug('proof: rate limit exceeded (%s).', this.hostname());
+      this.logger.debug('proof: rate limit exceeded (%s).', this.fullname());
       this.ban();
       return;
     }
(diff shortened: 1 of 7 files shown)


tuxcanfly:
Tested with a script: https://gist.github.com/tuxcanfly/40abdb79e9c6f7cbdfad70da2769bf2d

Prevents the DOS on `getproof` using the sliding window, much before the read stall on master.

test/test-window.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-env mocha */
```
tynes: Nit: all of the other tests have the `-test` suffix

tuxcanfly: Fixed.

test/window-test.js
```diff
@@ -0,0 +1,49 @@
...
+  return new Promise(resolve => setTimeout(resolve, time));
+}
+
+describe('SlidingWindow (Unit)', function() {
+  const window = new SlidingWindow({
+    window: 10,
+    limit: 100
+  });
+
+  beforeEach(() => {
```
tynes: Should these be `{before,after}Each` or just `before` and `after`? The tests pass with either. Calling `stop` will causes the `weight` calculation to be much higher when `score` is called

test/window-test.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it('should reset after window timeout', async () => {
+    let reset = false;
+
+    window.once('reset', () => {
+      reset = true;
+    });
+
+    await sleep(window.window);
```
tynes: I observe this test failing if this line is changed to:

```hs
await sleep(window.window - 1)
```

:+1:

tynes:
With a functional test, this would be in a really good place. After https://github.com/handshake-org/hsd/pull/331 is implemented, peers will no longer need keys but that PR still needs some [work/review](https://github.com/handshake-org/hsd/pull/331#issuecomment-577996611).

tuxcanfly:
@tynes made a couple of changes, if you're reviewing please make sure to checkout the latest.

lib/net/peer.js
```diff
@@ -496,6 +498,11 @@ class Peer extends EventEmitter {
     this.invTimer = setInterval(() => {
       this.flushInv();
     }, Peer.INV_INTERVAL);
+
```
tynes: It makes more sense to initialize the `SlidingWindow` here compared to the `constructor` so that its only initialized if there is a successful p2p handshake. :+1:

test/window-test.js
tuxcanfly: With `beforeEach` and `afterEach` the windows start over from the start of the test case, so any timing effects outside the test case are nullified.

test/window-test.js
tuxcanfly: :+1:

lib/net/peer.js
tuxcanfly: :+1:

lib/node/fullnode.js
```diff
@@ -93,6 +93,7 @@ class FullNode extends Node {
       identityKey: this.identityKey,
       maxOutbound: this.config.uint('max-outbound'),
       maxInbound: this.config.uint('max-inbound'),
+      proofRPS: this.config.uint('max-proof-rps'),
```
tynes: I think that this variable name should match the configuration name. It threw me off at first when I was trying to test this code because the rest of the variable names match the config names.

tuxcanfly: Fixed.

tynes:
@tuxcanfly Do you mind squashing these commits?

lib/net/peer.js
```diff
@@ -1960,6 +1971,13 @@ class Peer extends EventEmitter {
       'Sending proof (%s).',
       this.fullname());
 
+    this.proofWindow.increase(1);
+    if (!this.proofWindow.allow()) {
+      this.logger.debug('proof: rate limit exceeded (%s).', this.hostname());
```
tynes: This logger statement should use `realname` and a call to `fullname` is still in the code above

test/slidingwindow-test.js
```diff
@@ -0,0 +1,135 @@
...
+describe('SlidingWindow (Functional)', function() {
+  it('should rate limit getproof to max-proof-rps ', async () => {
+    const maxProofRPS = 20;
+
+    const one = new FullNode({
+      'memory': true,
+      'network': 'regtest',
+      'listen': true,
+      'max-proof-rps': maxProofRPS,
+      'host': '127.0.0.1',
```
tynes: I don't think we technically need to specify the hosts since now we are using the ports to differentiate.

test/peering-test.js
```diff
@@ -195,6 +200,14 @@ describe('Peering', function() {
     });
 
     it('should gossip blocks over standard and brontide', async () => {
+      const seen = {
```
tynes: Nit: removal of previous comment helping to clarify what this object is used for, since you can't see the whole test in a single screen.

test/slidingwindow-test.js
```diff
@@ -0,0 +1,135 @@
...
+      'only': [`${key}@127.0.0.1`] // one is using default ports.
+    });
+
+    await one.open();
+    await one.connect();
+
+    await two.open();
+    await two.connect();
+
+    await common.event(one.pool, 'peer open');
```
tynes: This might not actually catch the race condition.

test/slidingwindow-test.js
```diff
@@ -0,0 +1,135 @@
...
+        packets++;
+    });
+
+    let count = 0;
+    while (!seen) {
+      count++;
+      try {
+        await two.pool.resolve(hash) ;
+      } catch (e) {
+        break;
```
tynes: Maybe we should assert that this is the error type that we are expecting.

Issue #335, opened 2020-01-24
Multiple Locations for seeds


tynes:
Right now there are multiple locations that seed IPs are defined.

They exist inside of `lib/protocol/networks` and also `lib/network/seeds`. Both allow for seeds to be defined by network. The docstring in `lib/protocol/networks' says:

https://github.com/handshake-org/hsd/blob/e5afb04d7f0b401a3418fead5fd79a4e79f9de57/lib/protocol/networks.js#L45-L49

Now that brontide is off by default, we can use DNS seeders just like Bitcoin does without needing something like Bolt 10. We should probably remove the hardcoded IPs in the `networks` file and allow DNS servers to be placed there.

There is no `regtest` or `simnet` defined in `lib/network/seeds`.
https://github.com/handshake-org/hsd/blob/master/lib/net/seeds/index.js
https://github.com/handshake-org/hsd/blob/master/lib/protocol/networks.js#L717

This means we need to add `lib/net/seeds/{regtest,simnet}`

kilpatty:
Just to add onto this, could we open an issue (or have the discussion here) about order of hardcoded seeds vs dns seeds. Last I checked bitcoind favored DNS seeds before falling back to hardcoded IPs, whereas bcoin and hsd both favored hardcoded seeds first. 

I'm not sure if I'm informed enough to have an opinion either way, but if there is a legitimate case for either being first, perhaps we support a config option to specify which the user would prefer?

PR #336, opened 2020-01-25
seeds: separate dns seeds and ip seeds

tynes:
Since there are multiple locations where seeds are defined, this PR moves IP based seeds in `lib/net/seeds` so that `lib/protocol/networks` can be used for domain name based seeds.

Closes https://github.com/handshake-org/hsd/issues/335


tynes:
Hit 2 race conditions in the tests. Taking a note here so that these could be fixed in the future.

```
  1) DNS Servers
       Authoritative Resolver
         should return authenticated URI record:

      AssertionError [ERR_ASSERTION]: [Array] deepStrictEqual [Array]

      Expected values to be strictly deep-equal:
      
        [
          [Record] {
      ...
              algorithm: 13,
      +       expiration: 1611454392,
      +       inception: 1579831992,
      -       expiration: 1611454391,
      -       inception: 1579831991,
              keyTag: 60944,
      ...
          }
        ]
      
      assert.deepEqual(dns.answer, res.answer);
             ^

      at Context.it (/home/circleci/repo/test/dns-test.js:134:18)

  2) Wallet HTTP
       should mine to the primary/default wallet:

      Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

PR #337, opened 2020-01-26
bip9: update soft fork start times

tynes:
Updates the soft fork start times to `1581206400` or February 9th, 2020.

Don't trust, verify.
https://www.unixtimeconverter.io/1581206400

### DIFF
--- a/lib/protocol/networks.js
+++ b/lib/protocol/networks.js
@@ -415,7 +415,7 @@ main.deployments = {
   hardening: {
     name: 'hardening',
     bit: 0,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
     timeout: 1654041600, // June 1st, 2022
     threshold: -1,
     window: -1,
@@ -425,7 +425,7 @@ main.deployments = {
   rollover: {
     name: 'rollover',
     bit: 1,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
     timeout: 1654041600, // June 1st, 2022
     threshold: -1,
     window: -1,
@@ -661,7 +661,7 @@ testnet.deployments = {
   hardening: {
     name: 'hardening',
     bit: 0,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
     timeout: 1654041600, // June 1st, 2022
     threshold: -1,
     window: -1,
@@ -671,7 +671,7 @@ testnet.deployments = {
   rollover: {
     name: 'rollover',
     bit: 1,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
     timeout: 1654041600, // June 1st, 2022
     threshold: -1,
     window: -1,
@@ -812,7 +812,7 @@ regtest.deployments = {
   hardening: {
     name: 'hardening',
     bit: 0,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
     timeout: 1654041600, // June 1st, 2022
     threshold: -1,
     window: -1,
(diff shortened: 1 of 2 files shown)


boymanjor:
@pinheadmz sorry we discussed it, and this will supersede #332.

tynes:
This PR is an alternative to the proposal https://github.com/handshake-org/hsd/pull/332.

Its assumed that mining pools will be the majority hashrate in the future which will be able to vote on bip9 based soft forks. Miners based on `getblock` will automatically signal for any soft forks that have `STARTED` [0] because of how `hsd` works, this could be addressed in the future by updating `hs-miner` to use `getblocktemplate` or updating the `getblock` RPC in `hsd` so that it opts out by default.

[0] https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki#states

lib/protocol/networks.js
```diff
@@ -671,7 +671,7 @@ testnet.deployments = {
   rollover: {
     name: 'rollover',
     bit: 1,
-    startTime: 1559347200, // June 1st, 2019
+    startTime: 1581206400, // February 9th, 2020
```
pinheadmz: Assuming we launch a new testnet along with main net, it would make sense to deploy these soft forks _earlier_ on testnet. That's how Bitcoin Core rolls, seems like a good idea.

pinheadmz:
Feb 9 2020 seems super soon, isn't it? This means if we don't deploy a patch by then, both forks will be locked in by (like) March 8? That means no RSA1024 airdrops and no 2010-signed DNSSEC proofs, almost immediately after launch. Why bother allowing them at all?

Note that hstratum doesn't use any rpc calls it all - it generates work by directly calling [`node.miner.createBlock()`](https://github.com/HandshakeAlliance/hstratum/blob/8ace0c3cef6abceccace15147902fdb7f29bea4f/lib/stratum.js#L280-L285), which in turn uses `this.chain.computeBlockVersion(tip);` to compute the block version. Miner options can be set internally to hard-code a version number, but they are not exposed in the FullNode configuration.

If I understand correctly, the only way this patch DOESN'T enforce both soft forks *next month* is if a majority of miners switch to `getblocktemplate` and of the three mining implementations we have (hs-miner, hstratum, hsd) none have the ability to opt out.

pinheadmz:
I think the summary of "what happens in the future" for the two proposals is like this:

#332 When soft fork is required, deploy new version of hsd with "real" start times and `force: true`, majority of miners must upgrade their full node (how bitcoin deploys soft forks).

#337 Before main net launch, a majority of mining software must be configured to allow `getblocktemplate` "rules" options. When soft fork is required, miners must be notified to set their rules to include like so (example): `rpc getblocktemplate {"rules":["hardening"]}`.

tynes:
> This means if we don't deploy a patch by then, both forks will be locked in by (like) March 8?

How did you calculate this number? 

The threshold is `1916` on mainnet.

https://github.com/handshake-org/hsd/blob/673b71ebf7b7e8a8b40a556ba1c6064ebeb63cab/lib/protocol/networks.js#L407

144 blocks per day if 1 every ten minutes.

Assuming that every block votes yes, since that is what the software naively does now if `hsd` + `getblock` (`hs-miner`) or `hsd` + `hstratum` (`handyminer`) make up the majority of the hashrate then:

1916 / 144 = 13.3, so ~13 days after the start time would be when it locks in if all blocks signal.

February 9th + 13 days = February 22nd would be `LOCKED_IN`

The next block would call `chain.getDeployments` which is called for every block.
https://github.com/handshake-org/hsd/blob/673b71ebf7b7e8a8b40a556ba1c6064ebeb63cab/lib/blockchain/chain.js#L490

Since the state is `LOCKED_IN`, the next block would progress to `ACTIVE`
https://github.com/handshake-org/hsd/blob/673b71ebf7b7e8a8b40a556ba1c6064ebeb63cab/lib/blockchain/chain.js#L3158-L3159

Which means that the softfork would start on the next block after it is `LOCKED_IN`, which means February 22nd.

pinheadmz:
Eh, I rounded:
"two weeks at STARTED, two weeks at LOCKED_IN".

The threshold is 1916 but the period is the miner window is 2016. The stages change at the edge of each miner window, not immediately after the threshold has been reached.

boymanjor:
This was superseded by d32e0178d2fe252318ce3289d76d13bf6f2b3b9e.

PR #338, opened 2020-01-26
pool: fix refill peers

tuxcanfly:
After the recent standard + brontide update, the `Refilling peers` message was broken. Investigated and found one more usage where `peers.outbound` was being used. Fixed.


tynes:
Good find, nothing else shows up when grepping for `peers.inbound` or `peers.outbound`.

The logic could also be updated, since it will only ever trigger refilling brontide outbound connections when it also needs to fill standard outbound connections.

PR #339, opened 2020-01-26
pool: refactor fill outbound logic

tynes:
Refactor of `Pool.fillOutbound` and `Pool.addOutbound`. This duplicates the original logic for adding peers but now for both standard and brontide connections. The random element of sometimes not attempting to connect to brontide peers was removed, as the safety from not dos'ing yourself should come from a low outbound count, not waiting longer before trying to connect to peers. This way a node can quickly connect to another brontide based peer quickly if one drops, which is important for spv nodes doing name resolution. It is also important for nodes that only connect to brontide based peers.

### DIFF
--- a/lib/net/pool.js
+++ b/lib/net/pool.js
@@ -1167,7 +1167,7 @@ class Pool extends EventEmitter {
 
   createOutbound(addr) {
     const options = new PeerOptions(this.options);
-    options.type = getTypeByAddr(addr);
+    options.type = Peer.getTypeByAddr(addr);
 
     const peer = Peer.fromOutbound(options, addr);
 
@@ -3419,6 +3419,8 @@ class Pool extends EventEmitter {
     if (typeof type !== 'number')
       type = peerTypes.STANDARD;
 
+    assert(Peer.isKnownPeerType(type));
+
     // Hang back if we don't
     // have a loader peer yet.
     if (!this.peers.load)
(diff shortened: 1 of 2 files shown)


lib/net/pool.js
```diff
@@ -3456,20 +3442,29 @@ class Pool extends EventEmitter {
    */
```
tynes: The call to `addLoader` could increase the number of outbound peers, so calculating `need` before `addLoader` can cause you to overestimate the number of peers that you need and cause you to connect to an extra peer.

lib/net/pool.js
```diff
@@ -3407,31 +3407,17 @@ class Pool extends EventEmitter {
...
-    // Try to connect to a brontide peer 20% of
-    // the time when a the outbound brontide peers
-    // are not full.
-    if (type == null) {
-      if (needBrontideOutbound && random(10) < 2)
-        type = peerTypes.BRONTIDE;
-      else
-        type = peerTypes.STANDARD;
-    }
+    if (typeof type !== 'number')
```
boymanjor: Perhaps we should assert that type is one of the allowed `peerTypes`?

lib/net/pool.js
```diff
@@ -3456,20 +3442,29 @@ class Pool extends EventEmitter {
...
-
     if (!this.peers.load)
       this.addLoader();
 
-    if (need <= 0)
+    const {maxOutbound, brontideMaxOutbound} = this.options;
+    const need = maxOutbound - this.peers.standard.outbound;
+    const brontideNeed = brontideMaxOutbound - this.peers.brontide.outbound;
+
+    if (need <= 0 && brontideNeed <= 0)
```
boymanjor: We should check each case in a separate if statement. Otherwise, we may print an unnecessary log statement, e.g. `need === 0`, but `brontideNeed === 1`.

tynes: Added the `if` statements below.

PR #340, opened 2020-01-27
dns: implement new resource serialization.

chjj:
This should be much simpler and more akin to DNS. IP compression has been removed -- the only compression that remains is label compression.

Record types:

- DS
- NS
- GLUE4
- GLUE6
- SYNTH4
- SYNTH6
- TXT
- A
- AAAA

Glue records must point to a child subdomain to avoid cache poisoning. We currently support A and AAA records, but we could consider removing these. 

A few tests were removed and could possibly be rewritten.

Opening for discussion.


lib/dns/resource.js
```diff
@@ -83,369 +78,76 @@ const DUMMY = Buffer.alloc(0);
...
-      extra.compress(c);
+    return false;
+  }
 
-    return c;
+  hasDS() {
+    return this.hasType(hsTypes.DS);
   }
 
   encode() {
```
tynes: We could use `MAX_RESOURCE_SIZE` here to replace the magic number `512`.
https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/covenants/rules.js#L107

Not part of this PR, but it could also be used here:
https://github.com/handshake-org/hsd/blob/8252c97e890148c56903c3178c9d3869318a285e/lib/covenants/namestate.js#L838

tynes:
In the `RootCache`, the `get` method returns null after `item.time + 6 * 60 * 60 * 1000`. Are the numbers supposed to correspond to about the time of a tree interval?

https://github.com/handshake-org/hsd/blob/8252c97e890148c56903c3178c9d3869318a285e/lib/dns/server.js#L84-L92

If so:
- Could network constants be used to make the behavior safe between networks?
- Would it be safe if the item was added to the cache right before a tree interval?

tynes:
> We currently support A and AAA records, but we could consider removing these.

I'm not strongly opinionated on this topic because I see the argument for both perspectives. Allowing for `A` and `AAAA` would help with decentralization because it would allow non-developers to participate in hosting a website more easily. Otherwise, they would need to rely more on third parties or they may not even participate.

It would also save money on cloud costs for running a nameserver (since running it at your home would dox you, we don't want a repeat of ENS). This is could lead to *far* more websites being available through handshake.

lib/dns/resource.js
```diff
@@ -83,369 +78,76 @@ const DUMMY = Buffer.alloc(0);
...
-    const c = this.compress();
-    const size = this.getSize(c);
-    const bw = bio.write(size);
-    this.write(bw, c);
+    const bw = bio.write(512);
```
tynes: We could use `MAX_RESOURCE_SIZE` here to replace the magic number `512`.
https://github.com/handshake-org/hsd/blob/af86d4899791601d10212115c80a2f2136fdb004/lib/covenants/rules.js#L107

Not part of this PR, but it could also be used here:
https://github.com/handshake-org/hsd/blob/8252c97e890148c56903c3178c9d3869318a285e/lib/covenants/namestate.js#L838

smcki012:
FWIW: I’m all for @tynes statements in regards to A/AAAA records. The complexity of running a nameserver is a barrier to entry to avoid if we’re looking to ensure on-ramping and experimentation is as easy as possible. This would also help aid in avoiding early centralization around firms offering that service for users; which also diminishes the likelihood of those individuals running a node of their own. 

Applications, like Bob wallet, make desktop name management easier as well, and the additional support means a user can buy a name and get a site hoisted in mostly one place. A solid early on-ramping UX, which also aids in promoting a healthy fullnode count.

chjj:
I see a huge problem with A/AAAA records to be honest. If we have them, we might as well have MX records too. If we have MX records, we might as well have SRV records. If we have SRV records, we might as well have every DNS record in the root zone. I don't think they belong there. It's a _very_ poor substitute for a name server. No one would ever use it in production for anything serious. I think we should remove them.

boymanjor:
>     * Could network constants be used to make the behavior safe between networks?

Both `mainnet` and `testnet` share the same treeInterval (~6 hours). I think this is fine, but I would be in favor of shorter TTLs on `regtest` and `simnet` deployments.

>     * Would it be safe if the item was added to the cache right before a tree interval?

I think this speaks to the general problem of stale cache data. However, we already have measures in place to mitigate cache poisoning.

tynes:
> I see a huge problem with A/AAAA records to be honest. If we have them, we might as well have MX records too. If we have MX records, we might as well have SRV records. If we have SRV records, we might as well have every DNS record in the root zone. I don't think they belong there. It's a _very_ poor substitute for a name server. No one would ever use it in production for anything serious. I think we should remove them.

There is nothing stopping somebody from adding their own serialization format that supports A/AAAA records since there are plenty of undefined serialization types. The market will decide what sorts of records will be stored in the chain. If people find utility in A/AAAA records on chain, they will put them there *no matter you think*.

That said, I don't really care if they are in or not, I just think that having them in to start could help make the community more accessible to non-technical people.

chjj:
@tynes, it's a matter of what the resolver will support. People are perfectly free to come up with their own serialization. There's no reason for resolvers to support A/AAA.

chjj:
Also, I think it's nearly impossible to make the argument for A/AAAA records without also making the argument for MX records. Some people want an easy-to-setup web server, sure. But others may want an easy-to-setup mail server. Where does it end? The barrier to entry argument doesn't hold up at all when you consider this.

tynes:
@chjj I'd assume that somebody who implemented their own serialization could also update the resolver to support alternative record types. That would bifurcate the network since not all nodes would be able to resolve all names.

> The barrier to entry argument doesn't hold up at all when you consider this.

I agree with this, mailservers are common enough although an average non-technical user wouldn't be as interested in running their own mailserver compared to running their own website. Blockchains are cool and all but without users they will stay as this weird niche thing forever.

> It would also save money on cloud costs for running a nameserver

I may only end up running a couple of nameservers because I don't want to pay Google monthly to run them, especially when the sites are not high traffic.

tynes:
Without A/AAAA records on chain, I think that a GUI wrapping `bns` with a simple 1-click deploy would be very important for maintaining the decentralization of the network. Otherwise exchanges/registrars will be managing nearly all non-technical users keys. DNSSEC tools need to become commoditized and simple to use.

tynes:
@chjj from https://github.com/handshake-org/hsd/issues/125

> The final incompatibility between us and (for lack of a better word) stupid DNS clients is that _ab7qaaab._hns is not actually resolvable on ICANN's system. For whatever reason, certain clients may fallback to their default root hints when resolving HNS nameservers. We can account for this by adding a --synthetic-nameserver switch to HSD and HNSD. Optionally, a user can pass --synthetic-nameserver=hns-ip-translator.example.com., changing _ab7qaaab._hns to _ab7qaaab.hns-ip-translator.example.com.. A public translator on ICANN's system can account for these final compatability errors. This name server would decode the lowest label as base32, and serve either an A or AAAA record in response. This is centralized, but when resolving HNS names with proper software, nobody should ever actually hit this centralized service. It's merely there so buggy DNS clients work with HNS.

If the synthetic nameserver is open source and anybody can deploy one for themselves or for their users, it doesn't seem too centralized. I doubt that ICANN would censor these services.

Do you think the synthetic nameserver should live in its own repo?
Should we leave a note someplace that this still needs to be implemented?

pinheadmz:
I wonder if this PR should also prevent user grief when sending `UPDATE`s. I know I've messed this up a lot on testnet/simnet by misspelling a field and wasting miner fees.

Something like https://github.com/handshake-org/hsd/pull/101 might help, possibly also with a "Unknown type" error if the user makes a typo in `rpc sendupdate`.

boymanjor:
This code was merged at a72bab62a019e97488f8d5248dbb983649532823.

PR #344, opened 2020-01-27
genesis: regenerate

tynes:
Regenerate the genesis block. Deprecates https://github.com/handshake-org/hsd/pull/307


pinheadmz:
Confirmed the output of the script matches the commit. But I'm kinda bewildered - what has changed since #307 ?

Also noticed there is a circular dependency here: `scripts/genesis.js` requires `lib/protocol/networks.js` which requires `lib/protocol/genesis.js`... which is the file generated by the script! It doesn't actually need anything from that file for this process though, so to confirm the script output I just commented out `lib/protocol/genesis.js` to avoid the missing module error.

Generally speaking, wouldn't we also expect this process to be literally the last commit before releasing version 1 of `hsd` ?

kilpatty:
Yeah following up to @pinheadmz unless this is a testnet/simnet genesis regeneration, wouldn't this allow for miners to start mining the "main net" genesis early?

pinheadmz:
@chjj I was discussing with @kilpatty some interest in launching testnet a bit earlier than main net. What do you think about picking a BTC block hash from today and committing "THE" testnet genesis block?

boymanjor:
This is not meant to be the genesis block for mainnet. I'm closing this and #307 to reduce confusion.

PR #346, opened 2020-01-27
softforks: opt in, by default no, vote with config

tynes:
This PR is meant to make soft fork activation on the network opt in by default. The full nodes will currently opt in by default, which is potentially risky behavior as it takes power away from users.

Two new configuration flags are defined:

- `--bip9-hardening`
- `--bip9-rollover`

These are set as booleans and are `null` (falsy) by default.

It is also meant to alleviate any concerns with https://github.com/handshake-org/hsd/pull/337 in regards to the soft forks activating too early. Now miners will be able to restart their software with a new configuration flag to signal for the softforks when using the `getwork` RPC or `hstratum`.

Will follow up with tests if this approach is concept ack'd

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -3187,11 +3187,12 @@ class Chain extends AsyncEmitter {
 
     for (const deployment of this.network.deploys) {
       const state = await this.getState(prev, deployment);
+      const accept = this.options.bip9[deployment.name];
 
-      if (state === thresholdStates.LOCKED_IN
-          || state === thresholdStates.STARTED) {
+      if (state === thresholdStates.LOCKED_IN)
         version |= 1 << deployment.bit;
-      }
+      else if (state === thresholdStates.STARTED && accept)
+        version |= 1 << deployment.bit;
     }
 
     version >>>= 0;
@@ -3402,6 +3403,7 @@ class ChainOptions {
     this.indexTX = false;
     this.indexAddress = false;
     this.forceFlags = false;
+    this.bip9 = {};
 
     this.entryCache = 5000;
     this.maxOrphans = 20;
@@ -3510,6 +3512,11 @@ class ChainOptions {
     if (options.checkpoints != null) {
       assert(typeof options.checkpoints === 'boolean');
       this.checkpoints = options.checkpoints;
+    }
+
+    if (options.bip9 != null) {
+      assert(typeof options.bip9 === 'object');
+      this.bip9 = options.bip9;
     }
 
     if (this.spv || this.memory)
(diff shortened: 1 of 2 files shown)


pinheadmz:
I like this approach better than #337 but still have some concerns:

This means when KSK2010 gets rolled over or RSA1024 breaks, the miners will have to know about it, and know what to do, and do it. The nice thing about #332 is that all that is rolled into one thing: "upgrade your full node", which is something we _might_ be able to rely on miners knowing how to do.

There is also some technical mess about how this will interact with `force: true` in `rpc getblocktemplate`. If we leave `force: false` then `getblocktemplate` clients will STILL have to add the rule name to their rpc call argument, or the template will not set the signal bit. This could use a test to confirm, but I think setting the soft forks to `force: true` combined with the options in this PR should work, it's just a bit confusing compared to other bitcoin soft fork deployments. The `force` logic comes in to play here, in `rpc getblocktemplate`:

https://github.com/handshake-org/hsd/blob/8252c97e890148c56903c3178c9d3869318a285e/lib/node/rpc.js#L1449-L1455

In prose:
- Set the bits for all deployments (happens earlier when `miner.getBlock()` calls `chain.computeBlockVersion()` )
- If the deployment is `force: false` AND the client has not mentioned the rule, UN-set the bit
- If `force: true` the bit will always NOT be UN-set 🙄

tynes:
@pinheadmz This is meant to supplement #337 

To be honest, I think https://github.com/handshake-org/hsd/pull/332 is a hack. Without intervention, the network is *not* safe from the problems the soft forks are meant to address. This will create the need to coordinate to release new software, which we should minimize those types of moments unless it makes sense otherwise.

> This means when KSK2010 gets rolled over or RSA1024 breaks, the miners will have to know about it, and know what to do, and do it. The nice thing about #332 is that all that is rolled into one thing: "upgrade your full node", which is something we might be able to rely on miners knowing how to do.

You can achieve the exact same functionality with this PR by hardcoding in `config.bip9.hardening: true` and releasing that software.

> There is also some technical mess about how this will interact with force: true in rpc getblocktemplate.

We can add an extra argument to `chain.computeBlockVersion` called `force` or something like that which would override the config option and revert the function to original behavior.

I need to closely read the codepath to make sure this will work, but I think it should.

pinheadmz:
> You can achieve the exact same functionality with this PR by hardcoding in config.bip9.hardening: true and releasing that software.

Great point, OK I'm on board :-)

tynes:
https://github.com/handshake-org/hsd/pull/347 is an alternative to this PR.

boymanjor:
This PR was supplemental to #337.

PR #350, opened 2020-01-30
test: increase rpc, http timeouts

tuxcanfly:
Tests on circle ci were failing because of node client timeout as well as http test timeout. Fixed.


pinheadmz:
This may indicate a bigger issue, I noticed `hsd-rpc` commands on the CLI responding somewhat sluggishly compared to last week. Just generating blocks, running through the auction steps, and checking `getnameresource`. A few calls even timed out, it's hard to reproduce though.

PR #363, opened 2020-02-06
wallet-http: return 404 in getname

### LINKED ISSUE #356: wallet http name endpoints fail if name is unknown
Error message here could be cleaner:

```
$ curl x:api-key@127.0.0.1:14039/wallet/primary/name/fakenamethatdoesnotexist

{
  "error": {
    "message": "Cannot read property 'getJSON' of null"
  }
}
```

I think we can add some `if (!ns) {...}` checks in `wallet/http.js` in https://github.com/handshake-org/hsd/pull/163

ping @tynes

pinheadmz:
Closes #356

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -811,7 +811,10 @@ class HTTP extends Server {
       const network = this.network;
       const ns = await req.wallet.getNameStateByName(name);
 
-      res.json(200, ns.getJSON(height, network));
+      if (!ns)
+        return res.json(404);
+
+      return res.json(200, ns.getJSON(height, network));
     });
 
     // Wallet Auctions


lib/wallet/http.js
```diff
@@ -811,7 +811,10 @@ class HTTP extends Server {
       const network = this.network;
       const ns = await req.wallet.getNameStateByName(name);
 
-      res.json(200, ns.getJSON(height, network));
+      if (!ns)
+        return res.json(404);
```
boymanjor: @pinheadmz the code base typically calls `res.json()` then calls `return`.

lib/wallet/http.js
```diff
@@ -811,7 +811,10 @@ class HTTP extends Server {
...
+
+      return res.json(200, ns.getJSON(height, network));
```
boymanjor: ditto

lib/wallet/http.js
pinheadmz: that's fine -- the error is `ns.getJSON` when `ns` is `null`

pinheadmz: or do you mean you prefer an if/else instead of a return?

boymanjor:
```js
if (!ns) {
  res.json(404);
  return;
}
```

pinheadmz: hm, I think this is the convention though:

https://github.com/pinheadmz/hsd/blob/2ef54a503db4494270a190eed8910d8bc531d8ca/lib/wallet/http.js#L1142

although, I think I should change the error code to `400` not `404`

boymanjor: I guess there isn't a convention lol: https://github.com/handshake-org/hsd/blob/2ef54a503db4494270a190eed8910d8bc531d8ca/lib/wallet/http.js#L218

boymanjor: > although, I think I should change the error code to 400 not 404

Agreed.

pinheadmz: ha! ok how about I just change the error code to `400` and we 🍻

boymanjor: Discussed out of band. We're going to leave this as-is for now.

PR #374, opened 2020-02-16
add BIP9 signaling statistics to rpc getblockchaininfo

pinheadmz:
Port of https://github.com/bcoin-org/bcoin/pull/724
which is a port of bitcoin/bitcoin#9571

Test is ported from https://github.com/handshake-org/hsd/pull/332

adds a subsection to soft fork deployments indicating amount of miner signaling and a boolean if activation is "still possible" this period, given the signaling thus far and the time remaining in the window:

```
"statistics": {
      "period": 2016,
      "threshold": 1916,
      "elapsed": 377,
      "count": 290,
      "possible": true
    }
```

- [x] TODO: fix test to cover RPC call

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -3162,6 +3162,46 @@ class Chain extends AsyncEmitter {
   }
 
   /**
+   * Get signalling statistics for BIP9/versionbits soft fork
+   * @param {ChainEntry} prev - Previous chain entry.
+   * @param {String} id - Deployment id.
+   * @returns {Promise} - Returns JSON object.
+   */
+
+  async getBIP9Stats(prev, deployment) {
+    const bit = deployment.bit;
+    let window = this.network.minerWindow;
+    let threshold = this.network.activationThreshold;
+
+    // Deployments like `segsignal` (BIP91) have custom window & threshold
+    if (deployment.window !== -1)
+      window = deployment.window;
+
+    if (deployment.threshold !== -1)
+      threshold = deployment.threshold;
+
+    let count = 0;
+    let block = prev;
+
+    while((block.height + 1) % window !== 0) {
+      if (block.hasBit(bit))
+        count++;
+
+      block = await this.getPrevious(block);
+      if(!block)
+        break;
+    }
+
+    return {
+      period: window,
+      threshold: threshold,
+      elapsed: (prev.height + 1) % window,
+      count: count,
+      possible: (window - threshold) >= ((prev.height + 1) % window) - count
+    };
+  }
+
+  /**
    * Compute the version for a new block (BIP9: versionbits).
    * @see https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki
    * @param {ChainEntry} prev - Previous chain entry (usually the tip).
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -3169,6 +3169,10 @@ class Chain extends AsyncEmitter {
    */
 
   async getBIP9Stats(prev, deployment) {
+    const state = await this.getState(prev, deployment);
+    if (state !== thresholdStates.STARTED)
+      throw new Error(`Deployment "${deployment.name}" not in STARTED state.`);
+
     const bit = deployment.bit;
     let window = this.network.minerWindow;
     let threshold = this.network.activationThreshold;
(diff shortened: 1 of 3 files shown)


lib/blockchain/chain.js
```diff
@@ -3161,6 +3161,45 @@ class Chain extends AsyncEmitter {
     return state;
   }
 
+  /**
+   * Get signalling statistics for BIP9/versionbits soft fork
+   * @param {ChainEntry} prev - Previous chain entry.
+   * @param {String} id - Deployment id.
```
boymanjor: s/id/deployment/

boymanjor: This whole line needs to be updated.

lib/blockchain/chain.js
```diff
@@ -3161,6 +3161,45 @@ class Chain extends AsyncEmitter {
...
+   * @returns {Promise} - Returns JSON object.
+   */
+
+  async getBIP9Stats(prev, deployment) {
```
boymanjor: What is the expected behavior of this function when the soft fork is not in the `STARTED` state? Since this is a library function, I think we should handle all possible inputs.

pinheadmz: Hm, what do you suggest? Outside of `STARTED` the statistics are totally meaningless, but they would still be accurate (reporting the number of blocks with a certain bit set). The exception would be `possible` since activations are only possible or not during the `STARTED` phase.

We could make another call to `getState()` and assert that we are `STARTED` ?

chjj: I think throwing an error would be fine.

pinheadmz: Added a commit to catch this state and throw. I'm not sure if this is the best method, since it calls `getState()` again, meaning this:

`while((block.height + 1) % window !== 0)`

is run three times for each `rpc getblockchaininfo` call (to get the state in the rpc module, to get it again in the chain module, and then finally to actually count the statistics(

lib/blockchain/chain.js
boymanjor: @pinheadmz bump.

lib/blockchain/chain.js
boymanjor: @pinheadmz where does `getState` run the while loop you are describing?

boymanjor: Regardless, the while loop is bounded by `window`, so I'm not sure that it will make a big difference. If it does we can pass `state` to `getBIP9Stats`.

lib/blockchain/chain.js
pinheadmz: sorry / thanks -- pushed commit.

lib/blockchain/chain.js
pinheadmz: I mean, `getState` itself iterates through up to 2016 to check the state. That will happen twice, and then `getBIP9Stats` does it again to actually count the signalling bits .

lib/blockchain/chain.js
```diff
@@ -3161,6 +3161,49 @@ class Chain extends AsyncEmitter {
...
+    if (state !== thresholdStates.STARTED)
+      throw new Error(`Deployment "${deployment.name}" not in STARTED state.`);
+
+    const bit = deployment.bit;
+    let window = this.network.minerWindow;
+    let threshold = this.network.activationThreshold;
+
+    // Deployments like `segsignal` (BIP91) have custom window & threshold
+    if (deployment.window !== -1)
+      window = deployment.window;
```
boymanjor: nit: add a new line.

pinheadmz: fixed, thanks

test/bip9-chain-test.js
```diff
@@ -0,0 +1,151 @@
...
+    assert.strictEqual(state1, thresholdStates.DEFINED);
+    await addBIP9Blocks(1, true);
+    const state2 = await getHardeningState(chain.tip);
+    assert.strictEqual(state2, thresholdStates.STARTED);
+    assert(!await hasHardening());
+  });
+
+  it('should add blocks: does not reach LOCKED_IN', async () => {
+    // Not enough miner support signaled.
+    await addBIP9Blocks(minerWindow - activationThreshold, true);
```
boymanjor: I think this test would be more robust if we tested the boundaries of activation, i.e. `minerWindow - activationThreshold + 1` blocks do _not_ signal support.

pinheadmz: thanks for you patience on this one, I think the latest commit should satisfy

PR #382, opened 2020-03-03
test: use 192.0.2.1 as public host; fix tests

tuxcanfly:
Use a test IP address from: https://tools.ietf.org/html/rfc5737


tuxcanfly:
@boymanjor I guess that's OK since it's just test.

Issue #381, opened 2020-03-02
Assertion failed when calling sendredeem or sendupdate on imported wallet


buffrr:
I'm getting `Assertion failed` when I try to call `sendredeem` on lost bids or when calling `sendupdate` on a domain that I won after importing wallet.

Steps to reproduce:

1. Install a fresh hsd node & run it.
2. Import wallet with `mkwallet` using mnemonic phrase.
3. Attempt to redeem lost bids using `sendredeem` or call `sendupdate` to update any record on a TLD.

Error message: `Assertion failed`

Both commands show assertion failed. Following suggestion by @pinheadmz  I replaced [blgr](https://github.com/bcoin-org/blgr) in hsd with the master branch to get stack traces.

Logs output:

    Assertion failed.
       at Balance.applyTo (/home/aa/hsd/lib/wallet/txdb.js:3123:5)
       at TXDB.updateAccountBalance (/home/aa/hsd/lib/wallet/txdb.js:214:11)
       at <anonymous>

pinheadmz:
Did you rescan after importing the mnemonic phrase? Did you make bids from an different wallet / node?

buffrr:
I did call `rescan 0` after importing it. The wallet already had existing bids from a different node. I didn't make bids from other wallets. This was the only wallet used.

mslipper:
When viewing this wallet's balances, there are a bunch of locked coins. For example:

```
{
  "network": "main",
  "wid": 1,
  "id": "allison",
  "watchOnly": false,
  "accountDepth": 1,
  "token": "",
  "tokenDepth": 0,
  "master": {
  },
  "balance": {
    "account": -1,
    "tx": 16,
    "coin": 15,
    "unconfirmed": 100000,
    "confirmed": 100000,
    "lockedUnconfirmed": 0,
    "lockedConfirmed": 10000
  }
}
```

The line in HSD that is failing is this one:

https://github.com/handshake-org/hsd/blob/master/lib/wallet/txdb.js#L3123

This leads me to believe that either the wrong set of outputs are being selected, or the `applyTo` algorithm is wrong.

Both cases I've seen have had enough money to pay for the update.

mslipper:
Also - these wallets have been rescanned from zero and nonces have been imported.

mslipper:
I did some further digging, and had a hacked HSD node output the value of `delta` and `balance` in the `TXDB.updateAccountBalance` method. Here's the output:

```
[...]

PR #385, opened 2020-03-08
wallet: always apply balance deltas to lockedUnconfirmed balance

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/381

Explanation here: https://github.com/handshake-org/hsd/issues/381#issuecomment-596240003

TODO:
- [ ] write unit test.

### DIFF
--- a/lib/wallet/txdb.js
+++ b/lib/wallet/txdb.js
@@ -1643,9 +1643,8 @@ class TXDB {
     switch (covenant.type) {
       case types.CLAIM:
       case types.BID: {
-        if (height === -1)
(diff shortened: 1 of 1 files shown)


pinheadmz:
@boymanjor yah standby indeed, this patch is wrong. I'm counting unconfirmed TXs twice. I think the approach needs to be more like https://github.com/handshake-org/hsd/pull/262

pinheadmz:
this approach was totally wrong, new PR is here: https://github.com/handshake-org/hsd/pull/387

PR #387, opened 2020-03-10
wallet: update unconfirmed locked balance first during rescan

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/381

Replaces https://github.com/handshake-org/hsd/pull/385

Explanation here: https://github.com/handshake-org/hsd/issues/381#issuecomment-596240003

The test runs through the REVEAL of an auction two ways: First, by sending and then confirming each covenant. Second, by JUST confirming each covenant without first "sending". The difference from perspective of the txdb is that if a TX is not already inserted in an unconfirmed state, the balances will be wonky when the TX is confirmed.

Like #387 the patch is to first quickly update the "unconfirmed" data for a TX before proceeding (in the case where txdb sees a TX for the very first time, in a block)

### DIFF
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -1,5 +1,6 @@
 /* eslint-env mocha */
 /* eslint prefer-arrow-callback: "off" */
+/* eslint no-implicit-coercion: "off" */
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


test/wallet-test.js
```diff
@@ -1642,4 +1643,298 @@ describe('Wallet', function() {
...
+      wallet = await wdb.create();
+      // rollout all names
+      wdb.height = 52 * 144 * 7;
+    });
+
+    after(async () => {
+      await wdb.close();
+    });
+
+    it('Fund wallet', async () => {
```
nijynot: Most tests in `test/*.js` seems to follow a convention of starting with the word "should", maybe do something like this for all test cases in this PR?
```js
it('should fund wallet', async () => { ... })
```

pinheadmz: Thanks! Truth is, I'm not a huge fan of the "should" convention and I'm always trying to break it, especially in a case like this where the entire `Describe` block is testing one thing, and the `it` blocks are just like incremental steps :-)

But, I'm so happy you reviewed my PR I'll take your suggestion!

test/wallet-test.js
```diff
@@ -1642,4 +1643,298 @@ describe('Wallet', function() {
...
+      assert.strictEqual(bal.tx, 4);
+      assert.strictEqual(bal.coin, 4);
+      assert.strictEqual(bal.confirmed, 10e6 - (3 * fee));
+      assert.strictEqual(bal.unconfirmed, 10e6 - (3 * fee));
+      assert.strictEqual(bal.ulocked, value);
+      assert.strictEqual(bal.clocked, value);
+    });
+  });
+
+  describe('TXDB locked balance - RESCAN', function() {
```
boymanjor: nit: 'TXDB locked balance after simulated rescan" is a bit more descriptive.

pinheadmz:
@boymanjor thanks bruh. Updated those comments and pushed.

PR #392, opened 2020-03-16
Make export TransactionOptions happen

wi-ski:
I need this for a plugin

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -1693,5 +1693,5 @@ function enforce(value, msg) {
 /*
  * Expose
  */
-
+HTTP.TransactionOptions = TransactionOptions;
 module.exports = HTTP;


wi-ski:
@tynes -Is this awful? Can i have this?

pinheadmz:
Can you expand on your use case a bit? The options are internal to the http module.

wi-ski:
Ah - excuse me. I was in a headspace that made me feel like this would be an intended pattern.

Im building a plugin and would like to add a new route.  Im copy-pasting a handler (in essence) and it _would_ makes use of this class - I had naively assumed it would be exported from somewhere.

nijynot:
Probably worth creating a separate issue on how to handle exporting a class' options it's not possible to use both `module.exports` and `exports.*` at the same time unless you're using ES6 imports.
Don't know if adding the option class to the HTTP class is desireable.

tynes:
> Don't know if adding the option class to the HTTP class is desireable.

I don't see a particular problem with it as a similar pattern is done elsewhere in the codebase. Its also not an object that is created millions of times.

@wi-ski Could you detail the new route you are working on? You also have a linting error

pinheadmz:
I agree with @tynes 

Consider the style used here:

https://github.com/handshake-org/hsd/blob/98b33cc6e6911f5b136f99cd025ae9d380d03312/lib/primitives/mtx.js#L1923-L1932

Also yeah please add the `;` and please reword your commit message to match the style,
e.g. "wallet: expose transactionOptions in http"

wi-ski:
Updated :ok_hand:

wi-ski:
@tynes - Im porting the changes from this PR: https://github.com/handshake-org/hsd/pull/233/

Im setting up http handlers for stuff like `/update-for-account` and `/reveal-for-account`

lib/wallet/http.js
```diff
@@ -1693,5 +1693,5 @@ function enforce(value, msg) {
 /*
  * Expose
  */
-
+HTTP.TransactionOptions = TransactionOptions;
```
pinheadmz: nit: keep the empty newline after the comment.

pinheadmz:
Tested ok in REPL:

```
> const hsd = require('hsd')
undefined
> const Validator = require('bval')
undefined
> new hsd.wallet.HTTP.TransactionOptions(new Validator({rate: 123456}))
TransactionOptions {
  rate: 123456,
  maxFee: null,
  selection: null,
  smart: null,
  account: null,
  locktime: null,
  sort: null,
  subtractFee: null,
  subtractIndex: null,
  depth: null,
  paths: null,
  outputs: []
}
```

Just fix that newline nit when you get a chance and I'll 👍

lib/wallet/http.js
wi-ski:
```suggestion

HTTP.TransactionOptions = TransactionOptions;
```

wi-ski:
Isnt that something we can do when merging. Like squash and merge.

wi-ski:
@pinheadmz - Squashed :open_hands: Lemme know if there's anything else.

lib/wallet/http.js
pinheadmz: dude, sorry to be super annoying about this, but the newline should stay in :-)
I think when you squashed you might've removed the commit that added the newline.

PR #393, opened 2020-03-16
wallet: rpc signMessage updated with blake2b from sha256

realrasengan:


pinheadmz:
Thank you for contributing @realrasengan! Here's a bit of feedback on style when contributing to hsd:

1. Pull requests should have at least a short description. In this case, it's a pretty small change and the title covers it but just imagine someone new to hsd looking at this PR -- would they know what you are trying to change or why?

2. Commit messages should start with the module being updated, in this case the commit message should be something like:
"wallet: rpc signMessage updated with blake2b from hash256 "

3. Please add test coverage for all additions. [See the relevant test in bcoin here](https://github.com/bcoin-org/bcoin/blob/6beb54dddc1256cca379ad765b71fd5cdee427f3/test/wallet-rpc-test.js#L295-L339) You could probably copy most of that.

pinheadmz:
@nijynot

> hash256 is also ambigious and SHA256 or blake2b should be explicitly used instead, I think.

`hash256` is a label from bcrypto and represents the bitcoin-standard "double sha256". You can see it used throughout bcoin.

tynes:
It looks like the test failed due to a race condition. If somebody could retrigger the CI, that would be great.

Also @realrasengan good job signing your commit.

nijynot:
@pinheadmz Ahh, it's the double sha256 hash, gotcha. Thanks for clearing it up!
I'm fairly new to the codebase, so hope you all don't mind me making some mistakes here and there.

Good PR btw @realrasengan! :smile:

realrasengan:
Updated the title @pinheadmz and thank you @tynes and @nijynot

pinheadmz:
@realrasengan title is good, but we're hoping you can update the actual commit message. This can be done with a git amend or rebase: https://linuxize.com/post/change-git-commit-message/

Just for the sake of reviewing I went ahead and ported the test over from bcoin for you. After you fix the commit message you should be able to merge this PR-on-your-PR to add the test. Then we can merge the whole mess.

Thanks again for contributing!

realrasengan:
Thank you for the help @pinheadmz !  As per our conversation, these steps are now complete, please let me know if there is anything else I need to do and thanks again!

boymanjor:
>2. Commit messages should start with the module being updated, in this case the commit message should be something like:
"wallet: rpc signMessage updated with blake2b from hash256 "

This is not a reason to block this from merging, but I would modify this to also state that the message should be in the imperative tense, i.e. "wallet: update signmessage rpc with blake2b from hash256."

realrasengan:
Thank you @boymanjor !

Issue #394, opened 2020-03-16
Chain "value" (coin supply) seems innacurate


pinheadmz:
The coin supply chart at https://hnscan.com/charts/supply has indicated a recent tick-down in coin supply, which I'm investigating.

I wrote a script that iterates through the entire chain, [returning a total in/out/diff for each block and maintains a running grand total.](https://gist.github.com/pinheadmz/3ab8ac3dcd2d1ae6b7fc5e613ce043bb)

Observations:

1. no blocks returned a negative "diff" (miners are accumulating all available fees and no coins are being destroyed) 

2. compared to `getInfo` the chain state reports a totally different value:

### Script output
```
height:  6590 in:  2462599819738 out:  2464599819738 diff:     2000000000 grand total: 269816163166950
```

### ChainState
```
    state: {
      tx: 170307,
      coin: 108139,
      value: 262612642694412,
      burned: 1060028925848
    }
```

So according to chainstate, `7,203,520.472538 HNS` is somehow missing.
This isn't related to https://github.com/handshake-org/hsd/issues/380 but perhaps a PR to clean up both issues will be possible once this issue is found.

chjj:
This is a bug in the total supply calculation (this affects the UI only). We can fix this with a simple data migration. The supply is lower than it should be as a result of subtracting some outputs' values that it shouldn't be. I'll start working on a migration now and have it for the next release.

boymanjor:
Fixed by #396.

PR #396, opened 2020-03-17
chaindb: migrate and fix total supply calculation. see #394.

chjj:


lib/blockchain/layout.js
```diff
@@ -29,6 +29,7 @@ const bdb = require('bdb');
  *   w[height] -> name undo
  *   s -> tree state
  *   f -> bit field
+ *   M[migration-id] -> migration
```
pinheadmz: nit: should specify `dummy` ?

pinheadmz:
@kilpatty are you able to review this as well? Maybe run it with hnscan indexer?

kilpatty:
LGTM on mobile right now, but I’ll run a build of hnscan-backend with this in an hour or two and report back.

pinheadmz:
Ran both master and branch with this extra log:

```diff
diff --git a/lib/blockchain/chaindb.js b/lib/blockchain/chaindb.js
index 214b91bd..7f3198ce 100644
--- a/lib/blockchain/chaindb.js
+++ b/lib/blockchain/chaindb.js
@@ -1486,6 +1486,8 @@ class ChainDB {
     // Connect block and save data.
     await this.saveBlock(entry, block, view);
 
+    console.log(`height: ${entry.height} current: ${this.state.value} new: ${this.pending.value}    diff: ${this.pending.value - this.state.value}`)
+
     // Commit new chain state.
     this.put(layout.R.encode(), this.pending.commit(hash));
   }
```

Confirmed on master there were 218 blocks with negative value applied to chainstate and on branch, only 40.

I'm still kind of surprised there were any at all... what is destroying these coins?

chjj:
@pinheadmz, try this:

``` diff
diff --git a/lib/blockchain/chaindb.js b/lib/blockchain/chaindb.js
index 9e2c9af7..e1248742 100644
--- a/lib/blockchain/chaindb.js
+++ b/lib/blockchain/chaindb.js
@@ -1577,6 +1577,16 @@ class ChainDB {
     // Connect block and save data.
     await this.saveBlock(entry, block, view);

+    const added = this.pending.value - this.state.value;
+
+    if (added < 0) {
+      const burned = this.pending.burned - this.state.burned;
+
+      this.logger.info('height=%d current=%d new=%d added=%d burned=%d',
+                       entry.height, this.state.value, this.pending.value,
+                       added, burned);
+    }
+
     // Commit new chain state.
     this.put(layout.R.encode(), this.pending.commit(hash));
   }
```

In the negative blocks, you can clearly see that there are burned coins (REGISTERs).

Example:

```
[info] (chaindb) height=5773 current=266868346659591 new=266868345659591 added=-1000000 burned=2001000000
```

pinheadmz:
@chjj is there any chance a node could get interrupted during migration and be stuck in a weird state? i.e. Is the commitment to the new chainstate atomic along with the migration ID marker? (looks like it is a la `batch` just wanna confirm)

chjj:
@pinheadmz, it should be completely atomic.

Issue #378, opened 2020-02-22
Blind value not found when trying to send reveal


dylanbathurst:
I've been trying to send a reveal for a tld, initially from Bob wallet, but also from the cli on a completely separate hns full node and I keep getting the `Blind value not found` error. When I run `hsw-rpc getbids` i see the correct bid and I also see my bid when I run `hsw-rpc getauctioninfo coin` 

from @pinheadmz's request I first tried `hsw-rpc selectwallet mywallet` to make sure I was using the correct wallet.

pinheadmz:
Blinds are saved when the bid is made. In hsd wallet they are deterministically generated, so in the worst case you should be able to [regenerate the nonce](https://hsd-dev.org/api-docs/#importnonce).

How did you make the bid? And what is the relationship between Bob and your other full node? If you sent the bid from one, I wouldn't expect the other to know the blind without being precomputed.

dylanbathurst:
I see, I’ll try regenerating. I initially made the bid on bob but when I went to reveal I got that error, so I went over to my own had node on a separate computer and imported my seed to try it there, same error. I initially imported an existing seed into bob so I wonder if the issue with bob is that it’s not selecting my imported wallet before doing the reveal request...

dylanbathurst:
Is there a nonce that I can compare the output of `importnonce`?

pinheadmz:
Looks like the http endpoint (not rpc) has more verbose output:

https://hsd-dev.org/api-docs/#get-nonce-for-bid

You could compare the entire `blind` with the value in the BID covenant in the output of your BID transaction... is that what you mean?

ljm625:
well i was facing the same issue here... I tried hsw-rpc importnonce and it returns me a nonce. Then I retries 
hsw-rpc sendreveal
it still says Blind value not found.
I am sure the address and bid amount is correct. Anyone can shade some light on this? really appreciate

bemusementpark:
> In hsd wallet they are deterministically generated, so in the worst case you should be able to regenerate the nonce.

Should this be done automatically when 1. importing a wallet and/or 2. you call `reveal` (etc.) and there is no blind present?

pinheadmz:
@bemusementpark when importing a wallet from a mnemonic seed, there is no way to know the VALUE of the bid being blinded. Therefore a restored wallet can not regenerate the nonce without extra input from the user.

bemusementpark:
[...]

PR #402, opened 2020-03-24
Add new HTTP and RPC calls to get corresponding BlindValue from txdb for given blind

pinheadmz:
An imported wallet may "forget" the blinded value of a bid since it is not recoverable from the seed and blockchain data. This new rpc call queries the txdb for the preimage (BlindValue) for a hash (blind). It will be useful to inform the user that the bid value is missing before the user tries to sendreveal and ends up with a "Blind value not found" error. Wallet applications can test this rpc call and inform the user they need to run `importnonce`.

Related:
https://github.com/handshake-org/hsd/issues/378
https://github.com/kyokan/bob-wallet/issues/110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Wallet API changes
 
-- Adds new wallet rpc `getblindvalue` which accepts a blind (hash) and returns a
-`BlindValue` in JSON format if it is found in the txdb.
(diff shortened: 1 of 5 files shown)


pinheadmz:
Ok, I actually found a work around for this: if the wallet returns a bid where `own: true` but `value: undefined`, that means the BlindValue could not be found in the txdb. I'm happy to maintain this branch and revise if people are interested in reviewing.

PR #404, opened 2020-03-24
Specify the lockup period in README

rsolari:
Someone claiming their name was confused by the phrase "you must wait several blocks".  They thought claiming would take less than an hour. This copy is more clear.

For reference, the lockupPeriod is specified in networks.js:
https://github.com/handshake-org/hsd/blob/bbac8dd943767da62e46e52f93227b3da1048e17/lib/protocol/networks.js#L266

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ $ hsw-rpc sendclaim example
 
 This will create and broadcast the proof to all of your peers, ultimately
 ending up in a miner's mempool. Your claim should be mined within 5-20 minutes.
-Once mined, you must wait 43,200 blocks – about a 30 days – before your claim is considered
+Once the transaction is mined, you must wait about 30 days (43,200 blocks) before your claim is considered
 "mature".
 
 Once the claim has reached maturity, you are able to bypass the auction process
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -303,7 +303,7 @@ $ hsw-rpc sendclaim example
 
 This will create and broadcast the proof to all of your peers, ultimately
 ending up in a miner's mempool. Your claim should be mined within 5-20 minutes.
-Once mined, you must wait several blocks before your claim is considered
+Once mined, you must wait 43,200 blocks – about a 30 days – before your claim is considered
```
pinheadmz: nit "about 30 days"

pinheadmz:
Thanks Raf! Good lookin out.

README.md
tynes: nit: `you must wait about 30 days (43,200 blocks) before...` 
I think that reads better for humans

rsolari: Good points. Updated the copy.

README.md
```diff
@@ -303,7 +303,7 @@ $ hsw-rpc sendclaim example
 
 This will create and broadcast the proof to all of your peers, ultimately
 ending up in a miner's mempool. Your claim should be mined within 5-20 minutes.
-Once mined, you must wait several blocks before your claim is considered
+Once the transaction is mined, you must wait about 30 days (43,200 blocks) before your claim is considered
```
boymanjor: This number should be 4,320 blocks, i.e. 144 blocks/day * 30.

rsolari: Good catch. Fixed

PR #406, opened 2020-03-26
readme: update install instructions

tynes:
This updates the `README` with installation instructions based on user feedback. It splits things out into sections for different package managers and gives commands that can be easily copied and pasted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ __HSD__ is an implementation of the [Handshake][handshake] Protocol.
 
 `hsd` requires Node.js v10 or higher
 
+
+### Building From Source
+
 ```
 $ git clone git://github.com/handshake-org/hsd.git
 $ cd hsd
@@ -19,6 +22,36 @@ $ ./bin/hsd
 Note that `node-gyp` must be installed. See the
 [node-gyp](https://github.com/nodejs/node-gyp) documentation for more
 information.
+
+### npm
+
+It is not recommended to install `hsd` from npm's repositories
+but it is still possible to install with `npm` using a remote
+`git` repository.
+
+```
+$ npm install -g https://github.com/handshake-org/hsd.git
+```
+
+A `git` ref can be used to install a particular version by appending
+a `#` and the name of the `git` ref to the URL. For example,
+`https://github.com/handshake-org/hsd.git#v2.1.3`. It is recommended
+to use the [latest tagged release](https://github.com/handshake-org/hsd/releases).
+
+If adding `hsd` as a dependency to a project, use the command:
+
+```
+$ npm install https://github.com/handshake-org/hsd.git
+```
+
+### macOS
+
+`hsd` is available via [Homebrew](https://brew.sh). This will
+install all required dependencies as well as `unbound`.
+
+```
+$ brew install hsd
+```
 
 ## CLI
 


README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
 [node-gyp](https://github.com/nodejs/node-gyp) documentation for more
 information.
 
+### npm
+
+It is not recommended to install `hsd` from npm's repositories
+but it is still possible to install using `npm` using a remote
```
boymanjor: nit: "to install ~using~ _with_ `npm` using"

README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
...
+### macOS
+
+`hsd` is available via Homebrew. This will install all required
+dependencies as well as `unbound`.
+
+```
+$ brew install hsd
+```
+
+To install homebrew, see https://brew.sh
```
boymanjor: Let's remove this line and link to the homepage above, e.g.:

> `hsd` is available via [Homebrew](https://brew.sh).

README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
...
 
+### npm
+
+It is not recommended to install `hsd` from npm's repositories
+but it is still possible to install using `npm` using a remote
+`git` repository. To install a particular version, a git tag can
+be used. It is recommended to use the latest version.
+
+```
+$ npm install -g https://github.com/handshake-org/hsd.git#v2.1.3
```
pinheadmz: Hard-coding a version  number here means constant maintenance of the README. I'm not sure the exact syntax but it might be more convenient to use something like `semver^2.0.0`

README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
...
+```
+
+If adding `hsd` as a dependency to a project, use the command:
+
+```
+$ npm install https://github.com/handshake-org/hsd.git#v2.1.3
```
pinheadmz: include `--save` ?

pinheadmz:
As long as we're here, do you want to add some linux details as well?
Like https://github.com/pinheadmz/handout#1-install-the-necessary-dependencies-in-addition-to-nodejs

I use this command on all my cloud installs:

```
$ apt-get install build-essential python libunbound-dev
```

boymanjor:
@pinheadmz we should specify that these are instructions for Debian (and its derivatives), not Linux generally.

README.md
boymanjor: We could use whatever GitHub uses to represent refs in documentation.

README.md
boymanjor: @pinheadmz it depends on if we want to assume a reader understands how to use `npm`.

README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
 [node-gyp](https://github.com/nodejs/node-gyp) documentation for more
 information.
 
+### npm
+
+It is not recommended to install `hsd` from npm's repositories
+but it is still possible to install using `npm` using a remote
+`git` repository. To install a particular version, a git tag can
```
boymanjor: nit: technically any git ref can be used right?

README.md
```diff
@@ -20,6 +22,34 @@ Note that `node-gyp` must be installed. See the
...
+be used. It is recommended to use the latest version.
```
boymanjor: nit: latest _tagged_ release.

Assuming we mention that any git ref can be used, we can emphasize to use the latest tagged release.

README.md
tynes: `--save` is no longer required to add the the dependency to the `package.json`, I believe that is old `npm` behavior. You must pass `--no-save` to not add it.

```bash
$ npm --version
6.14.3

$ npm install --help

npm install (with no args, in package dir)
npm install [<@scope>/]<pkg>
npm install [<@scope>/]<pkg>@<tag>
npm install [<@scope>/]<pkg>@<version>
npm install [<@scope>/]<pkg>@<version range>
npm install <alias>@npm:<name>
npm install <folder>
npm install <tarball file>
npm install <tarball url>
npm install <git:// url>
npm install <github username>/<github project>

aliases: i, isntall, add
common options: [--save-prod|--save-dev|--save-optional] [--save-exact] [--no-save]
```

tynes:
> As long as we're here, do you want to add some linux details as well?
> Like https://github.com/pinheadmz/handout#1-install-the-necessary-dependencies-in-addition-to-nodejs
> 
> I use this command on all my cloud installs:
> 
> ```
> $ apt-get install build-essential python libunbound-dev
> ```

Ok if we add this once we have the Debian package set up? I'm in the process of doing that now.

tynes:
I pushed up fixes for the comments, this should be ready for another round of review.

PR #408, opened 2020-03-27
Wallet: enable importing names and "watch" for auction activity without bidding

pinheadmz:
This enables a user to import a name into their wallet in a similar way to importing an address or public key. A rescan is still required to recover any auction activity and in particular the OPEN for a name MUST be rescanned or the wallet will not record a `nameState` or track any bids.

The WalletDB global `name->wid` map is used as a "watch" marker to determine if a name's OPEN should be added to the txdb.

The new rpc command takes a rescan height parameter:

```
hsw-rpc importname sad 2016
```

Wallet applications should check the state of a name and use a name's `height - 1` as the rescan height. Assuming the auction is still in bidding phase, this will initiate a wallet rescan of maximum 5 days.

See https://github.com/kyokan/bob-wallet/issues/108

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -143,6 +143,7 @@ class RPC extends RPCBase {
     this.add('importaddress', this.importAddress);
     this.add('importprunedfunds', this.importPrunedFunds);
     this.add('importpubkey', this.importPubkey);
+    this.add('importname', this.importName);
     this.add('keypoolrefill', this.keyPoolRefill);
     this.add('listaccounts', this.listAccounts);
     this.add('listaddressgroupings', this.listAddressGroupings);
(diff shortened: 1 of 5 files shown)


lib/wallet/rpc.js
```diff
@@ -889,6 +890,28 @@ class RPC extends RPCBase {
     return null;
   }
 
+  async importName(args, help) {
+    if (help || args.length < 1 || args.length > 2) {
+      throw new RPCError(errs.MISC_ERROR,
+        'importpubkey "name" ( rescanHeight )');
```
boymanjor: s/importpubkey/importname/

lib/wallet/rpc.js
```diff
@@ -889,6 +890,28 @@ class RPC extends RPCBase {
...
+    }
+
+    const wallet = this.wallet;
+    const valid = new Validator(args);
+    const name = valid.str(0);
+    const rescan = valid.u32(1);
```
boymanjor: Let's call this variable `height`. Earlier in the code (`importPubkey`) we use `rescan` as a boolean indicating whether to rescan the wallet.

test/wallet-importname-test.js
```diff
@@ -0,0 +1,226 @@
...
+    assert(ns4);
+  });
+
+  describe('rpc importname', function() {
+    it('should not have name data in Charlie\'s wallet', async () => {
+      const ns1 = await charlie.getNameStateByName(name);
+      assert(ns1 === null);
+    });
+
+    it('should rpc importname with rescan', async () => {
```
boymanjor: nit: "should import name with rescan." The top-level describe makes "should rpc importname with rescan" redundant. Also the latter does read very well to me.

pinheadmz:
@boymanjor thanks bud, nits are picked!

pinheadmz:
rebase to  f8b6de697952e68ba2ab0737ba8cec4db3f94a57:

Rebased on master. This PR combined with #418 are required for upcoming feature in Bob to sync bids from an in-progress auction.

PR #413, opened 2020-03-31
wallet: fix change lookahed address

qimiaoguo:
change address will not be saved if call hsw-cli change

### DIFF
--- a/lib/wallet/account.js
+++ b/lib/wallet/account.js
@@ -344,7 +344,7 @@ class Account extends bio.Struct {
         break;
       case 1:
         key = this.deriveChange(this.changeDepth);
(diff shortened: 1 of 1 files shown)


qimiaoguo:
https://github.com/bcoin-org/bcoin/pull/503
https://github.com/bcoin-org/bcoin/issues/498

pinheadmz:
Thank you. Can you please just include the module name with your commit message:

`wallet: fix change lookahed address`

chjj:
@qimiaoguo, thank you for finding this. A patch update was just published.

I think we should consider a migration path for affected wallets. I think a simple migration would be to regenerate and save all change addresses in the lookahead. The user will most likely have to rescan after that (whether to do this automatically or require user intervention is something I'm unsure of).

Issue #414, opened 2020-03-31
Migration path for wallets affected by the change address bug


chjj:
See:

- https://github.com/handshake-org/hsd/issues/411
- https://github.com/handshake-org/hsd/pull/413

Quoting my post:

> I think we should consider a migration path for affected wallets. I think a simple migration would be to regenerate and save all change addresses in the lookahead. The user will most likely have to rescan after that (whether to do this automatically or require user intervention is something I'm unsure of).

I'd like to get a migration out within the next 24 hours.

pinheadmz:
Closed by #415

Issue #411, opened 2020-03-30
Wallet does not recognize its own change address


pinheadmz:
There have been at least two separate reports of this bug, and I've heard that it may have happened to the Purse bcoin wallet at least once in the past. It's very difficult to diagnose or reproduce but what seems to happen is this:

- The wallet sends a transaction and sends the change output to the next address in the BIP44 `change` path.
- When this transaction is added to the wallet, the change address is not recognized as being owned by the wallet, and has no path information.
- The wallet may continue to use the "ghost" address for change and does not increment the change address index of the wallet account

Additional related behavior: the wallet does not know its own private key for the address:
```
$ hsw-cli change --account=default
{... "address": address...}

$ hsw-cli key <address>
null
```

Even though the wallet may provide the change address again 

**Workaround:** Your coins are not lost, but so far the only way we have been able to recover a wallet in this state is to import the seed into a new wallet and rescan:

```
$ hsw-cli mkwallet --id=recovery --mnemonic="abandon abandon abandon ..."
$ hsw-cli --id=recovery rescan
```

pinheadmz:
This issue may persist on a wallet restored by mnemonic. This user had large gaps in their receive addresses and so I instructed them to manually regenerate addresses until the appropriate index had been reached and rescan again:

> I recovered the wallet, and use generate receiving and change address, and they are same with the original wallet, which contains only 54 received addresses and 26 changed addresses, and the 15, 18, 26 index’s key is lost

pinheadmz:
Closed by #414

PR #415, opened 2020-03-31
wallet: add migration to regenerate missing change addresses (WIP)

chjj:
See #414, #413, #411

### DIFF
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -245,7 +245,7 @@ class WalletDB extends EventEmitter {
       total += await wallet.migrateChange(b);
     }
 
-    if (this.options.migrate === 0)
+    if (this.options.migrate === 0 || total === 0)
       await b.write();
 
     if (total > 0) {
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -248,22 +248,23 @@ class WalletDB extends EventEmitter {
     if (this.options.migrate === 0 || total === 0)
       await b.write();
 
-    if (total > 0) {
-      if (this.options.migrate === 0) {
-        this.logger.warning('Regenerated %d change addresses.', total);
-        this.logger.warning('Rescanning...');
-
-        await this.scan();
-      } else {
-        throw new Error(
-          'Wallet is corrupted.\n' +
-          'Back up wallet and then restart with\n' +
-          '`hsd --wallet-migrate=0` or `hs-wallet --migrate=0`\n' +
-          '(Full node required)'
-        );
-      }
-    } else {
+    if (total === 0) {
       this.logger.info('All change addresses present.');
+      return;
+    }
+
+    if (this.options.migrate === 0) {
+      this.logger.warning('Regenerated %d change addresses.', total);
+      this.logger.warning('Rescanning...');
+
+      await this.scan();
+    } else {
+      throw new Error(
+        'Wallet is corrupted.\n' +
+        'Back up wallet and then restart with\n' +
+        '`hsd --wallet-migrate=0` or `hs-wallet --migrate=0`\n' +
+        '(Full node required)'
+      );
     }
   }
 
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -206,6 +206,65 @@ class WalletDB extends EventEmitter {
       wallet.id, wallet.wid, addr.toString(this.network));
 
     this.primary = wallet;
+
+    await this.migrateChange();
+  }
+
+  /**
+   * Run change address migration.
+   * @param {Wallet} wallet
+   */
+
+  async migrateChange() {
+    if (this.options.memory)
+      return;
+
+    if (await this.db.has(layout.M.encode(0)))
+      return;
+
+    const {prefix} = this.options;
+    const bak = path.join(prefix, `wallet.${Date.now()}.bak`);
+    const b = this.db.batch();
+
+    b.put(layout.M.encode(0), null);
+
+    if (!this.state.marked) {
+      await b.write();
+      return;
+    }
+
+    this.logger.warning('Running change address migration.');
+    this.logger.warning('Backing up database to %s.', bak);
+
+    await this.db.backup(bak);
+
+    const wids = await this.db.keys({
+      gte: layout.W.min(),
+      lte: layout.W.max(),
+      parse: key => layout.W.decode(key)[0]
+    });
+
+    let total = 0;
+
+    for (const wid of wids) {
+      const wallet = await this.get(wid);
+
+      this.logger.info('Regenerating change addresses (id=%s, wid=%d).',
+                       wallet.id, wid);
+
+      total += await wallet.migrateChange(b);
+    }
+
+    await b.write();
+
+    if (total > 0) {
+      this.logger.warning('Regenerated %d change addresses.', total);
+      this.logger.warning('Rescanning...');
+
+      await this.scan();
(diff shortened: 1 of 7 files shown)


lib/wallet/walletdb.js
```diff
@@ -205,6 +205,63 @@ class WalletDB extends EventEmitter {
...
+      return;
+
+    const {prefix} = this.options;
+    const bak = path.join(prefix, `wallet.${Date.now()}.bak`);
+    const b = this.db.batch();
+
+    b.put(layout.M.encode(0), null);
+
+    if (!this.state.marked)
+      return b.write();
```
pinheadmz: lint error here, do the other `returns` need to return `null`?

chjj:
Rebased and addressed nits.

lib/wallet/walletdb.js
```diff
@@ -206,6 +206,65 @@ class WalletDB extends EventEmitter {
...
+      total += await wallet.migrateChange(b);
+    }
+
+    await b.write();
+
+    if (total > 0) {
+      this.logger.warning('Regenerated %d change addresses.', total);
+      this.logger.warning('Rescanning...');
+
+      await this.scan();
```
pinheadmz, line 264: ooh, what if the user is running pruned? (probably not very likely at this point, but makes me think this maybe should not be automatic)

pinheadmz:
rebase to 1351d8035bdbdddebb22c37609c0faedcc7c32a1:
- rebase on current master
- include test from #417 
- replace automatic wallet rescan with a big ugly red log message. The motivation for this is pruned nodes, and potentially SPV nodes as well (https://github.com/bcoin-org/bcoin/issues/835). In the developer meeting it was agreed that hsd shouldn't ever force a rescan, it is unexpected behavior and for a node used in production for an active business it's not great.

Screen Shot 2020-07-22 at 2 59 07 PM

log message can be tested by adding this diff to the test:

```diff
diff --git a/test/wallet-change-test.js b/test/wallet-change-test.js
index 39376570..39d142d0 100644
--- a/test/wallet-change-test.js
+++ b/test/wallet-change-test.js
@@ -19,7 +19,9 @@ const node = new FullNode({
   prefix: path,
   memory: false,
   network: 'regtest',
-  plugins: [require('../lib/wallet/plugin')]
+  plugins: [require('../lib/wallet/plugin')],
+  logConsole: true,
+  logLevel: 'debug'
 });

 const {wdb} = node.require('walletdb');
```

pinheadmz:
@turbomaze

pinheadmz:
New thoughts on this, after testing https://github.com/handshake-org/hsd/pull/480 and having a hard time getting it to work.

Here's what I think the flow could be for this PR:

1. Upgraded hsd checks for walletDB version on launch, if it is not updated, quit the process with an error and instructions to back up wallet db and then...
2. To upgrade hsd, launch with extra option `--migrate-wallet-to-0` or something **in addition to all other params you normally launch with**

This way, all the migration logic can stay in hsd, no external scripts. The user adds the option to all their normal options whatever they are, including prefix and network and all that, so the migration process already has all the data it needs (this is an issue passing all that to the migration script). The user can back up their walletDB any way they want to (we will have to remove the backup logic form this PR as well) and the if the user decides NOT to migrate they still have the option to revert hsd to the last version and relaunch before anything has happened to their data.

Thoughts @tynes @turbomaze ?

pinheadmz:
Rebase to  c25799d87bf5382219a1b6c937e1771c07445375 :

New flow works like this:

- User upgrades to this version of hsd
- Every time new version of hsd is run, check DB migration id
  - if migration ID is present, continue with normal operation
  - if ID is NOT present, scan for missing change addresses:
    - This may take some time (a warning is needed in changelog)
    - The database is not altered in anyway
    - If all addresses are present, DB migration ID is saved, continue with normal operation
    - if addresses are missing, quit and throw an error:

Screen Shot 2020-08-07 at 2 51 17 PM

- Affected user starts again, this time adding `--wallet-migrate=0` to their launch.
  - above sequence is run again
    - this time, if addresses are missing they are regenerated and saved
    - wallet automatically initiates a rescan

Screen Shot 2020-08-07 at 3 03 03 PM

* Because a rescan is required, a standalone wallet user will have to ensure that the wallet node is connected to a hsd full node

Test script: https://gist.github.com/pinheadmz/25fcd11ec8ba3916830ebc01d888bd7b
This script checks out the old version of hsd, manually creates change addresses and dumps a wallet backup. Then it checks out this branch, runs the migration flag and dumps the wallet again.

pinheadmz:
@tynes @turbomaze if you guys are cool with this new scheme I'll spell it all out in the changelog in the next commit,

lib/wallet/walletdb.js
```diff
@@ -206,6 +206,65 @@ class WalletDB extends EventEmitter {
...
+
+      this.logger.info('Checking wallet (id=%s, wid=%d).',
+                       wallet.id, wid);
+
+      total += await wallet.migrateChange(b);
+    }
+
+    if (this.options.migrate === 0 || total === 0)
+      await b.write();
+
```
tynes, line 250: Just reasoning about the control flow here. There is no need to rescan if the total migrated is 0, so skip the inner if/else block. Only rescan if the user specifically opts in, since it is a more costly operation.

lib/wallet/walletdb.js
```diff
@@ -206,6 +206,65 @@ class WalletDB extends EventEmitter {
...
+      const wallet = await this.get(wid);
+
+      this.logger.info('Checking wallet (id=%s, wid=%d).',
+                       wallet.id, wid);
+
+      total += await wallet.migrateChange(b);
+    }
+
+    if (this.options.migrate === 0 || total === 0)
+      await b.write();
```
tynes, line 249: Where does `b` get written to disk in the other case?

lib/wallet/walletdb.js
pinheadmz: This will write the migration ID, though.

lib/wallet/walletdb.js
pinheadmz: It doesn't. It only gets written if either 

1) nothing is wrong, just save the new migration ID
or
2) the flag is set -- at this point the batch contains all the new change addresses. but we don't write them unless the user says to

lib/wallet/walletdb.js
tynes: cACK - this will be easy for people running infra in the cloud, compared to running a script, they just will need to update their config file. Go for spelling it out in the changelog

lib/wallet/walletdb.js
tynes: The migration index set to 0 needs to be written to disk either way after this process so that migrating from 0 to 1 is a more simple process and so that this logic is cut short on additional restarts.

pinheadmz: It should only get written when the wallet corruption is fixed, or if there is nothing to fix. hsd will throw an error if the user refuses to repair their DB, so I don't know how they will even get from 0 to 1 before getting to 0 first.

turbomaze:
> @tynes @turbomaze if you guys are cool with this new scheme I'll spell it all out in the changelog in the next commit,

Looks great!

pinheadmz:
rebase to 1789576961052e0e99ae1c7a5a0e6e4adbde89c1:
rebase on master

pinheadmz:
Ok added a message to the changelog - I think this is done. @chjj @tynes @turbomaze looking for final ACKs and then I say we merge and release hsd.

PR #426, opened 2020-04-15
rpc: add extra_nonce to block's JSON.

kilpatty:
This feels like it should belong here, as I don't know if there is any other way to query the extra_nonce from blocks.

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2923,6 +2923,7 @@ class RPC extends RPCBase {
       time: entry.time,
       mediantime: mtp,
       nonce: entry.nonce,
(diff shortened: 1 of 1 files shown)


tynes:
Good find. It could also be added in the `headerToJSON` function as well.

https://github.com/handshake-org/hsd/blob/9bdbadba46d6faf6478b76cc475ab5b5f9e49bab/lib/node/rpc.js#L2874-L2876

kilpatty:
Good catch, will update

tynes:
This looks good to me.  https://github.com/handshake-org/hsd/pull/426/commits/44bd1bf06dc864fc4d7d7a5c66217d0a63cd4b87

PR #428, opened 2020-04-15
dockerfile: update and add docs.

boymanjor:
This supersedes #288.


tynes:
Following the docker build instructions works for me. I think explicitly noting that it starts 6 servers by default (node http, wallet http, rs, ns, p2p, encrypted p2p), including a DNS server that needs to have UDP exposed could be helpful.

https://docs.docker.com/config/containers/container-networking/#published-ports

Issue #139, opened 2019-04-08
validateresource Node RPC


tynes:
It would be useful to add a new node RPC method `validateresource` that operates in a similar fashion to `validateaddress`.  This would allow developers to craft their resources locally against their node before updating any names with `sendupdate`. Currently if you try to `sendupdate` with an invalid resource, it will submit a transaction with a null resource.

See:
https://github.com/handshake-org/hsd/blob/master/lib/dns/resource.js
https://handshake-org.github.io/api-docs/#sendupdate
https://handshake-org.github.io/api-docs/#getnameresource

PR #429, opened 2020-04-16
node rpc: validateresource

tynes:
This PR adds a new Node RPC method called `validateresource`. It is useful for users who are in the process of creating their resource JSON to use with `sendupdate`. This PR adds error messages in the `fromJSON` methods of each of the records found in `lib/dns/resource.js`.

The RPC accepts a single object with the `records` property, which must be an array of objects.
Eventually, it could also accept a `version` property if another Resource version is implemented.

The error messages note the invalid JSON property name and the record type. This is so that good error messages can be returned to users. The error messages could be made generic (not including the record type in the error message) but that would mean that the RPC method handler itself would have to hold logic similar to `Resource.fromJSON` and keep track of which `record` failed its `fromJSON`. I don't like this approach as much since any new resource type will require its `fromJSON` logic to be added to the RPC method handler but am willing to update this PR if people are strongly opinionated in that way.

Closes https://github.com/handshake-org/hsd/issues/139

### DIFF
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -349,11 +349,11 @@ class Resource extends Struct {
   }
 
   fromJSON(json) {
-    assert(json && typeof json === 'object');
-    assert(Array.isArray(json.records));
+    assert(json && typeof json === 'object', 'Invalid json.');
+    assert(Array.isArray(json.records), 'Invalid records.');
 
     for (const item of json.records) {
-      assert(item && typeof item === 'object');
+      assert(item && typeof item === 'object', 'Invalid record.');
 
       const RD = stringToClass(item.type);
 
@@ -437,13 +437,19 @@ class DS extends Struct {
   }
 
   fromJSON(json) {
-    assert(json && typeof json === 'object');
-    assert(json.type === 'DS');
-    assert((json.keyTag & 0xffff) === json.keyTag);
-    assert((json.algorithm & 0xff) === json.algorithm);
-    assert((json.digestType & 0xff) === json.digestType);
-    assert(typeof json.digest === 'string');
-    assert((json.digest.length >>> 1) <= 255);
+    assert(json && typeof json === 'object', 'Invalid DS record.');
+    assert(json.type === 'DS',
+      'Invalid DS record. Property "type" must be "DS".');
+    assert((json.keyTag & 0xffff) === json.keyTag,
+      'Invalid DS record. Property "keyTag" must be a uint16.');
+    assert((json.algorithm & 0xff) === json.algorithm,
+      'Invalid DS record. Property "algorithm" must be a uint8.');
+    assert((json.digestType & 0xff) === json.digestType,
+      'Invalid DS record. Property "digestType" must be a uint8.');
+    assert(typeof json.digest === 'string',
+      'Invalid DS record. Property "digest" must be a String.');
+    assert((json.digest.length >>> 1) <= 255,
+      'Invalid DS record. Property "digest" is too large.');
 
     this.keyTag = json.keyTag;
     this.algorithm = json.algorithm;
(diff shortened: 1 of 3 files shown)


lib/node/rpc.js
```diff
@@ -241,6 +241,7 @@ class RPC extends RPCBase {
     this.add('grindname', this.grindName);
     this.add('sendrawclaim', this.sendRawClaim);
     this.add('sendrawairdrop', this.sendRawAirdrop);
+    this.add('verifyresource', this.verifyResource);
```
pinheadmz, line 244: Suggestion `validateresource`

pinheadmz:
Looks pretty good, playing with the branch locally and might add a test.

Do you think some of these error message could be even more specific?
e.g.
`TXT record "txt" must be an array`
`Resource "records" must be an array`
`Resource "records" items must be objects`
`DS record keyType must be int < 0xffff`
`DS record "digest" must be string`
etc

lib/dns/resource.js
```diff
@@ -615,10 +619,10 @@ class GLUE6 extends Struct {
   }
 
   fromJSON(json) {
-    assert(json && typeof json === 'object');
-    assert(json.type === 'GLUE6');
-    assert(isName(json.ns));
-    assert(IP.isIPv6String(json.address));
+    assert(json && typeof json === 'object', 'Invalid JSON in GLUE6 record.');
```
boymanjor, line 622: nit: s/JSON/json/

boymanjor:
> Looks pretty good, playing with the branch locally and might add a test.
> 
> Do you think some of these error message could be even more specific?
> e.g.
> `TXT record "txt" must be an array`
> `Resource "records" must be an array`
> `Resource "records" items must be objects`
> `DS record keyType must be int < 0xffff`
> `DS record "digest" must be string`
> etc

How about `Invalid DS record. Property "json" must be an Array.`

tynes:
New commits pushed up, I followed similar error messaging to @boymanjor's suggestion here:
`DS record "digest" must be string`

tynes:
This is ready for another round of review. I wrote tests for both success and failure cases.

lib/dns/resource.js
```diff
@@ -437,13 +437,19 @@ class DS extends Struct {
...
-    assert((json.keyTag & 0xffff) === json.keyTag);
-    assert((json.algorithm & 0xff) === json.algorithm);
-    assert((json.digestType & 0xff) === json.digestType);
-    assert(typeof json.digest === 'string');
-    assert((json.digest.length >>> 1) <= 255);
+    assert(json && typeof json === 'object', 'Invalid DS record.');
+    assert(json.type === 'DS',
+      'Invalid DS record. Property "type" must be "DS".');
+    assert((json.keyTag & 0xffff) === json.keyTag,
+      'Invalid DS record. Property "keyTag" must be a u16.');
```
boymanjor, line 444: nit: let's use `uint16` instead of `u16` to match prior art in the code base.

lib/node/rpc.js
```diff
@@ -2477,6 +2478,27 @@ class RPC extends RPCBase {
     return proof.hash().toString('hex');
   }
 
+  async validateResource(args, help) {
+    if (help || args.length < 1 || args.length > 2)
+      throw new RPCError(errs.MISC_ERROR, 'validateresource'
+        + '"{"records": [{...}]}"');
```
pinheadmz, line 2484: nit: this doesn't print out correctly.
Current:
```
$ hsd-rpc --network=regtest validateresource
validateresource"{"records": [{...}]}"
```

Suggestion:
```
$ hsd-rpc --network=regtest validateresource
validateresource '{"records": [{...}]}'
```

Note the single quotes as well, you may have to use `\'` in the code.

test/node-rpc-test.js
```diff
@@ -130,5 +130,99 @@ describe('RPC', function() {
       const info = await nclient.execute('getblock', [hash]);
       assert.deepEqual(info.confirmations, -1);
     });
+
+    it('should validateresource (valid)', async () => {
+      const records = [
+        [{type: 'NS', ns: 'ns1.handshake.org.'}],
+        [{type: 'DS', keyTag: 0xffff, algorithm: 0xff, digestType: 0xff, digest: '00'.repeat(32)}],
+        [{type: 'TXT', txt: ['i like turtles', 'then who phone']}],
```
pinheadmz, line 138: lol wut.

pinheadmz:
Great work on this dude, thanks for the thorough test. Only thing left for me is the help message.

lib/node/rpc.js
tynes: Good find! Pushed up a fix.

pinheadmz:
ACK @  d5d98219eb3c8f24bfb948a3fd9dc51b7b1fca74

PR #433, opened 2020-04-19
http: add endpoint for balance by address.

boymanjor:
This endpoint allows users to query the balance of an arbitrary
address. It includes the current height of the chain, the unlocked
and locked balances, and arrays of the unlocked and locked coins.

### DIFF
--- a/lib/node/http.js
+++ b/lib/node/http.js
@@ -21,6 +21,7 @@ const Claim = require('../primitives/claim');
 const Address = require('../primitives/address');
 const Network = require('../protocol/network');
 const pkg = require('../pkg');
+const rules = require('../covenants/rules');
 
 /**
  * HTTP
@@ -163,6 +164,37 @@ class HTTP extends Server {
         },
         memory: this.logger.memoryUsage()
       });
+    });
+
+    // Balance by address
+    this.get('/balance/:address', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const address = valid.str('address');
+
+      enforce(address, 'Address is required.');
+      enforce(!this.chain.options.spv, 'Cannot get balance in SPV mode.');
+
+      const addr = Address.fromString(address, this.network);
+      const coins = await this.node.getCoinsByAddress(addr);
+      const result = {
+        height: this.chain.height,
+        balance: 0,
+        coins: [],
+        lockedBalance: 0,
+        lockedCoins: []
+      };
+
+      for (const coin of coins) {
+        if (coin.covenant.type === rules.types.NONE) {
+          result.balance += coin.value;
+          result.coins.push(coin.getJSON(this.network));
+        } else {
+          result.lockedBalance += coin.value;
+          result.lockedCoins.push(coin.getJSON(this.network));
+        }
+      }
+
+      res.json(200, result);
     });
 
     // UTXO by address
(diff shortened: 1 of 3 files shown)


lib/node/http.js
```diff
@@ -165,6 +166,37 @@ class HTTP extends Server {
...
+      const result = {
+        height: this.chain.height,
+        balance: 0,
+        coins: [],
+        lockedBalance: 0,
+        lockedCoins: []
+      };
+
+      for (const coin of coins) {
+        if (coin.covenant.type === rules.types.NONE) {
```
tynes, line 188: You could save needing to import `lib/covenant/rules` by using `covenant.isNone` 

https://github.com/handshake-org/hsd/blob/cb182f47fe9a6d14aa5b6addb59b214dbff705b2/lib/primitives/covenant.js#L234

test/node-http-test.js
```diff
@@ -94,10 +96,133 @@ describe('Node HTTP', function() {
...
 
+  describe('JSON API', function () {
+    it('should get balance by address', async () => {
+      // Generate new cbAddress.
+      const cType = network.keyPrefix.coinType;
+      const nType = network.type;
+      const master = HDPrivateKey.generate();
+      const key = master.derivePath(`m/44'/${cType}'/0'/0/0`);
+      const cbAddress = Address.fromPubkey(key.publicKey).toString(nType);
+      const keyring = KeyRing.fromPrivate(key.privateKey);
```
pinheadmz, line 107: general note: I think you could deploy a `MemWallet` here to simplify this stuff. See `chain-test.js`

lib/node/http.js
```diff
@@ -165,6 +166,37 @@ class HTTP extends Server {
...
+          result.balance += coin.value;
+          result.coins.push(coin.getJSON(this.network));
+        } else {
+          result.lockedBalance += coin.value;
+          result.lockedCoins.push(coin.getJSON(this.network));
```
pinheadmz, line 193: See the function `lockBalances()` in `txdb.js` to double-check but I think other covenant types are considered "unlocked", especially `REDEEM` (actually that might be the only other one besides `NONE`)

lib/node/http.js
```diff
@@ -165,6 +166,37 @@ class HTTP extends Server {
       });
     });
 
+    // Balance by address
+    this.get('/balance/:address', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const address = valid.str('address');
```
pinheadmz, line 172: You might also want to accept a `minconf` parameter. Either that, or include "unconfirmedBalance" kind of stuff in the output. Unconfirmed coins will have a height of `-1` I think

test/node-http-test.js
```diff
@@ -94,10 +96,133 @@ describe('Node HTTP', function() {
...
+      {
+        // Create bid.
+        const coins = await nclient.getCoinsByAddresses([cbAddress]);
+        coins.sort((a, b) => a.height - b.height);
+        const covenants = coins.filter((coin) => {
+          return coin.covenant.type === types.OPEN;
+        });
+        const mtx = new MTX();
+        const ns = await mtx.view.getNameState(node.chain.db, nameHash);
+        const height = Buffer.from([ns.height, 0x00, 0x00, 0x00]);
```
pinheadmz, line 187: This looks a little confusing -- can you use `covenant.push32()` after `mtx.addOutput()` but before `sign`ing ?

lib/node/http.js
boymanjor: We're on the same page. This morning I started updating this PR to resemble the wallet balance API, i.e. `unconfirmed`, `confirmed`, `lockedUnconfirmed`, etc.

boymanjor:
Closing this for now. The approach is flawed.

PR #434, opened 2020-04-19
node rpc: gettxoutset

tynes:
This pull request adds a new node RPC called `gettxoutset`. It returns all `Coins` and I implemented it so that I can observe the UTXO set easily.

While its currently not causing a denial of service, it could far in the future. According to a block explorer, there are currently about 67 million Bitcoin UTXOs. There are currently only 187,035 Handshake UTXOs as of Saturday, April 18th.

https://www.blockchain.com/charts/utxo-count

Using https://github.com/handshake-org/hs-client/pull/27
Note: the timeout and limit are arbitrarily much larger than they need to be
```bash
$ ./bin/hsd-cli rpc --timeout 100000 --limit 1342177280 gettxoutset > coins.json
$ jq '. | length' < coins.json
187035
```

### DIFF
--- a/lib/blockchain/chaindb.js
+++ b/lib/blockchain/chaindb.js
@@ -1109,6 +1109,30 @@ class ChainDB {
   }
 
   /**
+   * Get all coins.
+   * @returns {Promise} - Returns {@link Coin}.
+   */
+
+  async getCoins() {
+    const iter = this.db.iterator({
+      gte: layout.c.min(),
+      lte: layout.c.max(),
+      values: true
+    });
+
+    const coins = [];
+
+    await iter.each((key, value) => {
+      const entry = CoinEntry.decode(value);
+      const [hash, index] = layout.c.decode(key);
+      const coin = entry.toCoin({hash, index});
+      coins.push(coin);
+    });
+
+    return coins;
+  }
+
+  /**
    * Check whether coins are still unspent. Necessary for bip30.
    * @see https://bitcointalk.org/index.php?topic=67738.0
    * @param {TX} tx
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2606,6 +2606,15 @@ class Chain extends AsyncEmitter {
   }
 
   /**
+   * Get all coins.
+   * @returns {Promise} - Returns {@link Coin}[].
+   */
+
+  getCoins() {
+    return this.db.getCoins();
+  }
+
+  /**
    * Get all transaction hashes to an address.
    * @param {Address[]} addrs
    * @returns {Promise} - Returns {@link Hash}[].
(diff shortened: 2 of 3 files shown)


pinheadmz:
If you just want to see the `length` of of `coins`, why not use `rpc gettxoutsetinfo` ?

https://hsd-dev.org/api-docs/#gettxoutsetinfo

I think we should be careful adding endpoints we know will cause major issues in the near future (like `rpc getnames` and eventually `get-coins-by-address`) in case developers start relying on them in their applications

tynes:
> I think we should be careful adding endpoints we know will cause major issues in the near future (like rpc getnames and eventually get-coins-by-address) in case developers start relying on them in their applications

I agree with you here, I implemented this because I wanted a really quick way to view the UTXO set. I think that this code could run into problems if a block is connected/disconnected at the same time (there is no lock), you might not end up with the exact UTXO set at a specific height but be part way through the transition function to the next block. We don't need to merge this, but I do plan on using this implementation in the near future to observe the UTXO set.

---

Would an HTTP endpoint with pagination be more likely to be an acceptable feature?

---

> If you just want to see the length of of coins, why not use rpc gettxoutsetinfo ?

I was just showing that the RPC worked and how I got the number that I cited above.

pinheadmz:
> Would an HTTP endpoint with pagination be more likely to be an acceptable feature?

I think that's headed the right direction, but like you said the UTXO set changes rapidly over time. I think what might be more generally useable is better stats in `gettxoutsetinfo`. Whatever it is you are curious about, can it just be summarized there (like stats on # of each covenant type in the UTXO set)? That kind of thing would have to be regenerated by a rescan, but then would be maintainable in the long run because it would just update on each block.

lib/blockchain/chaindb.js
```diff
@@ -1108,6 +1108,30 @@ class ChainDB {
     return coin.toCoin(prevout);
   }
 
+  /**
+   * Get all coins.
+   * @returns {Promise} - Returns {@link Coin}.
+   */
+
```
tynes, line 1115: If we ever wanted to add `scantxoutset`, this code path would be useful.
https://developer.bitcoin.org/reference/rpc/scantxoutset.html

I imagine a "filter" function passed into `getCoins` that returns a bool that would determine if the `coin` is added to `coins`. If no function is passed in, this filter function would always evaluate true. I think writing a basic parser for output descriptors would be pretty straight forward, although it would likely run into the the issue here: https://github.com/handshake-org/hs-client/pull/27

pinheadmz:
NACK, I think this is an overwhleming amount of data:

```
$ hsd-rpc gettxoutsetinfo
{
  "height": 52346,
  "bestblock": "0000000000000001e312b59ac2e32c34e98a2f9e90391b618faa6544b6024bee",
  "transactions": 2912964,
  "txouts": 1534947,
  "bytes_serialized": 0,
  "hash_serialized": 0,
  "total_amount": 351180078.711527,
  "total_burned": 16513242.761634
}
```

(clearly we have some issues with this rpc, 0 bytes?)

But if there's 1,534,947 UTXO you're gonna have a bad time waiting for this to return ;-)

Just going to close for now.

PR #436, opened 2020-04-22
deps: update blgr

tynes:
This pull request updates the blgr dependency so that it includes https://github.com/bcoin-org/blgr/pull/5.

This is very important for usage, as it enables printing of the error stack for any `logger.error` messages. It will make it much easier to remotely debug with users.

I opened an issue in `bcoin/blgr` about doing a new release here:
https://github.com/bcoin-org/blgr/issues/9

For now, I am using a fork of `bcoin/blgr` as part of this pull request and I'm happy to use the `bcoin/blgr` repo once there is a new release.
https://github.com/Handshake-Protocol/blgr

For users that are currently having trouble, you can run this updated code with the commands:

```bash
$ git clone https://github.com/tynes/hsd
$ cd hsd
$ git checkout update-blgr
$ npm i
$ ./bin/hsd --log-level error
```
Running with `--log-level error` will only print the error messages, so if you would like to still the the `info` and `debug` messages, use `--log-level spam`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -89,9 +89,8 @@
       }
     },
     "blgr": {
-      "version": "0.1.7",
-      "resolved": "https://registry.npmjs.org/blgr/-/blgr-0.1.7.tgz",
-      "integrity": "sha512-+jSaU2jnYqF+ec9e8nzjfjU7QhZIntkDNG8/AEwoxW7J3mmwvmUfAI5lm9BFYs1OzT+1UgIZTFHa2qWjTBURfw==",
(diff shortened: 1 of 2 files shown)


tynes:
The CI failed, it cannot find `blgr`. I recall there being some issue about CircleCI not correctly installing dependencies from github, it did work locally though. Will look for the fix when I have some time.

pinheadmz:
I agree this is an important upgrade but I don't think we should fork from @chjj / bcoin-org blgr just yet. All the hsd dependencies are knot together and upgraded in sync with bcrypto releases, etc. I'm just going to keep bugging JJ to merge and release the log file rotation in blgr

PR #438, opened 2020-04-27
wallet: allow updating of claimed names

tynes:
There is currently a bug in the wallet where it assumes that `REGISTER` outputs only come from `REVEAL` outputs. This prevents any name claim from ever updating their name using the wallet API. It is still possible to manually create the `REGISTER` transaction and broadcast it directly to the blockchain without using the wallet API. This is risky because a user could forget to create the transaction with the additional output containing the value "locked in the claim" to themselves.

https://github.com/handshake-org/hsd/blob/4178fd3c86612e939e554a72f9e4aa85a55a9c37/lib/wallet/txdb.js#L1703-L1710

`REGISTER` outputs can also come from `CLAIM` outputs when the name is reserved.

https://github.com/handshake-org/hsd/blob/4178fd3c86612e939e554a72f9e4aa85a55a9c37/lib/covenants/rules.js#L1218-L1227

These changes do not use `rules.isReserved` and instead directly use `reserved.has` to handle the case when the name is claimed before the claim period ends and then updated for the first time after the claim period ends.

Tests are included that test updating fake claims using the wallet (not the memwallet).

### DIFF
--- a/lib/wallet/txdb.js
+++ b/lib/wallet/txdb.js
@@ -1703,17 +1703,17 @@ class TXDB {
       case types.REGISTER: {
         assert(i < tx.inputs.length);
 
-        const nameHash = covenant.getHash(0);
         const prevout = tx.inputs[i].prevout;
 
-        const brv = await this.getReveal(nameHash, prevout);
-        assert(brv);
+        const coin = await this.getCoin(prevout.hash, prevout.index);
+        assert(coin);
+        assert(coin.covenant.isReveal() || coin.covenant.isClaim());
 
         if (height === -1) {
-          state.ulocked(path, -brv.value);
+          state.ulocked(path, -coin.value);
           state.ulocked(path, output.value);
         } else {
-          state.clocked(path, -brv.value);
+          state.clocked(path, -coin.value);
           state.clocked(path, output.value);
         }
 
@@ -1777,17 +1777,16 @@ class TXDB {
       case types.REGISTER: {
         assert(i < tx.inputs.length);
 
-        const nameHash = covenant.getHash(0);
-        const prevout = tx.inputs[i].prevout;
-
-        const brv = await this.getReveal(nameHash, prevout);
-        assert(brv);
+        const coins = await this.getSpentCoins(tx);
+        const coin = coins[i];
+        assert(coin);
+        assert(coin.covenant.isReveal() || coin.covenant.isClaim());
 
         if (height === -1) {
-          state.ulocked(path, brv.value);
+          state.ulocked(path, coin.value);
           state.ulocked(path, -output.value);
         } else {
-          state.clocked(path, brv.value);
+          state.clocked(path, coin.value);
           state.clocked(path, -output.value);
         }
 
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -16,6 +16,9 @@ const WalletDB = require('../lib/wallet/walletdb');
 const WorkerPool = require('../lib/workers/workerpool');
 const Address = require('../lib/primitives/address');
 const MTX = require('../lib/primitives/mtx');
+const ChainEntry = require('../lib/blockchain/chainentry');
+const {Resource} = require('../lib/dns/resource');
+const Block = require('../lib/primitives/block');
 const Coin = require('../lib/primitives/coin');
 const KeyRing = require('../lib/primitives/keyring');
 const Input = require('../lib/primitives/input');
@@ -24,6 +27,7 @@ const Script = require('../lib/script/script');
 const policy = require('../lib/protocol/policy');
 const HDPrivateKey = require('../lib/hd/private');
 const Wallet = require('../lib/wallet/wallet');
+const rules = require('../lib/covenants/rules');
 const {forValue} = require('./util/common');
 
 const KEY1 = 'xprv9s21ZrQH143K3Aj6xQBymM31Zb4BVc7wxqfUhMZrzewdDVCt'
@@ -70,8 +74,24 @@ function fakeBlock(height) {
     time: 500000000 + (height * (10 * 60)),
     bits: 0,
     nonce: 0,
-    height: height
+    height: height,
+    version: 0,
+    witnessRoot: Buffer.alloc(32),
+    treeRoot: Buffer.alloc(32),
+    reservedRoot: Buffer.alloc(32),
+    extraNonce: Buffer.alloc(24),
+    mask: Buffer.alloc(32)
   };
+}
+
+function curEntry(wdb) {
+  return new ChainEntry(curBlock(wdb));
+}
+
+function nextEntry(wdb) {
+  const cur = curEntry(wdb);
+  const next = new Block(nextBlock(wdb));
+  return ChainEntry.fromBlock(next, cur);
 }
 
 function dummyInput() {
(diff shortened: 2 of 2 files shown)


lib/wallet/txdb.js
```diff
@@ -1780,6 +1787,12 @@ class TXDB {
         const nameHash = covenant.getHash(0);
         const prevout = tx.inputs[i].prevout;
 
+        if (reserved.has(nameHash)) {
+          const credit = await this.getCredit(prevout.hash, prevout.index);
+          assert(credit);
+          return;
+        }
+
```
pinheadmz, line 1795: Notice at the top of this function if an output to the wallet is a CLAIM (or a BID), we add its value to the wallet's "locked" balance. This value needs to be unlocked here, the same way a REGISTER "unlocks" the difference between the REVEAL and names' value (second highest bid). That's what's happening on the next few lines. The wallet's "locked" balance is being increased by reveal amount but decreased by the REGISTER's own amount.

lib/wallet/txdb.js
```diff
@@ -1780,6 +1787,12 @@ class TXDB {
         const nameHash = covenant.getHash(0);
         const prevout = tx.inputs[i].prevout;
 
+        if (reserved.has(nameHash)) {
+          const credit = await this.getCredit(prevout.hash, prevout.index);
+          assert(credit);
+          return;
```
pinheadmz, line 1793: nit: could be `break` to match the rest of the switch?

lib/wallet/txdb.js
```diff
@@ -1706,6 +1707,12 @@ class TXDB {
         const nameHash = covenant.getHash(0);
         const prevout = tx.inputs[i].prevout;
 
+        if (reserved.has(nameHash)) {
```
pinheadmz, line 1710: Will `reserved.has()` work still work after the 4 year claim period expires and users can win these names in auction? We get the whole TX in this funciton, does it have the view also? The best would be if we could test `isClaim()` on the input coin

test/wallet-claim-test.js
```diff
@@ -0,0 +1,271 @@
...
+  }
+});
+
+const walletdb = node.require('walletdb');
+let wallet, claimPeriod;
+
+describe('Wallet Claims', function() {
+  before(async () => {
+    // Allow for fake ownership claims to be mined.
+    ownership.ignore = true;
```
pinheadmz, line 37: The other "fakeclaim" tests put `ownership.ignore = false` in the `finally` block of a try/catch. I think that's best practice because there are failure states where `after` doesn't get executed (i.e. if `before` fails) and could affect other tests.

test/wallet-claim-test.js
```diff
@@ -0,0 +1,271 @@
...
+  workers: false,
+  env: {
+    'HSD_WALLET_HTTP_PORT': ports.wallet.toString()
+  }
+});
+
+const walletdb = node.require('walletdb');
+let wallet, claimPeriod;
+
+describe('Wallet Claims', function() {
```
boymanjor, line 34: We should add tests asserting the coins are properly locked and unlocked during the claim process.

pinheadmz:
I sort of think the test should be written without RPC. It takes the HTTP stuff and API stuff out of the equation. Maybe take a look at the tests in other txdb-focused patches: #262 and #387. If you focus only on the single module you are updating, you may find you can take out a lot of unnecessary stuff like the blockchain! And you may not have to even mine blocks or advance the namestate through its life cycle.

lib/wallet/txdb.js
pinheadmz: Oh right, we're inside the wallet, just get the coin from the wallet :-) 👍

pinheadmz: (nit: `if (!brv)` should work right?)

boymanjor: Or maybe even this:
```js
case types.REGISTER: {
  assert(i < tx.inputs.length);

  const prevout = tx.inputs[i].prevout;
  const coin = await this.getCoin(prevout.hash, prevout.index);

  assert(coin.covenant.isReveal() || coin.covenant.isClaim());

  if (height === -1) {
    state.ulocked(path, -coin.value);
    state.ulocked(path, output.value);
  } else {
    state.clocked(path, -coin.value);
    state.clocked(path, output.value);
  }

  break;
}
```

pinheadmz: Ohhhh my yes.

lib/wallet/txdb.js
```diff
@@ -1777,17 +1777,16 @@ class TXDB {
       case types.REGISTER: {
         assert(i < tx.inputs.length);
 
-        const nameHash = covenant.getHash(0);
         const prevout = tx.inputs[i].prevout;
 
-        const brv = await this.getReveal(nameHash, prevout);
-        assert(brv);
+        const coin = await this.getCoin(prevout.hash, prevout.index);
```
boymanjor, line 1782: For `unlockBalances` we need to use the undo coins, e.g.:
```js
const prevout = tx.inputs[i].prevout;
const coins = await this.getSpentCoins(tx);
const coin = coins[i];

assert(coin);
assert(coin.covenant.isReveal() || coin.covenant.isClaim());

```

tynes:
I migrated the tests to `wallet-test.js` and am no longer using the RPC interface. Still need to add tests that the balances are updating correctly and will remove `wallet-claim-test.js`.

test/wallet-test.js
```diff
@@ -2279,4 +2289,115 @@ describe('Wallet', function() {
...
+
+      const tx = claim.toTX(network, wdb.state.height + 1);
+
+      const cur = new ChainEntry(curBlock(wdb));
+      const next = new Block(nextBlock(wdb));
+      const entry = ChainEntry.fromBlock(next, cur);
+      await wdb.addBlock(entry, [tx]);
+
+      const rawName = Buffer.from('cloudflare', 'ascii');
+      const nameHash = rules.hashName(rawName);
```
pinheadmz, line 2343: You could set `rawName` and `nameHash` at the top of the `describe` block and remove some repeated lines.

test/wallet-test.js
```diff
@@ -2279,4 +2289,115 @@ describe('Wallet', function() {
...
+        records: [{type: 'NS', ns: 'ns1.easyhandshake.com.'}]
+      });
+
+      const update = await wallet.sendUpdate('cloudflare', records);
+
+      // Confirm UPDATE
+      const cur = new ChainEntry(curBlock(wdb));
+      const next = new Block(nextBlock(wdb));
+      const entry = ChainEntry.fromBlock(next, cur);
+      await wdb.addBlock(entry, [update]);
```
pinheadmz, line 2390: This sequence comes up a few times as well, maybe could be replaced with a helper function at the top of the `describe` block.

lib/wallet/txdb.js
```diff
@@ -1777,17 +1778,17 @@ class TXDB {
       case types.REGISTER: {
         assert(i < tx.inputs.length);
 
-        const nameHash = covenant.getHash(0);
-        const prevout = tx.inputs[i].prevout;
+        const coins = await this.getSpentCoins(tx);
```
pinheadmz, line 1781: I don't this is covered by the test yet. You could add another `it` block at the end of the test you already have, store the latest block entry in a local variable, call `walletDB.removeBlock()`, then assert the balances are all as expected.

test/wallet-test.js
```diff
@@ -2279,4 +2299,136 @@ describe('Wallet', function() {
...
+      const nameinfo = await wallet.getNameState(nameHash);
+      assert.deepEqual(nameinfo, null);
+    });
+
+    it('should confirm cloudflare CLAIM', async () => {
+      const claim = await wallet.sendFakeClaim('cloudflare');
+      assert(claim);
+
+      const tx = claim.toTX(network, wdb.state.height + 1);
+
```
boymanjor: nit: do we need this newline?

test/wallet-test.js
```diff
@@ -2279,4 +2299,136 @@ describe('Wallet', function() {
...
+      }
+    });
+
+    it('should send an update for cloudflare', async () => {
+      const records = Resource.fromJSON({
+        records: [{type: 'NS', ns: 'ns1.easyhandshake.com.'}]
+      });
+
+      update = await wallet.sendUpdate('cloudflare', records);
+
```
boymanjor, line 2386: nit: newline.

test/wallet-test.js
```diff
@@ -2279,4 +2299,136 @@ describe('Wallet', function() {
...
+      assert.equal(json.name, 'cloudflare');
+
+      const resource = Resource.decode(ns.data);
+      assert.deepEqual(records.toJSON(), resource.toJSON());
+
+      // The unconfirmed and confirmed values should
+      // take into account the transaction fee. Assert
+      // against the value of the newly created output.
+      const val = update.output(1).value;
+
```
boymanjor, line 2401: nit: newline.

test/wallet-test.js
```diff
@@ -2279,4 +2299,136 @@ describe('Wallet', function() {
...
+      assert.equal(bal.ulocked, 0);
+      assert.equal(bal.clocked, 0);
+    });
+
+    it('should remove a block and update balances correctly', async () => {
+      const cur = curEntry(wdb);
+      await wdb.removeBlock(cur);
+
+      const bal = await wallet.getBalance();
+
```
boymanjor, line 2416: nit: newline.

test/wallet-test.js
```diff
@@ -2279,4 +2299,134 @@ describe('Wallet', function() {
...
+      const val = update.output(1).value;
+      assert.equal(bal.tx, 2);
+      assert.equal(bal.coin, 2);
+
+      // The unconfirmed balance includes value in the mempool
+      // and the chain itself. The reorg'd tx can be included
+      // in another block so the unconfirmed total does not
+      // include the tx fee, since that value has been effectively
```
boymanjor, line 2424: nit: "include the tx fee. That value has..."

test/wallet-test.js
```diff
@@ -2279,4 +2299,134 @@ describe('Wallet', function() {
...
+      // spent already.
+      assert.equal(bal.unconfirmed, val);
+      assert.equal(bal.confirmed, lockup);
+      assert.equal(bal.ulocked, 0);
```
boymanjor, line 2428: `bal.ulocked` is `0`, because the transaction that unlocks the `lockup` value is still in the wallet. However, it is unconfirmed.

lib/wallet/txdb.js
```diff
@@ -1777,17 +1777,16 @@ class TXDB {
       case types.REGISTER: {
         assert(i < tx.inputs.length);
 
-        const nameHash = covenant.getHash(0);
-        const prevout = tx.inputs[i].prevout;
-
-        const brv = await this.getReveal(nameHash, prevout);
-        assert(brv);
+        const coins = await this.getSpentCoins(tx);
```
pinheadmz, line 1780: Is there any possibility that `coins` will be null here and throw an error on `coins[i]` ?

boymanjor: `coins` should at least be an empty array.
https://github.com/handshake-org/hsd/blob/2d5ab277374ef751565caeaac84166c713ef9f4f/lib/wallet/txdb.js#L2760-L2777

test/wallet-test.js
```diff
@@ -70,10 +74,26 @@ function fakeBlock(height) {
...
     bits: 0,
     nonce: 0,
-    height: height
+    height: height,
+    version: 0,
+    witnessRoot: Buffer.alloc(32),
+    treeRoot: Buffer.alloc(32),
+    reservedRoot: Buffer.alloc(32),
+    extraNonce: Buffer.alloc(24),
+    mask: Buffer.alloc(32)
```
pinheadmz, line 83: I assume we need all these values to make `walletDB.removeBlock()` work?

lib/wallet/txdb.js
boymanjor: That said, I'm not sure if we've tested the `txdb.erase()` path. `walletdb.removeBlock()` exercises the `txdb.disconnect()` codepath. `txdb.unlockBalances()` is also called in `txdb.erase()`.

test/wallet-test.js
```diff
@@ -2279,4 +2299,131 @@ describe('Wallet', function() {
...
+      assert.equal(json.name, 'cloudflare');
+      assert.equal(json.state, 'LOCKED');
+
+      const bal = await wallet.getBalance();
+      assert.equal(bal.tx, 1);
+      assert.equal(bal.coin, 1);
+      assert.equal(bal.unconfirmed, lockup);
+      assert.equal(bal.confirmed, lockup);
+      assert.equal(bal.ulocked, lockup);
+      assert.equal(bal.clocked, lockup);
```
pinheadmz, line 2359: For the sake of being totally thorough, I wonder if the balance should be checked before the block is added to assert that the unconfirmed / confirmed values are computed properly through the process.

test/wallet-test.js
```diff
@@ -2279,4 +2299,131 @@ describe('Wallet', function() {
...
+      // take into account the transaction fee. Assert
+      // against the value of the newly created output.
+      const val = update.output(1).value;
+      const bal = await wallet.getBalance();
+      assert.equal(bal.tx, 2);
+      assert.equal(bal.coin, 2);
+      assert.equal(bal.unconfirmed, val);
+      assert.equal(bal.confirmed, val);
+      assert.equal(bal.ulocked, 0);
+      assert.equal(bal.clocked, 0);
```
pinheadmz, line 2406: Again, I'd check balance before and after confirmation just to ensure every little step is totally correct.

lib/wallet/txdb.js
boymanjor: We could test this using `txdb.abandon()`.

test/wallet-test.js
tynes: These values are required to create the `ChainEntry` objects needed to pass to `wdb.addBlock`. See the helper functions `curEntry` and `nextEntry` that correspond to `curBlock` and `nextBlock`

test/wallet-test.js
tynes: Assuming that no additional tests are added in between this `it` block and the `before` block, this is already being tested. See:
https://github.com/handshake-org/hsd/pull/438/files#diff-ccdf9d4d9aaae8ebafdd75a1f30a1b4dR2322-R2328

I do not like that assumption, so I will add in the assertions to this particular `it` block.

lib/wallet/txdb.js
tynes: Good looks, you are right that we haven't tested the `txdb.erase` path. Will add test coverage for that codepath.

tynes:
> We could test this using txdb.abandon().

Added test coverage of this codepath.

tynes:
Added pre and post assertions for wallet balance after indexing blocks/coins. This should be ready for another round of review @pinheadmz @boymanjor

test/wallet-test.js
pinheadmz: 👍

test/wallet-test.js
```diff
@@ -2279,4 +2299,181 @@ describe('Wallet', function() {
...
+      const post = await wallet.getBalance();
+      assert.equal(post.tx, 1);
+      assert.equal(post.coin, 1);
+      assert.equal(post.unconfirmed, lockup);
+      assert.equal(post.confirmed, lockup);
+      assert.equal(post.ulocked, lockup);
+      assert.equal(post.clocked, lockup);
+    });
+
+    it('should close the auction', async () => {
```
pinheadmz, line 2364: technically, not an auction. maybe "should terminate lockup period" ?

test/wallet-test.js
```diff
@@ -2279,4 +2299,181 @@ describe('Wallet', function() {
...
+      assert.equal(pre.tx, 2);
+      assert.equal(pre.coin, 2);
+      assert.equal(pre.unconfirmed, val);
+      assert.equal(pre.confirmed, lockup);
+      assert.equal(pre.ulocked, 0);
+      assert.equal(pre.clocked, lockup);
+
+      // Adding without a block will mark the transaction
+      // as pending, meaning that it was seen in the mempool.
+      await wallet.txdb.add(update);
```
pinheadmz, line 2458: This `txdb.add()` is unnecessary. Test passes without it. And that is kinda the point: you assert the tx is in the wdb in the `getBalance` just above, then abandon, then confirm. If you wanna be extra explicit (optional) you can probably test for `txdb.hasTX(hash)` before and after the abandon.

pinheadmz:
ok this is so close to landing! just down to the nittiest of nits.

test/wallet-test.js
boymanjor: "should advance past lockup period?"

tynes:
@pinheadmz @boymanjor Nits are resolved

Issue #442, opened 2020-05-03
rpc getrawtransaction (possibly others) display address as separate version/hash values


pinheadmz:
This is becoming a FAQ.

See: https://github.com/handshake-org/hs-airdrop/issues/36#issuecomment-586669062

and this question from telegram:

```
lever W, [May 3, 2020 at 9:09:16 AM]:
@pinheadmz Hi boss,  i got vout of tx by "getrawtransaction", but i found there is no address

{
        "value": 26876.53865,
        "n": 3,
        "address": {
          "version": 0,
          "hash": "f608789d349fb47c5de198900f85338f84ff2cc8"
        },
        "covenant": {
          "type": 0,
          "action": "NONE",
          "items": []
        }
      }
```

We should go through all the API responses and make sure all addresses are represented as strings, I'm not sure if there's any reason to express them separately like this.

tynes:
I think we are going to need to add an additional field to prevent breaking changes.

Some possibilities:

- `addr`
- `addressbech32`
- `addrstring`

pinheadmz:
Yes agreed, how about simply...

```
        "address": {
          "version": 0,
          "hash": "f608789d349fb47c5de198900f85338f84ff2cc8",
          "string": "hs1q9jtymgvegg7shxgqtejrvq7d6ky4sy6f4xm3xz"
        },
```

tynes:
I like that idea and can be achieved in a very simple way here:

https://github.com/handshake-org/hsd/blob/1d7251957dd5ce35eb173c3fcdca0ca2479dcb83/lib/node/rpc.js#L2869-L2873

PR #448, opened 2020-05-08
node/rpc: addrToJSON include bech32 string

tynes:
The `address` field in some of the Node RPC methods returns an object with a version and a hash. This pull request adds an additional field to that object called `string`, which returns the bech32 encoded address to make it easier for users that don't want to encode the object client side.

Closes https://github.com/handshake-org/hsd/issues/442

### DIFF
--- a/test/node-test.js
+++ b/test/node-test.js
@@ -616,6 +616,27 @@ describe('Node', function() {
     assert.strictEqual(tx.txid(), tx2.txid());
   });
 
+  it('should get raw transaction (verbose=true)', async () => {
+    const json = await node.rpc.call({
+      method: 'getrawtransaction',
+      params: [tx2.txid(), true],
+      id: '1'
+    }, {});
+
+    assert(!json.error);
+    const tx = TX.fromHex(json.result.hex);
+
+    assert.equal(json.result.vin.length, tx.inputs.length);
+    assert.equal(json.result.vout.length, tx.outputs.length);
+
+    for (const [i, vout] of json.result.vout.entries()) {
+      const output = tx.output(i);
+      assert.equal(vout.address.version, output.address.version);
+      assert.equal(vout.address.string, output.address.toString(node.network));
+      assert.equal(vout.address.hash, output.address.hash.toString('hex'));
+    }
+  });
+
   it('should prioritise transaction', async () => {
     const json = await node.rpc.call({
       method: 'prioritisetransaction',
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2869,7 +2869,8 @@ class RPC extends RPCBase {
   addrToJSON(addr) {
     return {
       version: addr.version,
-      hash: addr.hash.toString('hex')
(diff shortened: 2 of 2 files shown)


test/node-test.js
```diff
@@ -616,6 +616,27 @@ describe('Node', function() {
...
+
+    assert(!json.error);
+    const tx = TX.fromHex(json.result.hex);
+
+    assert.equal(json.result.vin.length, tx.inputs.length);
+    assert.equal(json.result.vout.length, tx.outputs.length);
+
+    for (const [i, vout] of json.result.vout.entries()) {
+      const output = tx.output(i);
+      assert.equal(output.address.version, output.address.version);
```
pinheadmz, line 634: is the first value supposed to be `vout.address.version` ?

pinheadmz:
I'm trying to find other occurrences of this in the other api modules, might be another one in `getblocktemplate`: 

https://github.com/handshake-org/hsd/blob/master/lib/node/rpc.js#L1511-L1537

pinheadmz:
Ok I grepped the api-docs and hsd for other occurrences, I think besides `getblocktemplate` this change should take care of everything. Should we bother updating the examples in the API docs? That might be tedious.

tynes:
> Should we bother updating the examples in the API docs? 

Yes, I think investing in the API docs is a good idea. I can push a PR to update them to the docs repo.

tynes:
> I'm trying to find other occurrences of this in the other api modules, might be another one in `getblocktemplate`:
> 
> https://github.com/handshake-org/hsd/blob/master/lib/node/rpc.js#L1511-L1537

In the response JSON, there are `claims` and `airdrops` that both render the address as a `hash` and `version` without the bech32 encoded string. For `claims`, there is no field called `address`, its `hash` and `version`. For `airdrops`, there is `version` and `address` where `address` is the `hash`.

This means to add in the bech32 encoded address to these, it would need to have inconsistent naming between the two. I was thinking `claims[0].address` and `airdrops[0].string`. Thoughts?

pinheadmz:
OK I think I understand:

https://github.com/handshake-org/hsd/blob/1d7251957dd5ce35eb173c3fcdca0ca2479dcb83/lib/node/rpc.js#L1516-L1517

https://github.com/handshake-org/hsd/blob/1d7251957dd5ce35eb173c3fcdca0ca2479dcb83/lib/node/rpc.js#L1530-L1531

I actually would consider the `address` field in the airdrop section to be a mistake, everywhere else the `hash` is labeled `hash` and the field `string` should be the bech32 string.

So I'm not sure what kind of application consumes this JSON output or if changing it will break anything. My understanding is that GBT is for workers to verify that their mining pool is being honest and fair with its block assembly. I'm not sure if any pools or workers are actually using it for this purpose.

I'll try to get mining pools to weigh in on this. Maybe for this PR we should just leave it alone.

tynes:
I'm leaning towards just leaving it alone for now.

PR #449, opened 2020-05-11
test: remove dead test vectors

tynes:
This PR deletes the file `test/data/resources-v0.json` as it is no longer used


tynes:
The tests here failed because of a race condition, not because of anything to do with the changes.

PR #451, opened 2020-05-14
rpc: resetrootcache

tynes:
When playing with my node, I found that resetting the root cache was sometimes helpful. This adds a node RPC method called `resetrootcache` and it drops the root server cache completely. It would also be technically possible to expose deleting a single name (or a set of names) out of the cache.

DNS servers employ [Dynamic DNS](https://en.wikipedia.org/wiki/Dynamic_DNS) to allow for remote clients to configure the zone, so maybe it makes sense to have expose an RPC interface for the DNS servers. This would give users a more familiar interface for the configuration.

Or maybe restarting the node to clear the cache is good enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## unreleased
 
+### Node API Changes
+
+- Adds a new node rpc `resetrootcache` that clears the root name server cache.
+
 ### Wallet API changes
 
 - Adds new wallet rpc `importname` that enables user to "watch" a name and track
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -95,6 +95,10 @@ class RootCache {
 
     return Message.decode(item.raw);
   }
+
+  reset() {
+    this.cache.reset();
+  }
 }
 
 /**
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -413,6 +413,10 @@ class RootServer extends DNSServer {
     this.logger.info('Root nameserver listening on port %d.', this.port);
   }
 
+  resetCache() {
+    this.cache.reset();
+  }
+
   serial() {
     const date = new Date();
     const y = date.getUTCFullYear() * 1e6;
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -243,6 +243,8 @@ class RPC extends RPCBase {
     this.add('sendrawairdrop', this.sendRawAirdrop);
     this.add('validateresource', this.validateResource);
 
+    this.add('resetrootcache', this.resetRootCache);
+
     // Compat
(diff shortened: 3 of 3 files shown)


pinheadmz:
This is cool and helpful thank you.

Test process:

- Start hsd in regtest clean
-  `dig @127.0.0.1 -p 25349 test-ns` (null response, SOA)
- run [regtest-names script](https://gist.github.com/pinheadmz/ca0e1a5739607307c7a515158ad89c98)
-  `dig @127.0.0.1 -p 25349 test-ns` (SAME null response, from cache) 
- `rpc resetrootcache`
-  `dig @127.0.0.1 -p 25349 test-ns` (results in updated data NS record, etc)

pinheadmz:
@tynes wanna add a description to the CHANGELOG and we can land this one I think?

tynes:
@pinheadmz Added a section to the `CHANGELOG.md` noting the `resetrootcache` RPC method's functionality.

pinheadmz:
I thought of one other thing here, this is the first time in node rpc that some other module is accessed directly. In the future if we have a setting for hsd that disables the DNS servers (or even a hack like https://github.com/kyokan/bob-wallet/issues/140#issuecomment-630318788) then this rpc will throw an error. Waddaya think? Maybe at least wrap the call to `node.ns` in a try/catch with an `RPCError`?

tynes:
Good thinking @pinheadmz, I agree that there should be a way to start the node without the `ns` and `rs`. It seems to me that it would be easiest to do so by adding a new file to `lib/node` that doesn't import the `ns` and `rs` but does everything else in `FullNode`. This makes maintenance a bit more difficult, especially since there is already the `SPVNode`. We could also have a flag like `--no-resolvers` or something like that.

pinheadmz:
Ok great, [this test](https://github.com/handshake-org/hsd/pull/451#issuecomment-630189399) still works, and the rpc command fails gracefully [if DNS is removed from hsd](https://gist.github.com/pinheadmz/25e5846cd879032892efddf1f988d350). Incidentally, the `rs.open()` and `ns.open()` should probably also have `if (this.ns)` etc checks as well, but that is future work. I think this is ok to merge - any last thoughts @tynes ?

tynes:
It does assume that the `ns` has a cache object present on it, maybe it would be better to add a method to the `ns` itself called `resetCache`?

pinheadmz:
good point. I think that would match the surrounding style as well.

tynes:
@pinheadmz Pushed a new commit and squashed

pinheadmz:
Ok still LGTM. If you get a chance to add this (and also `validateresource`) to the API docs, that would be 🍨 🍰 🍕 

I can fix changelog conflicts when we merge.

PR #457, opened 2020-05-19
wallet: rpc getNames add parameter that only returns names owned by U…

### LINKED ISSUE #420: Wallet rpc getnames should allow option to filter "owned by wallet only"
This is a FAQ "how do I know what names I own with my wallet?"

It's currently a two-step process:
- https://hsd-dev.org/api-docs/#getnames-hsw
- https://hsd-dev.org/api-docs/#get-wallet-coin

Get names from wallet, see who the OWNER is (it's a UTXO, tx hash and index) then check if that coin is in your wallet.

We could add a new parameter to `hsw-rpc getnames <own>` that only returns names owned by UTXOs controlled by the wallet.

ca98am79:
…TXOs controlled by the wallet

fixes #420

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -808,6 +808,9 @@ class HTTP extends Server {
 
     // Wallet Name States
     this.get('/wallet/:id/name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const own = valid.bool('own', false);
+
       const height = this.wdb.height;
       const network = this.network;
 
@@ -815,7 +818,14 @@ class HTTP extends Server {
       const items = [];
 
       for (const ns of names)
-        items.push(ns.getJSON(height, network));
+        if (own) {
+          const {hash, index} = ns.owner;
+          const coin = await req.wallet.getCoin(hash, index);
+          if (coin)
+            items.push(ns.getJSON(height, network));
+        } else {
+          items.push(ns.getJSON(height, network));
+        }
 
       res.json(200, items);
     });
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -817,15 +817,17 @@ class HTTP extends Server {
       const names = await req.wallet.getNames();
       const items = [];
 
-      for (const ns of names)
+      for (const ns of names) {
         if (own) {
           const {hash, index} = ns.owner;
           const coin = await req.wallet.getCoin(hash, index);
+
           if (coin)
             items.push(ns.getJSON(height, network));
         } else {
           items.push(ns.getJSON(height, network));
         }
+      }
 
       res.json(200, items);
     });
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -1868,18 +1868,27 @@ class RPC extends RPCBase {
   }
 
   async getNames(args, help) {
-    if (help || args.length !== 0)
-      throw new RPCError(errs.MISC_ERROR, 'getnames');
+    if (help || args.length > 1)
+      throw new RPCError(errs.MISC_ERROR, 'getnames ( own )');
 
+    const valid = new Validator(args);
     const wallet = this.wallet;
     const height = this.wdb.height;
     const network = this.network;
+    const own = valid.bool(0, false);
 
     const names = await wallet.getNames();
     const items = [];
 
     for (const ns of names)
-      items.push(ns.getJSON(height, network));
+      if (own) {
+        const {hash, index} = ns.owner;
+        const coin = await wallet.getCoin(hash, index);
+        if (coin)
+          items.push(ns.getJSON(height, network));
+      } else {
+        items.push(ns.getJSON(height, network));
+      }
 
     return items;
   }
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -1880,15 +1880,17 @@ class RPC extends RPCBase {
     const names = await wallet.getNames();
     const items = [];
 
-    for (const ns of names)
+    for (const ns of names) {
       if (own) {
         const {hash, index} = ns.owner;
         const coin = await wallet.getCoin(hash, index);
+
         if (coin)
           items.push(ns.getJSON(height, network));
       } else {
         items.push(ns.getJSON(height, network));
       }
+    }
 
     return items;
   }
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -53,6 +53,7 @@ const wallet2 = wclient.wallet('secondary');
 let name, cbAddress;
 const accountTwo = 'foobar';
 const ownedNames = [];
+const allNames = [];
 
 const {
   treeInterval,
@@ -438,6 +439,9 @@ describe('Wallet HTTP', function() {
 
     await mineBlocks(treeInterval + 1, cbAddress);
 
+    // Confirmed OPEN adds name to wallet's namemap
+    allNames.push(name);
+
     const json = await wallet.createBid({
       name: name,
       bid: 1000,
@@ -516,6 +520,8 @@ describe('Wallet HTTP', function() {
   it('should get name info', async () => {
     const names = await wallet.getNames();
 
+    assert.strictEqual(allNames.length, names.length);
+
     assert(names.length > 0);
     const [ns] = names;
 
@@ -548,6 +554,9 @@ describe('Wallet HTTP', function() {
 
     await mineBlocks(treeInterval + 1, cbAddress);
 
+    // Confirmed OPEN adds name to wallet's namemap
+    allNames.push(name);
+
     await wallet.createBid({
       name: name,
       bid: 0,
@@ -561,6 +570,9 @@ describe('Wallet HTTP', function() {
     });
 
     await mineBlocks(treeInterval + 1, cbAddress);
+
+    // Confirmed OPEN adds name to wallet's namemap
+    allNames.push(name);
 
     const fn = async () => await wallet.createBid({
       name: name,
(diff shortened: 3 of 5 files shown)


pinheadmz:
Great job with this Mike thank you! Do you think you add a test? It may have to be a bit broader than just the scope of this PR (I think) because I don't think we currently have wallet rpc tests that acquire names. (like, `rpc sendopen` `rpc sendbid` etc aren't executed in any of the test files).

I can help you with this if you need, it would kind of be a combination of `wallet-rpc-test.js` and `wallet-test.js`.

ca98am79:
ok, yes I would be happy to add a test for this

lib/wallet/rpc.js
```diff
@@ -1794,18 +1794,27 @@ class RPC extends RPCBase {
...
 
     for (const ns of names)
-      items.push(ns.getJSON(height, network));
+      if (own) {
+        const {hash, index} = ns.owner;
+        const coin = await wallet.getCoin(hash, index);
+        if (coin)
+          items.push(ns.getJSON(height, network));
+      } else {
+        items.push(ns.getJSON(height, network));
```
randomlogin, line 1816: The same as in the if clause.

lib/wallet/rpc.js
```diff
@@ -1794,18 +1794,27 @@ class RPC extends RPCBase {
...
 
     const names = await wallet.getNames();
     const items = [];
 
     for (const ns of names)
-      items.push(ns.getJSON(height, network));
+      if (own) {
+        const {hash, index} = ns.owner;
+        const coin = await wallet.getCoin(hash, index);
+        if (coin)
```
randomlogin, line 1813: The resulting statement is the same no matter of the if clause.

randomlogin:
And it would be nice to make possible to query some wallet by id, not by 'selecting' it.

pinheadmz:
> And it would be nice to make possible to query some wallet by id, not by 'selecting' it.

The rpc interface, like that in Bitcoin Core, is stateful. This means it can only target one wallet at a time. The REST HTTP API and its `--id=` parameter was added to bcoin to address this.

randomlogin:
> > And it would be nice to make possible to query some wallet by id, not by 'selecting' it.
> 
> The rpc interface, like that in Bitcoin Core, is stateful. This means it can only target one wallet at a time. The REST HTTP API and its `--id=` parameter was added to bcoin to address this.

Is it possible to merge it to hsd?

lib/wallet/rpc.js
pinheadmz: I don't think so - if `own` is set to `true` then `ns.getJSON` is only pushed to the output array if the name owner is a coin in the wallet. If `own` is set to `false` then all `ns` objects returned by `wallet.getNames()` will be returned.

We should be careful though how long this extra DB lookup takes - it could really slow down the response if there's too many names in a wallet.

pinheadmz:
@ca98am79 one more thing that would make this PR complete is to add the `own` parameter to the HTTP API as well:

https://github.com/handshake-org/hsd/blob/fb6ce9fe4a884048e54654a591f0b002e829b81f/lib/wallet/http.js#L788-L800

Let me know if this is something you feel like you wanna do? Otherwise I can take a stab at it myself, then we can get this PR cleaned up and merged! :-)

README.md
```diff
@@ -279,7 +279,7 @@ All wallet calls should be made with `$ hsw-rpc [call] [arguments...]`.
 
 - `getbids [name] [own]` - List own bids on a name.
 - `getreveals [name] [own]` - List own reveals on a name.
-- `getnames` - List all watched names and their statuses.
+- `getnames [own]` - List own names and their statuses.
```
pinheadmz, line 282: If https://github.com/handshake-org/hsd/pull/494 gets merged this may be a conflict. We can leave it for now and see how the other PR goes but either way the docs at https://github.com/handshake-org/handshake-org.github.io will need to be updated.

test/wallet-http-test.js
```diff
@@ -883,6 +883,18 @@ describe('Wallet HTTP', function() {
     assert.deepEqual(state, res);
   });
 
+  it('should get all wallet names', async () => {
```
pinheadmz, line 886: https://github.com/handshake-org/hsd/pull/488 will probably get merged soon -- you can wait and make sure but I think these API calls could use the built in client methods!

test/wallet-http-test.js
```diff
@@ -883,6 +883,18 @@ describe('Wallet HTTP', function() {
...
+    const names = await wallet.client.get(`/wallet/${wallet.id}/name`);
+
+    assert.equal(names.length, 9);
+  });
+
+  it('should only get wallet-owned name resources', async () => {
```
pinheadmz, line 892: It's not actually getting the resource (the on chain DNS data) but more like the "info" (height, value, utxo owner, etc). I think you can just match this phrasing to the last test and say `"should only get wallet-owned names"`

test/wallet-http-test.js
```diff
@@ -883,6 +883,18 @@ describe('Wallet HTTP', function() {
...
+    const names = await wallet.client.get(`/wallet/${wallet.id}/name`, {own: true});
+
+    assert.equal(names.length, 3);
```
pinheadmz, line 895: This is ok but what would be really great is proof that these three names are the names this wallet won in auctions. One way to do this would be create an array `const ownedNames = []` up around line 50 with the other globals. Then in the `it` blocks where the wallet wins a name, `push` that name into the `ownedNames` array. Finally, down here in this `it` block, go through the array of `names` you get from the client and make sure the names you expect are in there.

README.md
ca98am79: I removed this and can update the docs if this gets merged

pinheadmz:
Did a benchmark of sorts by running [this test script](https://gist.github.com/pinheadmz/8103e1de606d2a530eccdb32cfcc66c6) which generates a wallet and bids on 1000 names, only revealing (and therefore winning an owning) 500 of them.

The results were somewhat expected:

```
ALLNAMES: 26.573ms
OWNEDNAMES: 46.835ms
All: 1000
Owned: 500
```
Filtering out the un-owned names from the database call essentially doubles the response time. This is probably ok (at least it's not an exponential increase), but I want to make sure other reviewers are OK with that. We can optimize this in the future for larger wallets by migrating the walletDB, similar to the coin selection issue (have separate buckets for owned and un-owned names in the DB so we don't have to filter anything out).

Falci:
I know this is not the first method with the `own` param.
I had problem with them in the past, because `own=false` also includes `own=true`.
It seems a bit ambiguous to me:
`own=true` filters by the wallet
`own=false` filter out by the wallet or doesn't filter at all?

I know it has a consistent behaviour of "not filtering".

Maybe it could be called `ownOnly`.

This is not a comment the should affect your PR, but we should keep it in mind and maybe come with a consistent refactor/rename?

pinheadmz:
> This is not a comment the should affect your PR, but we should keep it in mind and maybe come with a consistent refactor/rename?

I think as long as we are consistent it should be ok -- and then we can add a note in the API docs introduction that explains that `own` always means "only show me what I own" and the default is to return everything.

pinheadmz:
@ca98am79 sorry this got stale, I'm reviewing our PR backlog finally. So, this PR needs to updated.

Rebase on master:

```
git checkout master
git pull
git checkout 420-wallet-rpc-getnames-own
git rebase master
```

The above commands should sync master branch on your computer with the latest on github, and then re-stack your new commits on top of latest master. I think #529 added more names to the wallets in the test, and now your test fails. This is why we need to try and keep our test self-contained! But that can be a lot of work as well. When I rebase this on master locally and run the tests, here's what fails:

```
  1628 passing (53s)
  1 pending
  2 failing

  1) Wallet HTTP
       should get all wallet names:

      AssertionError [ERR_ASSERTION]: 13 strictEqual 11

      Expected values to be strictly equal:

      13 !== 11

      assert.equal(names.length, 11);
             ^

      at Context.<anonymous> (/Users/matthewzipkin/Desktop/work/hsd/test/wallet-http-test.js:1047:12)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

  2) Wallet HTTP
       should only get wallet-owned names:

      AssertionError [ERR_ASSERTION]: 4 strictEqual 3

      Expected values to be strictly equal:

      4 !== 3

      assert.equal(names.length, 3);
             ^

      at Context.<anonymous> (/Users/matthewzipkin/Desktop/work/hsd/test/wallet-http-test.js:1055:12)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

```

ALSO, the addition to the CHANGELOG needs to be bumped up to the top "unreleased" section.

ALSO-ALSO, we don't have to do this now (unless you want to) but this kind of API call should probably be filtering for expired names as well. We're still a year of mainnet away from any name actually expiring, but it feels like and API to "get my own names only" should probably only return names you can actually still update. Lemme know if you wanna try to address that in this PR or if we should start another branch for that after this one is merged.

lib/wallet/http.js
```diff
@@ -787,14 +787,24 @@ class HTTP extends Server {
...
+
       const height = this.wdb.height;
       const network = this.network;
 
       const names = await req.wallet.getNames();
       const items = [];
 
       for (const ns of names)
-        items.push(ns.getJSON(height, network));
+        if (own) {
```
nodar-chkuaselidze, line 800: Even though it can be omitted, throughout the codebase any multi-line block needs to be in `{}`.

lib/wallet/rpc.js
```diff
@@ -1794,18 +1794,27 @@ class RPC extends RPCBase {
...
+    const valid = new Validator(args);
     const wallet = this.wallet;
     const height = this.wdb.height;
     const network = this.network;
+    const own = valid.bool(0, false);
 
     const names = await wallet.getNames();
     const items = [];
 
     for (const ns of names)
```
nodar-chkuaselidze, line 1809: Same as before. `for (...) {`

nodar-chkuaselidze (changes requested):

rootmateo:
Fantastic work! You guys are killing it

PR #460, opened 2020-05-21
Allow bid and lockup with 0 value in wallet-http

pinheadmz:
Bids and Lockups are allowed to have value of 0, but get rejected by wallet http, which interprets the values as `false` and throws a "missing bid" error.

Also fixes `assert.reject` statements in wallet-http-test.js that weren't actually testing for the correct error message or executing asynchronously.

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -1046,8 +1046,8 @@ class HTTP extends Server {
       const sign = valid.bool('sign', true);
 
       assert(name, 'Name is required.');
-      assert(bid, 'Bid is required.');
-      assert(lockup, 'Lockup is required.');
+      assert(bid != null, 'Bid is required.');
+      assert(lockup != null, 'Lockup is required.');
       assert(broadcast ? sign : true, 'Must sign when broadcasting.');
 
       const options = TransactionOptions.fromValidator(valid);
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -483,6 +483,20 @@ describe('Wallet HTTP', function() {
     await assert.rejects(fn, {message: 'Lockup is required.'});
   });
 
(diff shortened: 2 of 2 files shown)


lib/wallet/http.js
```diff
@@ -1046,8 +1046,8 @@ class HTTP extends Server {
       const sign = valid.bool('sign', true);
 
       assert(name, 'Name is required.');
-      assert(bid, 'Bid is required.');
-      assert(lockup, 'Lockup is required.');
+      assert(bid !== null, 'Bid is required.');
```
boymanjor, line 1049: nit: should we use type coercion here, i.e. `bid != null`? It will protect against the case where`bid` is `undefined`.

pinheadmz: I don't think `bval` returns `undefined` in any case. And I'm pretty sure it only returns `false` specifically if the type is `bool`.

pinheadmz: I think you're right though, will update

pinheadmz:
test failure is non deterministic timeout, force pushing to retrigger CI

tynes:
I confirmed that the updated assertion statements on the bid value and lockup value work as intended, allowing for value 0 to be passed. Without them the newly added test will fail.

test/wallet-http-test.js
```diff
@@ -796,7 +810,7 @@ describe('Wallet HTTP', function() {
       name: name
     }));
 
-    assert.rejects(fn, 'No reveals to redeem.');
```
tynes, line 799: No `await` statement here?

pinheadmz: Thank you!
rebased to 347e8616f0896302374492099c62046a883367ac

pinheadmz:
Damn arbitrary test timeout failures... retriggering CI...

PR #461, opened 2020-05-21
Adding a link to the roadmap

ericmeltzer:


brandondees:
:house: :bike: 

These docs links should go at the top instead of mid-way down where they're easy to miss

smcki012:
ACK

pinheadmz:
NACK. That particular document is already outdated, and in my opinion we should keep the actual hsd code repository agnostic to issues in other repos, and unrelated projects like browser integration. I like the github "projects" function and maybe we can use our time on the bi-weekly developer call to organize what features and fixes will land in upcoming versions. As more developers get involved we can start assigning issues and PR reviews.

I have a personal [development proposal](https://gist.github.com/pinheadmz/96dab51ba9ff9c0b05accd4786f8f7bb) based somewhat on this original list, and ideally we'd see multiple developers posting their personal goals on their own pages!

So I guess my feeling is, the code repository and "list of things the community want" could both be included in some other, higher-level overview of the project. Github has some nice features for this already. https://github.com/handshake-org/hsd is a piece of a bigger puzzle, I don't think the README needs to be the generic portal.

pinheadmz:
Closing this for now. I encourage the community to create a general portal kind of website that include links to hsd and other implementations, as well as a roadmap (or index of roadmaps and development proposals). In this way, hsd is just a piece of a big system.

PR #466, opened 2020-06-03
chaindb: getcoinsbyaddress - handle string addrs

tuxcanfly:
`getCoinsByAddress` currently expects an `Address` instance. I need to query for an address balance, from a plugin which receives a string input. I cannot cast the string to `Address` since they are from different modules. It would be nice to loosen the interface here to accept a string so that plugins can fetch coins by passing a string.

Otherwise, maybe we could add a RPC endpoint which accepts a string. I am currently using the HTTP API endpoint `/coin/address/:address` from the plugin.

### DIFF
--- a/lib/blockchain/chaindb.js
+++ b/lib/blockchain/chaindb.js
@@ -1362,7 +1362,10 @@ class ChainDB {
 
     const coins = [];
 
-    for (const addr of addrs) {
+    for (let addr of addrs) {
+      if (typeof addr === 'string')
+        addr = Address.fromString(addr);
+
       const hash = Address.getHash(addr);
 
       const keys = await this.db.keys({


pinheadmz:
Hm... why didn't CI run on this ? 🤔

pinheadmz:
Looks right, should there be a test in `chain-test.js` to make sure the errors are appropriate if the input is not an Address or string, or if the string is not a valid address, or if the network is wrong, etc?

tynes:
I think that this change makes sense, but the case where an array of addresses passed in is not covered. This gives it a half baked feeling where you can only pass a single string or a single `Address` or an array of `Address` objects.

https://github.com/handshake-org/hsd/blob/22be7faa37b1a948924883394b28747c348d8014/lib/blockchain/chaindb.js#L1360-L1361

See code in bcoin, note that this is from v1.0.2 and no longer exists in v2.0.0.
https://github.com/bcoin-org/bcoin/blob/58ea98dadbbbcc2066a1d4c946cea28f1d2f942b/lib/blockchain/chaindb.js#L1165

tuxcanfly:
@pinheadmz re error handling: errors from parsing the address are bubbled up. I don't see an easy way to make assertions about the address being valid, other than trying to parse it straight up. I think this is fine because even before this change, we were assuming the address is valid e.g see.: `Address.getHash()`

lib/blockchain/chaindb.js
```diff
@@ -1360,6 +1360,10 @@ class ChainDB {
     if (!Array.isArray(addrs))
       addrs = [addrs];
 
+    for (let addr of addrs)
```
tynes, line 1363: Are you sure this will work? I don't think the `addrs` array is being mutated here. Maybe this should go in the for loop below, right above `Address.getHash`. Then it also saves the need to iterate through the array twice.

lib/blockchain/chaindb.js
```diff
@@ -1360,6 +1360,10 @@ class ChainDB {
...
+      if (typeof addr === 'string')
+        addr = Address.fromString(addr);
+
     const coins = [];
 
     for (const addr of addrs) {
```
pinheadmz, line 1369: I think Mark is right - You can change THIS for loop to be `let addr of addrs...` and then mutate strings into objects before it starts processing.

lib/blockchain/chaindb.js
tuxcanfly: Fixed.

lib/blockchain/chaindb.js
tuxcanfly: Thanks, fixed.

pinheadmz:
LGTM @tuxcanfly d'ya wanna squash it? Then I think we're done.

PR #469, opened 2020-06-11
blockstore: flat file block storage

tynes:
This pull request adds a flat file storage backend for Handshake blocks and undo coins. The blocks are stored in the same style as `bitcoind`, allowing for an indexer like [electrs](https://github.com/Blockstream/electrs) to be used to create a highly performant Handshake blockchain index. Huge performance increases where seen for the initial block download when bcoin switched to the flat file backend for blocks.

TODO:

- [ ] Migration script
- [ ] Thorough Review
- [ ] Fix the 3 broken tests in `blockstore-test.js`
- [ ] Benchmarking

Port of the [bcoin blockstore](https://github.com/bcoin-org/bcoin/tree/master/lib/blockstore).


nodech:
Merged in #612

PR #475, opened 2020-07-16
wallet-rpc: switch the order of transactions returned to most recent

kilpatty:
Re: discussion in the Telegram group. Looks like the intended behavior of this RPC call is to return the most recent transactions first, and then the "from/skip" parameter jumps ahead x number of txs.

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -1232,7 +1232,7 @@ class RPC extends RPCBase {
     const valid = new Validator(args);
     let name = valid.str(0);
     const count = valid.u32(1, 10);
-    const from = valid.u32(2, 0);
+    const skip = valid.u32(2, 0);
     const watchOnly = valid.bool(3, false);
 
     if (wallet.watchOnly !== watchOnly)
@@ -1248,11 +1248,12 @@ class RPC extends RPCBase {
 
     common.sortTX(txs);
 
-    const end = from + count;
-    const to = Math.min(end, txs.length);
+    const start = txs.length - skip - 1;
+    const end = start - count;
+    const to = Math.max(end, -1);
     const out = [];
 
-    for (let i = from; i < to; i++) {
+    for (let i = start; i > to; i--) {
       const wtx = txs[i];
       const json = await this._toListTX(wtx);
       out.push(json);


lib/wallet/rpc.js
```diff
@@ -1232,7 +1232,7 @@ class RPC extends RPCBase {
     const valid = new Validator(args);
     let name = valid.str(0);
     const count = valid.u32(1, 10);
-    const from = valid.u32(2, 0);
+    const skip = valid.u32(2, 0);
```
pinheadmz, line 1235: We should also change the argument name in the help message up on line 1228

lib/wallet/rpc.js
```diff
@@ -1248,11 +1248,12 @@ class RPC extends RPCBase {
 
     common.sortTX(txs);
 
-    const end = from + count;
-    const to = Math.min(end, txs.length);
+    const start = txs.length - skip - 1;
+    const end = start - count;
+    const to = Math.max(end, -1);
```
pinheadmz, line 1253: I think I'd prefer a `0` here and a `>=` below

pinheadmz:
See https://github.com/bcoin-org/bcoin/pull/605 -- this rpc call is already problematic for two reasons:

1. It's not paginated, so a wallet with a huge history is going to timeout anyway
2. Transactions in the `txdb` aren't necessarily stored by time, this is a big part of what bcoin-org/bcoin#605 is fixing above.

In practice, it does seem like `txdb` is returning transactions sorted by time and not txid:

```
--> hsw-rpc --network=regtest listtransactions | grep -e '"time"'
    "time": 1595446991,
    "time": 1595446994,
    "time": 1595446997,
    "time": 1595447000,
    "time": 1595447004,
    "time": 1595447008,
    "time": 1595447013,
    "time": 1595447017,
--> hsw-rpc --network=regtest listtransactions | grep -e '"txid"'
    "txid": "f05e64fe8465c7316bfcd9e2f7188b4bb9806133167c5b042dfe42dfca49da11",
    "txid": "8b54409ab0ed03628a9e8dca0c96f4c538aa706e817370ee64bc7fbb3625d660",
    "txid": "6a18e9336b183d422f70e1f0b41291dd21aa96563acfddd6bb94f7cf3c3a3ce1",
    "txid": "e68d660f8b7772a7afff6db245dce3f54eadb9f8a943d402096ade787b958e07",
    "txid": "bb597f168fb3fb8f7c26ac7aee58fe38be1dda31acba74a597b83e1bad35b35b",
    "txid": "7211b52a67fa3601ba22d59412b5e0c234cb05ec7db718ae21f8a36b61b0817c",
    "txid": "d4659a0622d83edf0bda98a7f76e44105e5cf6425656d712f3518c1a6633c001",
    "txid": "6628302a4ec2725db511766f22369618cf62585d89b5e06559fdd15252bb9067",
```

But that is not intentional and might be a result of the order in which transactions were inserted into the database:

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/wallet/txdb.js#L2581-L2594

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/wallet/layout.js#L61

So although I think this PR is useful and makes the API match [Bitcoin Core rpc listtransactions](https://chainquery.com/bitcoin-cli/listtransactions) it should be noted that its behavior is actually unpredictable.

Also, if we merge this we need to add a message to the CHANGELOG and update the api-docs.

lib/wallet/rpc.js
kilpatty: Good call, adding this in

kilpatty:
Following up here, I think without pagination, even these sight edits are somewhat useless when it comes to a wallet with a large amount of transactions. 

I'm currently seeing roughly 1-2 second response times for this RPC call, and so I think the next step would be to mimic bitcoin core and introduce some kind of pagination. I'll need to do some reading to see how they do it, and if it will require a change to how the transactions are indexed in the walletDB, but will report back here when I get the time to investigate.

pinheadmz:
Be sure to check out bcoin #605, Braydon wrote pagination stuff already, it just never got reviewed or merged.

pinheadmz:
@kilpatty should we close this and work on a deeper fix instead?

kilpatty:
I might actually pull the edits from bcoin #605 into this PR and rename. This is a relatively high priority for me as I need it for some internal software, so I'll try to get on it tonight and/or tomorrow. If you want me to just pop it into a new one happy to do that as well.

pinheadmz:
Ok cool, up to you on which branch you push to :-)

PR #476, opened 2020-07-17
Removes bid assert for nonce

Falci:
`assert(bid)` will fail for `0`.
[covenants/rules](https://github.com/handshake-org/hsd/blob/8c7053e45be78be638455e935dff7a8f8c3a5f4d/lib/covenants/rules.js#L385) accepts `0`;

This is affecting the Bob's Repair bid feature: kyokan/bob-wallet/issues/163


pinheadmz:
Thank you! I don't think we should completely remove the check, however. See how this was accomplished in this related pull request :
https://github.com/handshake-org/hsd/pull/460

PR #477, opened 2020-07-17
Fix nonce with bid=0

Falci:
Similar to #460 but for nonce.
Related to kyokan/bob-wallet/issues/163


pinheadmz:
ACK, thank you. One nit: could you add the module name to your commit message? Sorry we need to add contributor docs to make this more explicit, but hsd commits usually look like this:

`wallet-http: Fix nonce with bid=0`

Falci:
Done.

Issue #463, opened 2020-05-31
Wallet creation of zero-bid transactions 


randomlogin:
**Description**: hsw allows creation of zero-bid transaction without any notification that this transaction will not be accepted.

The created transaction can be seen in the wallet's pending transactions list, however the node will not accept it due to dust limits (the transaction itself has nonzero fee), so it will not be seen in node's own mempool and not relayed to anyonel.

**Reproducing**: ```hsw-cli rpc sendbid <name> 0 0```

**Expected behaviour**. Either of these:
1. Tell the user that transaction fails verification, nevertheless create the transaction. 
2. Fail to create the transaction if some special flag like '--ignore-validation-rules` is not passed.

**Related issues**:
#460 Allow bid and lockup with 0 value in wallet-http
#441 wallet: do check covenants for dust in createTX.

pinheadmz:
Thanks @randomlogin 

Issues like this have come up before and it's because bcoin and hsd wallet ignore standardness errors when sending TXs. 

https://github.com/pinheadmz/hsd/blob/bid0/lib/node/fullnode.js#L327-L353

I've never been a fan of this logic, I think it was written for a problem purse had with bcoin?

PR #479, opened 2020-07-21
wallet: check auction TXs for dust and null address before sending

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/463 maybe a few more...

These sanity checks are already applied to regular money-sending `NONE`-covenant transactions in the [`createTX()` method](https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/wallet/wallet.js#L3276-L3306)

However there is nothing yet to prevent a user from sending a BID with a lockup value that is below the dust threshold.

This is another issue with my least favorite line in the codebase, the "broadcast anyway" clause:

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/node/fullnode.js#L335-L342

As much as I want to change the "broadcast anyway" behavior we still need the extra sanity check in the wallet for SPV nodes, which can't check new TXs against mempool policy.

### DIFF
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -3400,6 +3400,19 @@ class Wallet extends EventEmitter {
     if (ancestors.size + 1 > this.maxAncestors)
       throw new Error('TX exceeds maximum unconfirmed ancestors.');
 
+    for (const output of tx.outputs) {
+      if (output.isDust())
+        throw new Error('Output is dust.');
+
+      if (output.value > 0) {
+        if (!output.address)
+          throw new Error('Cannot send to unknown address.');
+
+        if (output.address.isNull())
+          throw new Error('Cannot send to null address.');
+      }
+    }
+
     await this.wdb.addTX(tx);
 
     this.logger.debug('Sending wallet tx (%s): %x', this.id, tx.hash());
--- a/test/wallet-auction-test.js
+++ b/test/wallet-auction-test.js
@@ -12,6 +12,7 @@ const Miner = require('../lib/mining/miner');
 const WalletDB = require('../lib/wallet/walletdb');
 const Network = require('../lib/protocol/network');
 const rules = require('../lib/covenants/rules');
+const Address = require('../lib/primitives/address');
 
 const network = Network.get('regtest');
 const NAME1 = rules.grindName(5, 2, network);
@@ -121,6 +122,17 @@ describe('Wallet Auction', function() {
       assert(block);
       assert(await chain.add(block));
     }
+  });
+
+  it('should fail to send bid to null address', async () => {
+    const mtx = await winner.makeBid(NAME1, 1000, 2000, 0);
+    mtx.outputs[0].address = new Address();
+    await winner.fill(mtx);
+    await winner.finalize(mtx);
+
+    const fn = async () => await winner.sendMTX(mtx);
+
+    await assert.rejects(fn, {message: 'Cannot send to null address.'});
   });
 
   it('should fail to re-open auction during BIDDING phase', async () => {
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -508,7 +508,7 @@ describe('Wallet HTTP', function() {
     await assert.rejects(fn, {message: 'Lockup is required.'});
   });
 
-  it('should send bid with 0 value and 0 lockup', async () => {
+  it('should send bid with 0 value and non-dust lockup', async () => {
     await wallet.client.post(`/wallet/${wallet.id}/open`, {
       name: name
     });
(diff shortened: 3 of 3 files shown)


pinheadmz:
also worth mentioning: the reason this hasn't come up in tests yet is because regtest allows nonstandard TXs:

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/protocol/networks.js#L841

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/mempool/mempool.js#L1437-L1440

https://github.com/handshake-org/hsd/blob/a3049d5b7e7feb431419e3c4d092bb71e17c86fc/lib/primitives/tx.js#L1128-L1151

test/wallet-auction-test.js
```diff
@@ -123,6 +124,17 @@ describe('Wallet Auction', function() {
...
 
+  it('should fail to send bid to null address', async () => {
+    const mtx = await winner.makeBid(NAME1, 1000, 2000, 0);
+    mtx.outputs[0].address = new Address();
+    await winner.fill(mtx);
+    await winner.finalize(mtx);
+
+    const fn = async () => await winner.sendMTX(mtx);
+
+    assert.rejects(fn, {message: 'Cannot send to null address.'});
```
tynes, line 135: I changed the error message to verify that it does indeed fail when the message is different.

test/wallet-http-test.js
```diff
@@ -483,18 +483,34 @@ describe('Wallet HTTP', function() {
...
 
-  it('should send bid with 0 value and 0 lockup', async () => {
+  it('should send bid with 0 value and non-dust lockup', async () => {
     await wallet.client.post(`/wallet/${wallet.id}/open`, {
       name: name
     });
 
     await mineBlocks(treeInterval + 1, cbAddress);
 
     await wallet.client.post(`/wallet/${wallet.id}/bid`, {
```
tynes, line 493: The wallet http methods have been included in a `hs-client` release so we no longer have to use the underlying `client.post` method. Out of scope for this PR but low hanging fruit for another one.

test/wallet-auction-test.js
tynes: You should add an `await` here, it still works without it but we have seen race conditions in the past without it. It fails even without an await when the error message doesn't match.

test/wallet-http-test.js
```diff
@@ -483,18 +483,34 @@ describe('Wallet HTTP', function() {
...
+
+    await mineBlocks(treeInterval + 1, cbAddress);
+
+    const fn = async () => await wallet.client.post(`/wallet/${wallet.id}/bid`, {
       name: name,
       bid: 0,
       lockup: 0
     });
+
+    await assert.rejects(fn, {message: 'Output is dust.'});
```
tynes, line 513: Verified that changing the error message causes the test to fail.

test/wallet-auction-test.js
pinheadmz: thanks will fix

test/wallet-http-test.js
pinheadmz: Good catch, opened: https://github.com/handshake-org/hsd/issues/486

test/wallet-http-test.js
pinheadmz: 👍

pinheadmz:
rebase to 0dcf5e19d94037aa2cdef5a945b9f1c40588a7e6:

address https://github.com/handshake-org/hsd/pull/479#discussion_r464773564

pinheadmz:
@tynes added that `await` - can a brutha get an ACK?

lib/wallet/wallet.js
```diff
@@ -3400,6 +3400,19 @@ class Wallet extends EventEmitter {
...
 
+    for (const output of tx.outputs) {
+      if (output.isDust())
+        throw new Error('Output is dust.');
+
+      if (output.value > 0) {
+        if (!output.address)
+          throw new Error('Cannot send to unknown address.');
+
+        if (output.address.isNull())
```
tynes, line 3411: What if somebody wants to provably burn some HNS? This would prevent somebody from using the API  to do so. I don't see a consensus or policy rule that prevents such an output from existing.

pinheadmz: I think they can use a nulldata address then. This check I think is to catch user error when an address isn't specified in the output and the null value remains:

https://github.com/handshake-org/hsd/blob/940f463259e760f2c60d0b5a53a5bdab1dd83213/lib/primitives/address.js#L99-L112

tynes: Ok, looking at the implementation of `isNull`, it doesn't check the address.version so this would still let somebody provably burn since its not checking `Address.isNulldata`

tynes: The same block of code is here: https://github.com/handshake-org/hsd/blob/89d3ab4c81969e6e14306d567c8f1411ad36c254/lib/wallet/wallet.js#L3288

tynes:
ACK 0dcf5e19d94037aa2cdef5a945b9f1c40588a7e6

lib/wallet/wallet.js
pinheadmz: Right - if they used nulldata (which is just address version 31) they just couldn't ALSO use a `ZERO_HASH` as the data part.

pinheadmz:
rebase to  cf6d606e173a0b03abf692d1dca63bbe7832580b:

rebase on master, let's see the new github workflows run!

PR #480, opened 2020-07-23
walletdb: change migration script

tynes:
This PR moves the migration out of the node boot process and instead uses a script to do the migration. Users giving feedback seemed to like doing the migration through a script instead of through stopping, pulling a new version and restarting the node.

### DIFF
--- a/migrate/wallet0.js
+++ b/migrate/wallet0.js
@@ -1,0 +1,50 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const WalletDB = require('../lib/wallet/walletdb');
+const path = require('path');
+const os = require('os');
+const Logger = require('blgr');
+
+(async () => {
+  const args = process.argv.slice(2);
+
+  let network = 'main';
+  let prefix = path.join(os.homedir(), '.hsd');
+  let level = 'info';
+
+  while (args.length > 0) {
+    const arg = args.shift()
+    switch (arg) {
+      case '-n':
+      case '--network':
+        network = args.shift();
+        continue;
+      case '-p':
+      case '--prefix':
+        prefix = args.shift();
+      case '-l':
+      case '--log-level':
+        level = args.shift();
+    }
+  }
+
+  const logger = new Logger(level);
+  await logger.open
+
+  const wdb = new WalletDB({
+    network: network,
+    logger: logger,
+    prefix: prefix,
+    memory: false
+  });
+
+  await wdb.open();
+  await wdb.migrateChange();
+  await wdb.close()
+
+})().catch(err => {
+  console.log(err)
+  process.exit(1);
+});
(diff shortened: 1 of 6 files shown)


pinheadmz:
I'm OK with this approach, from the developer call it sounds like any update where backing up the walletDB is prudent should not be done automatically, and we already discussed why rescanning automatically isn't a good idea either. I think this is good design practice for hsd going forward, although I'll also add here that in https://github.com/handshake-org/hsd/pull/396 the migration was executed automatically in updated software. In contrast, in https://github.com/bcoin-org/bcoin/pull/703 the user had to migrate the chainDB manually after updating bcoin, but before launching it again.

So I think @tynes if you have the time what I think would be good here is this:

- In walletDB, check for the migration flag `this.db.has(layout.M.encode(0))` in `walletdb.open()` and log an error. This will be output to debug log on every single launch of hsd until the issue is fixed. That error should either specify "see changelog" or directly reference the script `migrate/wallet0.js`. 

- Add a description of the process (including how to back up the walletDB) in the CHANGELOG

- In the root README, right after `INSTALL` and a section `UPDATING` and if nothing else, just add "always read the changelog before updating"

Another option would be test for the migrate flag and instead of loging an error, just stop hsd with `process.exit()` and throw a descriptive error message -- thought on this @turbomaze ?

Will test the actual migration script shortly

migrate/wallet0.js
```diff
@@ -0,0 +1,50 @@
...
+      case '--prefix':
+        prefix = args.shift();
+      case '-l':
+      case '--log-level':
+        level = args.shift();
+    }
+  }
+
+  const logger = new Logger(level);
+  await logger.open
```
pinheadmz, line 34: missed `()` and a few other lint errors.

Suggestion:

```diff
diff --git a/migrate/wallet0.js b/migrate/wallet0.js
index 8373ca0c..b2d65762 100755
--- a/migrate/wallet0.js
+++ b/migrate/wallet0.js
@@ -15,7 +15,7 @@ const Logger = require('blgr');
   let level = 'info';
 
   while (args.length > 0) {
-    const arg = args.shift()
+    const arg = args.shift();
     switch (arg) {
       case '-n':
       case '--network':
@@ -31,7 +31,7 @@ const Logger = require('blgr');
   }
 
   const logger = new Logger(level);
-  await logger.open
+  await logger.open();
 
   const wdb = new WalletDB({
     network: network,
@@ -42,9 +42,8 @@ const Logger = require('blgr');
 
   await wdb.open();
   await wdb.migrateChange();
-  await wdb.close()
-
-})().catch(err => {
-  console.log(err)
+  await wdb.close();
+})().catch((err) => {
+  console.log(err);
   process.exit(1);
 });
```

pinheadmz:
Hoping we can close this and rally around merging https://github.com/handshake-org/hsd/pull/415

tynes:
Closing in favor of #415

Issue #446, opened 2020-05-08
update changelog


pinheadmz:
- Include all new API commands
- List major bug fixes, especially wallet change migration (explain bug and indicate that rescan is required after upgrade)

PR #482, opened 2020-07-27
Update CHANGELOG with recent significant changes

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/446

(Did I miss anything?)

I think going forward, we should be vigilant about asking contributors to add details to CHANGELOG when necessary. I think little bug fixes (like allow bids of `0` value, etc) can be ignored. But definitely API changes should be recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,27 @@
 
 ## unreleased
 
+### Node API changes
+
+- A new RPC call `validateresource` was added to validate Handshake `Resource`
+JSON and will return an error message on an invalid `Resource`. The input JSON
+object is the format expected by `rpc sendupdate`.
+
+- A new RPC call `getdnssecproof` was added to build and return the DNSSEC
+proof used for reserved name claims. This can be used to test if a reserved
+name is ready for a CLAIM.
+
 ### Wallet API changes
 
 - Adds new wallet rpc `importname` that enables user to "watch" a name and track
 its auction progress without bidding on it directly.
+
+### Wallet changes
+
+- A bug was fixed that prevented reserved names that had been CLAIMed from
+REGISTERing. If unpatched software was used to CLAIM a name already, that wallet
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,26 @@
 
 ## unreleased
 
+### Node API changes
+
```
tynes, line 6: While the name of the RPC is descriptive, I think that its useful to say what the RPC does exactly. In this case, `validateresource` will validate Handshake Resource JSON and will return an error message on an invalid Resource

CHANGELOG.md
```diff
@@ -2,11 +2,26 @@
...
+- A new RPC call `validateresource` was added to provide meaningful feedback
+to users before executing `rpc sendupdate` or `rpc sendregister`.
+
+- A new RPC call `getdnssecproof` was added to test if a reserved name is ready
```
tynes, line 10: `getdnssecproof` builds and returns the DNSSEC proof used for name claims. This can be used to test if a reserved name is ready...

pinheadmz:
rebase to a16a37cd57ff5233cc6034b8c6f8f7d2080196c8:

Address feedback from @tynes

tynes:
> (Did I miss anything?)

https://github.com/handshake-org/hsd/commit/60da1f87ae570d13f5d4316911b2588f5747d5fc
https://github.com/handshake-org/hsd/commit/3db8bf8b6aa904335035780ab12d04dfcd6921dd - this is important because it is the only way to learn if a block is in the main chain or not from the API
https://github.com/handshake-org/hsd/commit/36825359f003794a766c3231458bb0e3a91c74f6

pinheadmz:
@tynes thank you! Added a commit.

pinheadmz:
Looking for one last ACK on this @tynes then we'll merge

tynes:
ACK 31e129e46d8248390ca18253c761e01a04ff7315

PR #484, opened 2020-07-29
github: switch to workflows, coveralls

tuxcanfly:
While here, removed scripts which depend on ci tools.


package.json
```diff
@@ -62,14 +62,10 @@
     "hsw-cli": "./bin/hsw-cli"
   },
   "scripts": {
-    "lint": "eslint $(cat .eslintfiles) || exit 0",
-    "lint-ci": "eslint $(cat .eslintfiles)",
-    "lint-file": "eslint",
```
pinheadmz, line 67: I still use `npm run lint`, have eslint installed in global node_modules. Do we need to remove this? (the `lint-ci` I understand, sure)

pinheadmz:
So I guess this has to be in the master branch of the repo to run it, and then the results auto-magically are output in the "actions" tab. I can see it on your repo: https://github.com/tuxcanfly/hsd/actions but the outputs don't look great - did it run the tests? Also looks like you can copy embed html for a status badge and add that to the readme, that might be cool.

package.json
tuxcanfly: I don't think scripts should reference tools which are not locally installed. If you are using a global `eslint` might as well run it directly.

pinheadmz: You're keeping the eslint files in the root directory though. What if eslint was listed in `peerDependencies` or something?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: npm test
+      uses: actions/setup-node@v1
+    - run: sudo apt-get install -y libunbound-dev
+    - run: npm ci
+    - run: npm run build --if-present
+    - run: npm install -g nyc bmocha
+    - run: nyc --reporter lcov bmocha -- --reporter spec test/*.js
+
+    - name: eslint
+      uses: getsentry/action-eslint-fix@v1
```
pinheadmz, line 29: no `run` statement? I do kinda like when CI catches lint errors from contributors, it can help identify bugs quickly and keeps the codebase style consistent.

tuxcanfly: Don't need a run statement as it is a plugin.

tuxcanfly: Must mention I had an issue with this plugin and ended up switching to `hallee/eslint-action@1.0.3`

tuxcanfly: You can try the changes locally by pushing this branch to your fork.

tuxcanfly: > no `run` statement? I do kinda like when CI catches lint errors from contributors, it can help identify bugs quickly and keeps the codebase style consistent.

See here for an example of how linter can do annotations:

https://github.com/tuxcanfly/hsd/runs/927111110?check_suite_focus=true

lib/dns/resource.js
```diff
@@ -880,6 +880,7 @@ function typeToClass(type) {
 function stringToClass(type) {
   assert(typeof type === 'string');
 
+  // eslint-disable-next-line no-prototype-builtins
```
pinheadmz, line 883: So I just noticed this locally as well - `no-prototype-builtins` was added to ESlint somewhere along the way. But the CircleCI script uses https://github.com/bcoin-org/bslint which wraps `eslint@5.10.0`.

We could add the exception in .eslintrc.json ? Or maybe require bslint as a  devDependency? Like bmocha?


How would you feel about that? Then the `npm run lint` command could just call bslint the same way the test commands call bmocha.

tuxcanfly: Yeah `devDependency` route sounds good.

tuxcanfly: OK, using `bslint` now. DIdn't want to clutter deps, so latest is being installed on the CI.

.github/workflows/node.js.yml
tuxcanfly: Decided to drop annotations for now. Running simple `eslint`.

tuxcanfly:
@pinheadmz I decided to keep the workflow as simple as possible. So it only gets the deps, runs eslint and uses the coverall action. Fancy stuff like annotations can be done later.

pinheadmz:
Ok I think this is pretty cool. I tried it out on my own fork with a few PRs: https://github.com/pinheadmz/hsd/pulls

One of the PRs I broke a bunch of tests by changing code in the tests (not the source) and the CI seems to be stuck on that one (yellow indicator, job not completing, also tried cancel and restart workflow): https://github.com/pinheadmz/hsd/pull/4/checks?check_run_id=933065232

The [coverage reports](https://github.com/pinheadmz/hsd/pull/5#issuecomment-667270299) are basically working - I think corevalls isn't set up on my github account so it doesnt have access to source files to display the code coverage.

We spoke out of band about linting: I think, add `bslint` to `devDependencies` and replace `esint` with `bslint` in the commands in `package.json`. Then we can just remove the file `.eslintfiles` since basically everything is covered already, and the command CI runs tests `.` -- I confirmed the linter checked the `lib/` and `test/` directories on CI.

Great stuff @tuxcanfly !!!

Also hoping to get eyes on this from @tynes and hopefully @braydonf as well since this might be a good move for bcoin as well.

pinheadmz:
That one run took six hours 😜

```
Run failed for FAIL-TEST (481570e)

Repository: pinheadmz/hsd
Workflow: ci
Duration: 6 hours and 29.0 seconds
```

tuxcanfly:
OK added `bslint` as dev dependency.

Note: the executable is still called `eslint` presumably for compatibility.

https://github.com/chjj/bslint/blob/master/bin/eslint

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - uses: actions/checkout@v2
+    - name: setup
+      uses: actions/setup-node@v1
```
pinheadmz, line 18: Should we specify a node version here? Looks like by default at this time its `v10.22.0` which is fine because thats our minimum required. We should just keep an eye on it though (https://github.com/actions/setup-node) in case they bump the default version upstream and we want to keep our tests running on v10

tuxcanfly: I think we want to keep it running on the latest 10.x, not necessarily older minor versions. We can also do a matrix of 10.x, 12.x and 14.x if we want.

tuxcanfly: Demo for the matrix: https://github.com/tuxcanfly/hsd/pull/21

I think latest versions make sense because there might be security or other updates which we might want to support.

tuxcanfly: While on the topic of matrix, we should checkout running across OS'es like macos and windows. Seems like GitHub supports these: 

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions

Also, maybe we should be running `test-browser` for browser tests across each matrix item.

It adds to the build-time but would prevent recurring cross-platform bugs we encounter (windows).

pinheadmz: The nodejs matrix is super cool - github says it only took 3m 55s to complete, that's more than acceptable I think. I think OS and browser testing would be great too - if it's too much work, could we just run those extra tests on version-tagging commits or something? (for releases, not just every PR)

tuxcanfly: Sounds good. Even a manual trigger for those wouldn't be too bad.

That could go into another PR since it needs updating the workflow for cross-platform builds.

For now, I suggest we should adopt the node 10.x, 12.x and 14.x build matrix.

pinheadmz: yup agreed, lets add that then try to get ACKs

tuxcanfly: OK, done.

pinheadmz:
Any thoughts on this? I don't think I realized there was one `bin` file that isn't javascript and would throw the linter. I don't mind the warning, but this I suppose is why the `.eslintfiles` existed in the first place

```
$ npm run lint

> hsd@2.1.5 lint /Users/matthewzipkin/Desktop/work/hsd
> eslint bin/* etc/* .


/Users/matthewzipkin/Desktop/work/hsd/bin/hsw
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override

✖ 1 problem (0 errors, 1 warning)

```

tuxcanfly:
Hmm https://github.com/eslint/eslint/issues/5623

pinheadmz:
oh nice, ` eslint bin/* etc/* . --quiet` quiet flag!

tuxcanfly:
Yeah but it might suppress other useful warnings :/

pinheadmz:
Ok cool. At this point, I'm gonna stop painting the bike shed. Running this branch with a few dummy PRs here: https://github.com/pinheadmz/hsd/actions just to take one last look then I think I'm good. Would be nice to get an ACK from  @tynes or @chjj before we switch over the developer flow.

package.json
```diff
@@ -62,14 +63,12 @@
     "hsw-cli": "./bin/hsw-cli"
   },
   "scripts": {
-    "lint": "eslint $(cat .eslintfiles) || exit 0",
-    "lint-ci": "eslint $(cat .eslintfiles)",
+    "lint": "eslint bin/* etc/* .",
```
tynes, line 66: Why moving away from the `.eslintfiles` approach?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Install
+      run: sudo apt-get install -y libunbound-dev |
+           npm install nyc coveralls
+
+    - name: Lint
+      run: npm run lint
+
+    - name: Test
+      run: node_modules/.bin/nyc --reporter lcov node_modules/.bin/bmocha -- --reporter spec test/*.js
```
tynes, line 32: Maybe this code coverage command could be turned into an npm script so that it can also be ran locally?

package.json
tuxcanfly: Actually this was giving trouble. Restored `.eslintfiles` @pinheadmz

.github/workflows/node.js.yml
tuxcanfly: So, the thinking was that this is purely CI stuff.  Another reason is that we don't really have a pure-dep fork of `nyc` or `istanbul`. Adding those to devDeps would add a ton of unaudited packages to the dependency graph.

tynes: Previously, `eslint` wasn't in the dev dependencies but a script existed that referenced it. I'm not super opinionated on needing this, I just think having the ability to look into code coverage locally is a nice to have.

tuxcanfly: OK, restored the `test-ci` script.

tuxcanfly:
OK should be good to go after a squash @pinheadmz @tynes

pinheadmz:
Just wanted to note this here, since one of my tests failed by timeout (mempool ancestors, which generates and signs >50 transactions):

https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

> 2-core CPU
> 7 GB of RAM memory
> 14 GB of SSD disk space

Since its 2 core instead of 4 (like my laptop, for example) I think that will affect the number of workers available and we can expect some tests to timeout still. We may end up bumping up timeouts on some tests throughout the testing suite

package.json
```diff
@@ -62,14 +63,13 @@
...
-    "lint": "eslint $(cat .eslintfiles) || exit 0",
-    "lint-ci": "eslint $(cat .eslintfiles)",
+    "lint": "eslint $(cat .eslintfiles)",
     "lint-file": "eslint",
     "test": "bmocha --reporter spec test/*.js",
     "test-browser": "NODE_BACKEND=js bmocha --reporter spec test/*.js",
     "test-file": "bmocha --reporter spec",
     "test-file-browser": "NODE_BACKEND=js bmocha --reporter spec",
-    "test-ci": "istanbul cover --report lcovonly node_modules/.bin/bmocha -- --reporter spec test/*.js"
+    "test-ci": "node_modules/.bin/nyc --reporter lcov node_modules/.bin/bmocha -- --reporter spec test/*.js"
```
pinheadmz, line 72: Just a note, this can be abbreviated since `npm run ...` adds `node_modules/.bin/*` to the environment $PATH when running commands: https://docs.npmjs.com/cli/run-script 

but I agree enough bike shedding on this one for now

tuxcanfly:
@pinheadmz squashed, also fixed the nit around `node_modules/.bin` being no longer needed.

PR #488, opened 2020-08-07
update wallet http test to use hs-client

### LINKED ISSUE #486: http tests still use client.post() when abstract methods are now available
See https://github.com/handshake-org/hsd/pull/479#discussion_r464773202

ca98am79:
fixes #486


tynes:
I can confirm that the tests all pass locally.

PR #495, opened 2020-08-12
node rpc: scantxoutset

tynes:
This is a simple implementation of `scantxoutset`. This RPC returns a list of the outputs that match the output descriptor.

It currently works with the output descriptor types:

- `wpk`
- `wsh`
- `addr`
- `opreturn`

Some descriptor types defined in https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md do not make sense for Handshake. It should be possible to add `multi` and `sortedmulti` with the current implementation.

The `Descriptor` class should be reimplemented and more thoroughly tested before this should be considered ready for a merge. I think a regex based approach for parsing the descriptor would be more flexible. The xpub stuff would be a nice to have. 

Example usage:

```
hsd-rpc --timeout 100000 scantxoutset 'addr(hs1qnryzj7n8awq7cd398q5t2cs6vqd6yv525y2c4l)'
```

### DIFF
--- a/lib/utils/descriptor.js
+++ b/lib/utils/descriptor.js
@@ -18,8 +18,6 @@ const typeCache = new Set([
   'wpkh',
   'wsh',
   'addr',
-  'multi',
-  'sortedmulti',
   'opreturn'
 ]);
 
--- a/lib/utils/descriptor.js
+++ b/lib/utils/descriptor.js
@@ -1,15 +1,27 @@
 /**
- *
+ * descriptor.js - Output descriptor for hsd
+ * Copyright (c) 2020, The Handshake Developers (MIT License).
+ * Copyright (c) 2020, Mark Tyneway (MIT License).
+ * https://github.com/handshake-org/hsd
  */
 
+'use strict';
+
 const assert = require('bsert');
-const Script = require('../script/script')
+const Script = require('../script/script');
 const secp256k1 = require('bcrypto/lib/secp256k1');
 const Address = require('../primitives/address');
 const {BufferSet} = require('buffer-map');
 
 // all are top level
-const typeCache = new Set(['wpkh', 'wsh', 'addr', 'multi', 'sortedmulti', 'opreturn']);
+const typeCache = new Set([
+  'wpkh',
+  'wsh',
+  'addr',
+  'multi',
+  'sortedmulti',
+  'opreturn'
+]);
 
 class Descriptor {
   constructor(descriptor) {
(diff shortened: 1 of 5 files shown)


test/descriptor-test.js
```diff
@@ -0,0 +1,63 @@
...
+    const address = new Address();
+    const str = `addr(${address.toString()})`;
+
+    const desc = Descriptor.fromString(str);
+    assert.equal(desc.descriptor, str);
+    assert(desc.test(address));
+  });
+
+  it('should create opreturn', () => {
+    const data = 'Bitcoin & BitDNS can be used separately'; // --Satoshi
```
pinheadmz, line 55: hahahaha put this on a hoodie.

lib/utils/descriptor.js
```diff
@@ -0,0 +1,115 @@
...
+const typeCache = new Set([
+  'wpkh',
+  'wsh',
+  'addr',
+  'opreturn'
+]);
+
+class Descriptor {
+  constructor(descriptor) {
+    this.descriptor = descriptor || '';
```
pinheadmz, line 26: Does this syntax just cast to string?

tynes: If `descriptor` is falsy, `null` or `undefined`, then assign the `''` to `this.descriptor`. It is expected that `descriptor` is a string.

pinheadmz: ahhhhhh gotcha. would it be more hsd style like

```
this.descriptor = '';

if (descriptor != null) {
  assert(typeof descriptor === 'String');
  this.descriptor = descriptor;
}
```
etc?

tynes: I skipped implementing the `fromOptions` pattern because it felt kind of verbose.

You can see other places that lack it in the codebase, although this could be for performance reasons since adding extra methods will increase the size of the object in memory.

https://github.com/handshake-org/hsd/blob/d1b42a6a50332ab64de4750297b468de76b05b76/lib/net/packets.js#L772-L785

This is not a performance critical piece of code, so I could follow the pattern if people have a strong opinion on it.

tynes:
@pinheadmz bought up a good point, we could rename `opreturn` to `nulldata`. I'm not super opinionated on this. `opreturn` is not part of the Bitcoin spec, you would have to use `hex` to get opreturn outputs on Bitcoin. I left out `hex` since there are no bare scripts, but now that I think about it, `hex` in the case of handshake could be `version || data`. Then you could use it to get opreturn outputs when the version is 31.  The `opreturn` descriptor can accept ascii or hex, depending on if there is a 0x prefix.

PR #496, opened 2020-08-14
wallet-rpc: derive the receive address by derivation index

viteshan:
derive receive address by wallet-rpc

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -132,6 +132,7 @@ class RPC extends RPCBase {
     this.add('getaddressinfo', this.getAddressInfo);
     this.add('getbalance', this.getBalance);
     this.add('getnewaddress', this.getNewAddress);
+    this.add('derivereceive', this.deriveReceive);
     this.add('getrawchangeaddress', this.getRawChangeAddress);
     this.add('getreceivedbyaccount', this.getReceivedByAccount);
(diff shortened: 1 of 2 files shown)


pinheadmz:
This was discussed in https://github.com/bcoin-org/bcoin/pull/560 and the discussion was around if it's dangerous to allow users to generate addresses that are not saved by the wallet. If you request address 100 but the internal wallet depth is only 10, you won't receive transactions at that address. Can you explain your use case a bit? Maybe we can reopen the discussion. I think in the bcoin PR I also had a "force" flag, mainly the wallet would not return an address below its current depth.

viteshan:
in my case, i need generate some addresses for user deposit. derive the address first, then import it.
newaddress can’t get a deterministic result, for me.

pinheadmz:
> newaddress can’t get a deterministic result, for me.

Is there a bug then?

Have you seen the API docs? https://hsd-dev.org/api-docs/#generate-receiving-address

Also a good setup for BIP44 wallets is to create a new account for each user. When the wallet receives a deposit it will tell you which account was deposited to and your application DB can credit that user's account.

nodar-chkuaselidze:
If we want to provide this feature, considering users are aware of the risks - we could use bitcoin compatible utility functions as mentioned in https://github.com/bcoin-org/bcoin/pull/560:
BTC: https://github.com/bitcoin/bitcoin/pull/14667

NACK.

Issue #481, opened 2020-07-25
OPEN covenants have 0 value


pinheadmz:
As of July 5, block height 22517, the Handshake UTXO set included 152,153 OPEN covenants all with an output value of `0`.

Here's a UTXO from one of the very first auctions ever (block 4032) that will never be purged from the utxo set:

```
{
  "version": 0,
  "height": 4032,
  "value": 0,
  "address": "hs1qflkm3n802zfzt0j9337d9hzxk4khctzj26t7sc",
  "covenant": {
    "type": 2,
    "action": "OPEN",
    "items": [
      "a8b64e7d7695f94eed27ec69ce9e02ed3f92784f38a9619abc25006a0a1d5571",
      "00000000",
      "656e64616e676572"
    ]
  },
  "coinbase": false,
  "hash": "1eb102b67b33463b4e63baac897cca6bbeef45a4ced538a77d1d35ab9feb93af",
  "index": 0
}

```

This is clearly a huge source of bloat. It consumes full node resources, and causes issues in the wallet by loading the txdb with coins it will never spend (#478 -- although there are several other improvements we need to make to fix that issue completely).

So, let's brainstorm some fixes.

One idea is to make OPEN outputs officially unspendable. REVOKE outputs are already unspendable and so are outputs to "nulldata" addresses (Handshake's segwit-only version of OP_RETURN). This would have to be enforced by a soft fork, and then OPENs would no longer be added to the utxo set. I did a quick grep of the wallet module for "spendable" and didn't find anything -- I hope the wallet knows which of its coins are unspendable! **This actually might be another issue** and we should make sure the wallet does not try to spend REVOKE outputs, and also prevent a wallet from thinking it can spend from a null address even if that address has been manually imported.

Another solution is to change the wallet behavior to use the OPEN covenant as a change output. This will reduce blockchain bloat and actually make these coins useful to the wallet. A typical open TX would then just have one input (any value) and one output (the OPEN) with the miner fee subtracted. This brings up kind of an interesting sub-issue though: should OPEN outputs in this scheme pay to a receive address or a change address (see BIP44)? Change address makes some sense, but then (for example) Bob Wallet would need to be updated to look for this, because otherwise it ignores change outputs and the user wouldn't see that a tx was an OPEN. Similarly, if we use receive addresses, Bob Wallet will need an update to recognize the tx as a self-send (I do not think it does this yet) otherwise the portfolio screen will appear to have an "incoming" OPEN tx that adds value to the wallet.

[...]

PR #498, opened 2020-08-14
[POLICY] address: fix isUnknown() to allow nullData and version 0 addresses

pinheadmz:
Fixes a bug in `Address.isUnknown()` that mis-identifies nulldata addresses as non-standard and mis-identifies all other witness program versions as standard (version 0 addresses are obviously OK)

Note that Bitcoin Core has allowed more flexibility in relaying unknown witness versions: https://github.com/bitcoin/bitcoin/pull/15846 But I don't think we need to worry about that yet.

Unfortunately before nulldata addresses can be used for practical purposes like [provably prunable outputs](https://github.com/handshake-org/hsd/issues/481#issuecomment-664658894), this patch needs to be released and widely deployed, especially to miners.

### DIFF
--- a/lib/primitives/address.js
+++ b/lib/primitives/address.js
@@ -371,9 +371,9 @@ class Address extends bio.Struct {
       case 0:
         return this.hash.length !== 20 && this.hash.length !== 32;
       case 31:
-        return true;
+        return false;
     }
-    return false;
+    return true;
   }
 
   /**
--- a/test/address-test.js
+++ b/test/address-test.js
@@ -14,4 +14,39 @@ describe('Address', function() {
     const expect = 'hs1qd42hrldu5yqee58se4uj6xctm7nk28r70e84vx';
     assert.strictEqual(addr.toString('main'), expect);
   });
+
+  it('should check standardness of address', () => {
+    const addr = new Address();
+
+    // Standard p2wpkh
+    addr.version = 0;
+    addr.hash = Buffer.alloc(20);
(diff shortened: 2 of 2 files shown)


lib/primitives/address.js
```diff
@@ -371,9 +371,9 @@ class Address extends bio.Struct {
       case 0:
         return this.hash.length !== 20 && this.hash.length !== 32;
       case 31:
-        return true;
+        return false;
     }
-    return false;
+    return true;
   }
```
tynes, line 377: Reading this fix, it semantically makes sense. Good catch.

test/address-test.js
```diff
@@ -14,4 +14,39 @@ describe('Address', function() {
...
+    addr.version = 0;
+    addr.hash = Buffer.alloc(32);
+    assert(!addr.isUnknown());
+
+    // nullData, any valid length
+    for (let i = 2; i <= 40; i++) {
+      addr.version = 31;
+      addr.hash = Buffer.alloc(i);
+      assert(!addr.isUnknown());
+    }
```
tynes, line 36: The way that you are constructing the nulldata output, you could technically create one with an invalid length and still have it pass here. I think that is fine since invalid outputs are considered unknown.

pinheadmz: True, there is no specified data length for witness program version 31, and so that is outside the scope of `isUnknown()`. Notice however that consensus limits data to `2 >= length >= 40` in `Address.isValid()` and that's why I chose those values for the test.

Issue #478, opened 2020-07-20
Coin select slowdown


nulven:
Sending transactions, particularly opens, seems to have slowed down significantly. After investigating a bit further it seems that `wallet.fill` is the root cause.

Here’s a wallet I created to test. It has ~16 thousand, 1 coin none txos, sent over about 160 blocks. In my testing, I was getting open times of around 150 ms and transaction times of 50 ms.

If you want to test yourself, here’s the steps I took to see the time to send an open.

In `~/.hsd/regtest`
`unzip /path/to/file/wallet.zip -d wallet`

[wallet.zip](https://github.com/handshake-org/hsd/files/4949294/wallet.zip)
Change lines 1639-1648 in `hsd/lib/wallet/wallet.js` to the following to log the time.
```
async sendOpen(name, force, options) {
    console.time('SEND OPEN');
   const unlock = await this.fundLock.lock();
   try {
       return await this._sendOpen(name, force, options);
   } finally {
       unlock();
       console.timeEnd('SEND OPEN');
   }
}
```

Run the node
`./bin/hsd --network=regtest --index-address=false --index-tx=false`

Grind a new name, send the open, and check the logs for ‘SEND OPEN’
`./bin/hsd-cli rpc --network=regtest grindname 10`
`./bin/hsw-cli rpc sendopen --network=regtest [NAME]`

pinheadmz:
How did you generate the 16000 utxo? If it was all by block generation you might have this issue:

https://github.com/bcoin-org/bcoin/issues/882

A useful check would be the values of the utxos. If they are mostly 0 or below dust I think we will have an easy fix

turbomaze:
@pinheadmz thanks for linking to that issue, the thread there is very relevant. We initially tested this by mining lots of blocks, and we noticed that mining into the wallet we were testing seemed to exacerbate the issue. However, to simplify the test case, we switched to sending many tiny txs from another wallet. Not sure if it's below dust value or not @nulven.

This issue is exacerbated in wallets with actual auction txos. I wonder if all of the 0 value open txos (and 0 value reveal/redeems for bids with 0 value or 0 blind??) cause a similar issue. What is the easy fix you have in mind?

pinheadmz:
So I tried using your wallet.zip but since the chain data is missing the wallet had issues.

[...]

PR #499, opened 2020-08-14
Send OPENs to nulldata address

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/481
Chips away at https://github.com/handshake-org/hsd/issues/478
~Requires network deployment (especially miners) of https://github.com/handshake-org/hsd/pull/498~ ✅ 

Motivation is partly explained in #481 but here are the pros and cons:

### PROs

- reduce size of OPEN covenants by 18 bytes
- OPEN covenants do not bloat the UTXO set with 0-value outputs (technically spendable but why would you?)
- OPEN covenants do not bloat the wallet's UTXO set, which adds to the burden of the coin selector
- ~Prevents an annoyance where an attacker can send an OPEN to someone else's wallet, forcing that wallet to watch and collect all bids for that auction, whether or not the unwilling recipient cares about the name (also adds a 0-value UTXO to the victim's wallet, [because OPENs aren't checked for dust](https://github.com/handshake-org/hsd/blob/d1b42a6a50332ab64de4750297b468de76b05b76/lib/primitives/covenant.js#L352-L366))~ <- This will require a second patch in txdb if we decide it is a big enough issue.

### CONs

- We must wait until a majority of the relay nodes and miners have updated with #498 which fixes a bug in relay policy.
 - Because there is no address associated with OPEN, applications that used to use the address to detect when a wallet has sent an OPEN must be refactored to instead check the inputs to the TX ("Did my wallet send this tx? Then OK, tell the user they have OPENed an auction" -- as opposed to "This OPEN is being sent to my own address -- [example in Bob Wallet](https://github.com/kyokan/bob-wallet/blob/master/app/ducks/walletActions.js#L253-L255)")

### Review

commit 4b75f8aa8e45d325f8b7b6501767ff98fceebda9  address: don't restrict data length to 20 or 32
  - I'm not sure if we ever needed this check in the first place, opened an issue upstream: https://github.com/bcoin-org/bcoin/issues/989

commit 0301f47e6163b55712c2bd97af4adb54790bfed7  wallet: do not remove wallet-name-map when disconnecting a block
  - At the beginning of a rescan, a lot of wallet state is thrown out (not transactions themselves, but whether or not they are confirmed and in what block). We also clear out the name map, which is a simple structure that stores which wallet IDs are "watching" a name. I do not think we need to reset this, and removed it here.

commit 4075ae856d2195ad4c4b9f8c98debd50f1cb25a6  wallet: send OPENs to unspendable nullData address
  - this is the fix, comments should explain the patch. Luckily we have `importName()` now! https://github.com/handshake-org/hsd/pull/408

commit 8edaea67e01a3e18856b97384ace398de4a7ce1c  test: OPEN outputs no longer inserted as coins into txdb
  - notice how certain tests had to be adjusted because the OPEN output is no longer stored as a coin in the wallet

commit c1932472d1093f793e69db946161be595f6b5a32  pkg: update CHANGELOG
  - hopefully this explains the change well enough to developers using hsd in their own wallet applications

### DIFF
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -1091,9 +1091,6 @@ class Wallet extends EventEmitter {
   async _importName(name) {
     const nameHash = rules.hashName(name);
 
-    if (await this.txdb.hasNameState(nameHash))
-      throw new Error('Name already exists.');
-
     const b = this.db.batch();
     await this.wdb.addNameMap(b, nameHash, this.wid);
     await b.write();
@@ -1561,7 +1558,17 @@ class Wallet extends EventEmitter {
     if (start !== 0 && start !== height)
       throw new Error('Name is already opening.');
 
-    const addr = await this.receiveAddress(acct);
+    // Send 0-value output to provably unspendable address:
+    // hs1lqqqqhuxwgy (mainnet) is the smallest valid nulldata address
+    // and represents witness version 31 folowed by data of 0x0000.
+    // Unspendable outputs do not bloat the UTXO set or the wallet's coin set.
+    const addr = Address.fromNulldata(Buffer.alloc(2));
+
+    // Since this address is not in our wallet, we must "manually" add the name
+    // to the wallet database's name map. Normally the wallet would do this in
+    // txdb.js connectNames() when a TX is inserted into the DB, and the wallet
+    // recognizes the output address.
+    await this.importName(name);
 
     const output = new Output();
     output.address = addr;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@ REGISTERing. If unpatched software was used to CLAIM a name already, that wallet
 database is irreversibly corrupted and must be replaced.
 See https://github.com/handshake-org/hsd/issues/454
 
+- By default the wallet will now send OPEN covenants to an unspendable `nullData`
+address (witness program version `31`, with minimum required data `0x0000`). This
+will prevent 0-value coins from bloating the UTXO set in chainDB and walletDB.
+It will also reduce the size of OPEN outputs in the blockchain by 18 bytes.
+Applications that relied on the output address to detect when a user has sent an
+OPEN must now rely on the input coins to an OPEN transaction for this purpose.
+
 ## v2.0.0
 
 ### Wallet API changes
--- a/lib/primitives/address.js
+++ b/lib/primitives/address.js
@@ -542,8 +542,6 @@ class Address extends bio.Struct {
       throw new Error('Object is not an address.');
 
     if (Buffer.isBuffer(data)) {
-      if (data.length !== 20 && data.length !== 32)
-        throw new Error('Object is not an address.');
       return data;
     }
 
--- a/lib/wallet/txdb.js
+++ b/lib/wallet/txdb.js
@@ -2155,7 +2155,6 @@ class TXDB {
       ns.applyState(delta);
 
       if (ns.isNull()) {
-        await this.removeNameMap(b, nameHash);
         b.del(layout.A.encode(nameHash));
       } else {
         b.put(layout.A.encode(nameHash), ns.encode());
--- a/test/wallet-importname-test.js
+++ b/test/wallet-importname-test.js
@@ -100,11 +100,8 @@ describe('Wallet Import Name', function() {
     assert(ns4 === null);
   });
 
-  it('should not re-import an existing name', async () => {
-    await assert.rejects(
-      alice.importName(name),
-      {message: 'Name already exists.'}
-    );
+  it('should ignore re-importing existing name', async () => {
+    await alice.importName(name);
   });
 
   it('should bid on names from Alice\'s wallet', async () => {
@@ -223,13 +220,9 @@ describe('Wallet Import Name', function() {
         assert(!bid.own);
     });
 
-    it('should not re-import name', async () => {
+    it('should ignore re-importing name', async () => {
       await wclient.execute('selectwallet', ['charlie']);
-
-      await assert.rejects(
-        wclient.execute('importname', [name, 0]),
-        {message: 'Name already exists.'}
-      );
+      await wclient.execute('importname', [name, 0]);
     });
   });
 });
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -1962,7 +1962,7 @@ describe('Wallet', function() {
       // Check
       let bal = await wallet.getBalance();
       assert.strictEqual(bal.tx, 2);
-      assert.strictEqual(bal.coin, 2);
+      assert.strictEqual(bal.coin, 1); // OPENs are sent to a null address
       assert.strictEqual(bal.confirmed, 10e6);
       assert.strictEqual(bal.unconfirmed, 10e6 - fee);
       assert.strictEqual(bal.ulocked, 0);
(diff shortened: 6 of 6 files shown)


pinheadmz:
@turbomaze

turbomaze:
Thanks zippy! This is a great change

pinheadmz:
Back when I wrote this PR I tried to test it so I could add an example to the description here. I didn't realize in advance that [nulldata addresses were accidentally classified as nonstandard](https://github.com/handshake-org/hsd/pull/498) and would be rejected by all network nodes and not relayed. I actually set my own node to accept non standard transactions to complete the test. That transaction (an OPEN for the name `nulldata`) was finally [confirmed in a block last week](https://hnscan.com/tx/12f0b0cd96c1ffbcdda145fb3a3f8580b756805ba9c47836e5b018f857096dd7). The auction opened normally, received one bid and is currently in the reveal phase.

Because this was (I think) the first nulldata address on the chain, HNScan.com actually crashed on this block because of a an error in its backend when parsing such an address (unrelated to hsd or this PR).

Anyway - the test basically worked. We can open auctions with nulldata addresses and apparently there are enough v2.2 nodes on the network (including miners) for such transactions to get confirmed. I hesitate to merge this PR too soon though because relay may still be quite choppy. Here's the peer version breakdown from easyhandshake today:

```
Map {
  '/hnsd:0.0.0/' => 3,
  '/hsd:2.0.1/' => 2,
  '/hsd:2.0.2/' => 21,
  '/hsd:2.0.3/' => 2,
  '/hsd:2.1.1/' => 3,
  '/hsd:2.1.2/' => 19,
  '/hsd:2.1.3/' => 9,
  '/hsd:2.1.5/' => 99,
  '/hsd:2.2.0/' => 42
}
```

and the TX that broke everything :-)

```
{
  "hash": "12f0b0cd96c1ffbcdda145fb3a3f8580b756805ba9c47836e5b018f857096dd7",
  "witnessHash": "1b6dc72c33face502ea341092b81d1db7bb46c739eabce81ab62c824cde08e73",
  "fee": 16900,
  "rate": 100000,
  "mtime": 1600192815,
  "height": 33058,
  "block": "000000000000000186256e2aa609ec3782076eb784e92d3264c9b22579690e99",
  "time": 1600178873,
  "index": 1,
  "version": 0,
  "inputs": [
    {
      "prevout": {
        "hash": "229fec21746714b3ec2c3f28ac8d1843f3f864f4bf6b67028e8c80f3c883178a",
        "index": 1
      },
      "witness": [
        "ab8e2e20da0fe6d7ef0e8ebd64c1201f6f3691da643e226fc88e7b7fadcfd2234eabb53f0a34fd30f82d0429ef7c57e85d60096b7e023ac7df6a068465f048fe01",
        "02d61a5afbd7057d2db6128594c4d06bfae6e7d85ad1aca5ea1aa7e46d1762989f"
      ],
      "sequence": 4294967295,
      "coin": {
        "version": 0,
        "height": 11685,
        "value": 6281800,
        "address": "hs1qxmsxlz4465uupqdtxrzq8y6p24633ht5r97smu",
        "covenant": {
          "type": 0,
          "action": "NONE",
          "items": []
        },
        "coinbase": false
      }
    }
  ],
  "outputs": [
    {
      "value": 0,
      "address": "hs1lqqqqhuxwgy",
      "covenant": {
        "type": 2,
        "action": "OPEN",
        "items": [
          "109e1992c74d0fcabbc6f3774268f967022c4884fad33b6853341792183329b9",
          "00000000",
          "6e756c6c64617461"
        ]
      }
    },
    {
      "value": 6264900,
      "address": "hs1qlje6h3zzy8syqu69p9jx9lh5hfqngmse079uf4",
      "covenant": {
        "type": 0,
        "action": "NONE",
        "items": []
      }
    }
  ],
  "locktime": 0,
  "hex": "0000000001229fec21746714b3ec2c3f28ac8d1843f3f864f4bf6b67028e8c80f3c883178a01000000ffffffff0200000000000000001f020000020320109e1992c74d0fcabbc6f3774268f967022c4884fad33b6853341792183329b90400000000086e756c6c6461746144985f00000000000014fcb3abc44221e0407345096462fef4ba41346e190000000000000241ab8e2e20da0fe6d7ef0e8ebd64c1201f6f3691da643e226fc88e7b7fadcfd2234eabb53f0a34fd30f82d0429ef7c57e85d60096b7e023ac7df6a068465f048fe012102d61a5afbd7057d2db6128594c4d06bfae6e7d85ad1aca5ea1aa7e46d1762989f",
  "confirmations": 855
}

```

lib/wallet/wallet.js
```diff
@@ -1561,7 +1558,17 @@ class Wallet extends EventEmitter {
...
+    // hs1lqqqqhuxwgy (mainnet) is the smallest valid nulldata address
+    // and represents witness version 31 folowed by data of 0x0000.
+    // Unspendable outputs do not bloat the UTXO set or the wallet's coin set.
+    const addr = Address.fromNulldata(Buffer.alloc(2));
+
+    // Since this address is not in our wallet, we must "manually" add the name
+    // to the wallet database's name map. Normally the wallet would do this in
+    // txdb.js connectNames() when a TX is inserted into the DB, and the wallet
+    // recognizes the output address.
+    await this.importName(name);
```
nodar-chkuaselidze, line 1571: Maybe I missed it,

but does not this need to change rescan logic as well, so wallets can recover the OPENs they have sent?

pinheadmz: `importName()` will add the name to the wallet's name map, which means the wallet will "track" all auction activity like bids etc even if the user did not bid themselves (this is weird, but its how the wallet works already). This will persist for rescans as well, especially because of this change: https://github.com/handshake-org/hsd/pull/499/commits/0301f47e6163b55712c2bd97af4adb54790bfed7

Since the wallet _sent_ the open, it should recover the TX normally because its own keys were used for the inputs and change.

OPENs are weird things - they are just a marker for the chain and wallet (this is why they have 0 value), announcing to the network a namehash and its preimage. JJ told me OPENs weren't even added to the system until late in development because they didn't seem necessary, but definitely made wallet logic a lot easier.

We have a few rescan tests now in hsd, I'll rebase this PR and we can double check

nodar-chkuaselidze: sure,

I was wondering about importing wallet into fresh wallet instance.
I can test after rebase.

pinheadmz: > I was wondering about importing wallet into fresh wallet instance.

Yes! this needs to be patched. Its' going to be in `txdb.insert()` or `txdb.confirm()` I think

pinheadmz:
Closing this for now, I think the approach is a bit flawed. First of all, the bloat is not that bad. An OPEN UTXO is at most only about 130 bytes, meaning this year with over a million names on chain we're only "wasting" 130 MB. Second, sending to nulldata address just to make an output unspendable seems a bit hacky. What we should do is just... make 0-value OPENs unspendable! Whether or not that ever becomes a consensus rule, it can certainly be factored in to the wallet so we don't add these coins to the txdb.

What makes the most sense is to refactor the coinselector in the wallet, and refactor the layout of txdb so coins can be pulled out by value (instead of randomly by hash and then filtered by value and spent-ness)

Falci:
> make 0-value OPENs unspendable

What would happen to the "0-value OPENs" already spent?

pinheadmz:
> > make 0-value OPENs unspendable
> 
> What would happen to the "0-value OPENs" already spent, when a new node tries to sync?

This is why soft forks have activation height!

PR #500, opened 2020-08-18
test: demonstrate interactive name swaps

pinheadmz:
This test isn't designed intentionally to increase test coverage but more to document one method of atomically swapping HNS coins for a name by using an interactive protocol and partially-signed transactions. We don't need to merge it but it may prevent some edge case regression in the future.

The flow goes like this:

- Alice owns a name and agrees to sell it to Bob for a certain price (they communicate any way possible, not on-chain)
- Bob sends Alice an address to receive the name
- Alice signs and broadcasts a TRANSFER covenant committing to Bob's address
  - at this point, Alice still unilaterally owns the name.
  - Bob can see the TX on the blockchain (so can everyone) and confirm his address is in the covenant
  - at any time before the FINALIZE is broadcast, Alice can CANCEL the transfer.
    - Bob can see this CANCEL on chain (back to step 1)
    - no money or name is lost by either party if Alice decides to CANCEL, just the two mining fees so far spent by Alice.
- After 288 blocks a FINALIZE is valid, and is constructed cooperatively by both parties
  - Alice can complete her end of the protocol at any time, but Bob can not broadcast the FINALIZE until after 288 blocks
  - Alice may decide to CANCEL if Bob doesn't complete the FINALIZE within a certain time
    - this is not enforced by the blockchain, it's just something Alice can do, and should warn Bob about it in step 1
  - Alice constructs and partially signs a FINALIZE that sends value to her own receive address, and consumes the TRANSFER covenant as an input. This requires `SIGHASH_ANYONECANPAY | SIGHASH_SINGLEREVERSE` so that Bob can still add funds.
  - Alice sends the serialized, partially-signed FINALIZE to Bob as a hex string
  - Bob decodes the blob, verifies what he can
    - his address is in the TRANSFER covenant and the FINALIZE output is valid
    - Alice's signature is valid, and the value output to her address is the agreed price
  - Bob adds funds to the TX inputs and adds a change output if necessary
  - Bob signs his inputs with `SIGHASH_ALL`
  - Bob broadcasts the final transaction

### DIFF
--- a/test/interactive-swap-test.js
+++ b/test/interactive-swap-test.js
@@ -1,0 +1,304 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const Network = require('../lib/protocol/network');
+const FullNode = require('../lib/node/fullnode');
+const MTX = require('../lib/primitives/mtx');
+const Address = require('../lib/primitives/address');
+const Output = require('../lib/primitives/output');
+const Script = require('../lib/script/script');
+const rules = require('../lib/covenants/rules');
+const {types} = rules;
+const {Resource} = require('../lib/dns/resource');
+const {WalletClient} = require('hs-client');
+
+const network = Network.get('regtest');
+
+const ports = {
+  p2p: 14331,
+  node: 14332,
+  wallet: 14333
+};
+const node = new FullNode({
+  memory: true,
+  network: 'regtest',
+  plugins: [require('../lib/wallet/plugin')],
+  env: {
+    'HSD_WALLET_HTTP_PORT': ports.wallet.toString()
+  }
+});
+
+const wclient = new WalletClient({
+  port: ports.wallet
+});
+
+const {wdb} = node.require('walletdb');
+
+let alice, bob, aliceReceive, bobReceive;
+let aliceOriginalBalance, bobOriginalBalance, bobFee;
+
+// These are data that will be communicated between Alice and Bob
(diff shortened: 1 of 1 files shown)


pinheadmz:
Also worth noting that this protocol can be executed entirely with RPC commands, but it gets messy (you need to do coin selection on the client side from `listunspent`, get the private key with `dumpprivkey`, and pass it to the node rpc `signrawtransaction`)

tynes:
Not sure if this belongs in this repo, although it is really cool.

test/interactive-swap-test.js
```diff
@@ -0,0 +1,304 @@
...
+    assert(json.state === 'CLOSED');
+  });
+
+  it('should not be able to send a TRANSFER before REGISTER', async () => {
+    assert.rejects(async () => {
+      await alice.sendTransfer(name, bobReceive);
+    }, {
+      message: 'Name must be registered.'
+    });
+  });
```
Falci, line 113: This test doesn't seems to be related with your PoC

pinheadmz: I see your point technically, it does not prove anything about the atomic swap itself. However:

1) you can never have too much test coverage -- any test can catch a regression in a future code update

2) this test in particular is designed to demonstrate a user process. The reason I included this was to indicate to implementers (like Kyokan, see https://github.com/kyokan/bob-wallet/pull/208) that registration must be confirmed before the atomic swap can take place. Otherwise we get users complaining to support that they got a "name must be registered" error when they tried to swap.

chjj:
I'd like to merge it. More tests are always a good thing.

PR #506, opened 2020-08-31
wallet: bump lookahead to 200

### LINKED ISSUE #407: Wallet restore misses transactions due to dense transaction activity: (lookahead / BIP44 gap limit)
When you have lots of names (>100), rescanning the wallet causes names to be lost. Checkout the following screenshot:

Screen Shot 2020-03-26 at 5 48 03 PM

Here, I'm executing a wallet rescan then getting the list of names from the wallet once the rescan completes. (I look at the logs in another terminal to determine when the rescan is done.) As you can see, the number of names in the wallet changes even though the wallet itself hasn't made any new transactions. The number of names increases randomly whenever I rescan then get the names in the wallet.

pinheadmz:
Closes #407 
...but in a lazy way that more kicks the can down the road than anything.
See [this comment](https://github.com/handshake-org/hsd/issues/407#issuecomment-668625559) specifically for details.

TODO:
- [x] run benchmarks.
- [x] fix tests (generate 200 addresses from bitcoin core and maybe verify with one other implementation for test vectors)

# Update:
This was lowered from 1000 to 200 because in the serialized account metadata, the lookahead value is only a single byte:

https://github.com/handshake-org/hsd/blob/7826128ff72c20fdf42c659309dcffea7a28f6cc/lib/wallet/account.js#L747

# benchmarks:

### master branch
`account: ops=1000, time=3.276540621, rate=305.19994`

### PR branch
`account: ops=1000, time=46.609575762, rate=21.45482`

No big surprise there - creating an account makes generating 20x more keys (380 total because of receive & change).
So we've increased account creation time from ~3ms to ~46ms


tynes:
It's nice to see the benchmark for account creation. I'm curious about how it would impact the performance of a rescan. If it's using a bloom filter then it should just be building the bloom filter that takes longer.

As performance becomes an issue, maybe https://github.com/chjj/bthreads/ can be used.

pinheadmz:
Interesting idea to use threads to generate keys in parallel. I know walletDB has workers but I'm not sure if they're used in rescanning. Off the top of my head, that seems like the kind of thing that must be done serially because of the UTXO system.

PR #507, opened 2020-08-31
wallet: add 'spendable_balance' to getwalletinfo RPC

ca98am79:
This enables a user to view their spendable balance using the `getwalletinfo` RPC

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -748,12 +748,14 @@ class RPC extends RPCBase {
 
     const wallet = this.wallet;
     const balance = await wallet.getBalance();
+    const spendableBalance = balance.unconfirmed - balance.ulocked;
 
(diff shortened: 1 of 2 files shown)


nodech:
I am wondering if we should modify this and/or add `getbalancedetails`(or `info`) rpc methods to include balance details which can be used by users however they want. It could be advanced version of `getbalance`.

pinheadmz:
> I am wondering if we should modify this and/or add `getbalancedetails`(or `info`) rpc methods to include balance details which can be used by users however they want. It could be advanced version of `getbalance`.

Bitcoin core has extra details in [getwalletinfo](https://chainquery.com/bitcoin-cli/getwalletinfo) and also has separate rpc `getwalletbalances` which also calculates the immature newly mined coins (see #662)

Issue #505, opened 2020-08-27
getwalletinfo rpc return incorrect data


leehyong:
hello guys,
 When I use `getwalletinfo` rpc to query data, I suspect that *balance* is  equal to *unconfirmed_balance* according to the returned data, which may be wrong.
So I lookup the code about `getwalletinfo`, which is as follows in the `getWalletInfo` function body:
```
return {
      walletid: wallet.id,
      walletversion: 6,
      balance: Amount.coin(balance.unconfirmed, true),
      unconfirmed_balance: Amount.coin(balance.unconfirmed, true),
      txcount: balance.tx,
      keypoololdest: 0,
      keypoolsize: 0,
      unlocked_until: wallet.master.until,
      paytxfee: Amount.coin(this.wdb.feeRate, true)
    };
```
Is this a bug ?

pinheadmz:
So, the hsd RPC API is a port from bcoin, which was based on Bitcoin Core's RPC API from several years ago. A lot of the quirks in the RPC are due to managing compatibility with Bitcoin Core from that era. With that in mind, consider that bcoin and hsd have an unusual definition for "unconfirmed" balance. From [the API docs](https://hsd-dev.org/api-docs/#getunconfirmedbalance):

> In hsd balances, confirmed refers to the total balance of coins confirmed in the blockchain. unconfirmed refers to that total IN ADDITION to any transactions still unconfirmed in the mempool. Another way to think about it is your unconfirmed balance is the FUTURE total value of your wallet after everything is confirmed.

So I think this was actually by design to conform the bcoin definition of "unconfirmed" to sort of work with what is expected from the Bitcoin Core RPC.

I'm not opposed to changing this RPC output since we don't need to match anything from Bitcoin in hsd. Feel free to open a pull request and we can try to get community feedback on how this data should be returned.

ca98am79:
This may be a separate discussion, but it would be nice to add `available_balance` as both `balance` and `unconfirmed_balance` take into account all HNS that is burned / locked up and unavailable to use in any way.

ca98am79:
@pinheadmz it is by design that these two lines are exactly the same?

```
balance: Amount.coin(balance.unconfirmed, true),
unconfirmed_balance: Amount.coin(balance.unconfirmed, true),
```

it seems redundant

pinheadmz:
[...]

PR #508, opened 2020-08-31
wallet: change getwalletinfo RPC 'balance' to be confirmed balance, i…

ca98am79:
…nstead of repeating unconfirmed balance

see issue #505

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -752,7 +752,7 @@ class RPC extends RPCBase {
     return {
       walletid: wallet.id,
       walletversion: 6,
-      balance: Amount.coin(balance.unconfirmed, true),
+      balance: Amount.coin(balance.confirmed, true),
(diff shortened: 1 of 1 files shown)


lib/wallet/rpc.js
```diff
@@ -752,7 +752,7 @@ class RPC extends RPCBase {
     return {
       walletid: wallet.id,
       walletversion: 6,
-      balance: Amount.coin(balance.unconfirmed, true),
+      balance: Amount.coin(balance.confirmed, true),
```
pinheadmz, line 755: Ok my thinking on this:  Let's not break backwards compatibility by changing what is returned here. We don't want to mess up any application relying on this value always being the "safest" (confirmed) balance.

What I think could be useful is just adding more values. We already have `unconfirmed_balance`, let's just add `confirmed_balance` and `spendable_balance` (combine with https://github.com/handshake-org/hsd/pull/507)

Issue #467, opened 2020-06-03
Proposal: Deep Rescan


pinheadmz:
How wallet `rescan` works:

1. [rollback to specified height](https://github.com/handshake-org/hsd/blob/master/lib/wallet/walletdb.js#L1977-L2001)
    - this means [un-confirming all TX history](https://github.com/handshake-org/hsd/blob/master/lib/wallet/txdb.js#L1420-L1440) from tip backwards to target height
2. rewind and play blockchain back into wallet, re-confirming TXs from step 1 and potentially inserting (and then immediately confirming) new TXs that were not already in the wallet.

So what `rescan` does NOT do is start from a completely empty state with a mnemonic seed and rebuild TX history from scratch. That is typically called a `"recovery".

We've had a few critical txdb issues recently that result in assertion errors:

https://github.com/handshake-org/hsd/issues/454

https://github.com/handshake-org/hsd/pull/464

Users have to recover (not rescan) their wallet because of the corrupt balance. To make matters worse, the fact that a corrupted wallet still lives in the DB (even if its not used anymore) means these assertion errors can not be avoided without completely re-initializing the entire db.

[There is a method to remove wallets from the db](https://github.com/handshake-org/hsd/blob/master/lib/wallet/walletdb.js#L902-L926) but it is not exposed in the API for obvious reasons, and is not allowed to remove the `primary` wallet.

SO -- Should there be a new wallet method called `deep rescan` or an extra argument to `rescan` that does a deep clean of the wallet before replaying the blockchain?

The benefit to the user would be the ability to fix corrupted state _without_ creating a brand new walletDB. The DB would stay in place and the wallet name would stay in place.

pinheadmz:
Another benefit of this over a wallet import-from-seed is that we should be able to keep the database of address hashes and serialized account metadata, so the lookahead issue is mitigated. We would only reset balance and tx history. I think also it should be required that the deep rescan start from genesis, (a) to reduce user options since this is a bit of a nuclear bomb and (b) to ensure that we don't accidentally recover a REVEAL without first recovering the BID, etc.

PR #464, opened 2020-06-01
Adjust txdb locked coins balance for FINALIZE covenants

pinheadmz:
I think we may have overlooked the impact of FINALIZE covenants on locked coins balances in the wallet. They are unusual because it is the only case where locked coins _leave the wallet_. Meaning someone else's locked coin balance goes up. The current txdb functions `lockBalances()` and `unlockBalances()` process covenants received into the wallet, however in this case we must also catch FINALIZE as it is being _sent_, since our own locked coin balance goes down (along with the total balance - name and coins get transferred) 

TODO:
- Evaluate the impact of this change on exiting wallets in use. Do we need another wallet DB migration?
- Will some users have to completely restore wallet from seed and rescan like the [claim-to-register issue?](https://github.com/handshake-org/hsd/issues/454)
- Let's make sure REVOKE and TRANSFER aren't lingering issues as well

According to my covenant statistics script, at this time there are 14 `FINALIZE` covenants confirmed on the blockchain, for a total value of 3,700.500000 HNS. Both the sender and receiver of these names will have inaccurate locked coin balances, that may end up throwing an assertion error depending on future wallet activity.

Similar PRs recently merged affecting txdb locked coins computation:

https://github.com/handshake-org/hsd/pull/438
https://github.com/handshake-org/hsd/pull/387
https://github.com/handshake-org/hsd/pull/262

Issue #454, opened 2020-05-15
Name Claim Registration


tynes:
A bug existed in the wallet that prevented any name that was claimed from the reserved name list from being updated. This means that claimed names have been unable to write DNS records to the chain or claim their locked HNS using the wallet `sendupdate` RPC API or `POST /wallet/:id/update` HTTP API. Note that this does not impact the consensus rules, it was always possible to manually create the transaction and submit it to the blockchain. This solely impacts the wallet's bookkeeping and a failed assertion prevented the transaction from being broadcasted after it was created and signed.

This bug was fixed here https://github.com/handshake-org/hsd/pull/438 but requires the wallet holding the name to be recreated. If you update to the latest codebase with and attempt to register the name without recreating the wallet, you will see an error. This does not mean that your name is lost or that it cannot be registered.

A wallet can be recreated using the mnemonic. If you are participating in any auctions with that wallet, it is recommended that you wait until the auction is over with before recreating the wallet. See discussion here https://github.com/handshake-org/hsd/issues/378.

To get the mnemonic for your wallet, use the command:

```bash
$ hsw-cli master
{
  "encrypted": false,
  "key": {
    "xprivkey": "rprvKE8qsHtkmUxUSNxoQq6feh9FS667F4WtV6Rd4auVyrWg4ZnSKudy2iquTygUP2FEFbNdee56pbHcGRn3VXjiANCqZAENcmhDFVepYviExDCd"
  },
  "mnemonic": {
    "bits": 256,
    "language": "english",
    "entropy": "fc3045263cc3b6a5b46ac61b02d30900bb250fe0fa8bdbb447e1cba3a00c6d17",
    "phrase": "wire lion end kangaroo derive fan spider flee brand bird seat accident rather margin loop fabric sweet peasant wear nuclear deliver alert reform train"
  }
}
```
The `.mnemonic.phrase` is what you need to recreate the wallet. Be sure to pass the `--id` flag with the name of the wallet as the value if you are not using the primary wallet.

You can now use that mnemonic to create a new wallet. The `--id` flag is used to name the wallet.

```bash
$ hsw-cli mkwallet --id newwallet --mnemonic "wire lion end kangaroo derive fan spider flee brand bird seat accident rather margin loop fabric sweet peasant wear nuclear deliver alert reform train"
```

You can now verify that the wallet was created with the correct mnemonic with the command:

```bash
$ hsw-cli --id newwallet master
```

[...]

PR #511, opened 2020-09-03
wallet: deep clean

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/467

~Requires https://github.com/handshake-org/hsd/pull/512 (CI tests will fail until rebase)~ ✅ 

This will likely be required for users with walletDB corruption due to the `CLAIM->REGISTER` bug (https://github.com/handshake-org/hsd/issues/454) and/or the `FINALIZE` bug (https://github.com/handshake-org/hsd/pull/464)

walletDB deepClean():
- deletes all mapping of blocks, outpoints and tx hashes to wallet ids
- deletes (almost) the entire txdb (all history, all balance data, for all wallets)
- keeps all BID blinds (otherwise a human must remember the hidden values)
- keeps all account metadata (account name, type, address depth, etc)
- keeps all name->wallet ID mapping (although this may be unnecessary)
- keeps all address-hash path mapping (addresses from all accounts)

wallet HTTP /deepclean:
- requires admin token, like /rescan
- requires setting parameter `I_HAVE_BACKED_UP_MY_WALLET` to `true`

A deep clean **does not** automatically initiate a rescan -- but one must be initiated by the user after deep clean is complete.

TODO:
- [x] add bmocha tests
- [x] test with shell script that corrupts wallet on unpatched version of hsd, then upgrades to master branch

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -200,6 +200,26 @@ class HTTP extends Server {
       await this.wdb.rescan(height);
     });
 
+    // Deep Clean
+    this.post('/deepclean', async (req, res) => {
+      if (!req.admin) {
+        res.json(403);
+        return;
+      }
+
+      const valid = Validator.fromRequest(req);
+      const disclaimer = valid.bool('I_HAVE_BACKED_UP_MY_WALLET', false);
+
+      enforce(
+        disclaimer,
+        'Deep Clean requires I_HAVE_BACKED_UP_MY_WALLET=true'
+      );
+
+      res.json(200, { success: true });
+
+      await this.wdb.deepClean();
+    });
+
     // Resend
     this.post('/resend', async (req, res) => {
       if (!req.admin) {
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -559,12 +559,31 @@ class WalletDB extends EventEmitter {
       lte: layout.T.max()
     });
 
-    this.logger.warning('Clearing all tx history for all wallets...');
-    // t[wid]* -> txdb
-    await removeRange({
-      gte: layout.t.min(),
-      lte: layout.t.max()
+    const wids = await this.db.keys({
+      gte: layout.W.min(),
+      lte: layout.W.max(),
+      parse: key => layout.W.decode(key)[0]
     });
+
+    for (const wid of wids) {
+      const wallet = await this.get(wid);
+      this.logger.warning(
+        'Clearing all tx history for wallet: %s (%d)',
+        wallet.id, wid
+      );
+
+      // remove all txdb data *except* blinds ('v')
+      const key = 'v'.charCodeAt();
+      const prefix = layout.t.encode(wid);
+      await removeRange({
+        gte: Buffer.concat([prefix, Buffer.alloc(1)]),
+        lt:  Buffer.concat([prefix, Buffer.from([key])])
+      });
+      await removeRange({
+        gt: Buffer.concat([prefix, Buffer.from([key + 1])]),
+        lte: Buffer.concat([prefix, Buffer.from([0xff])])
+      });
+    }
 
     await b.write();
 
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -504,6 +504,74 @@ class WalletDB extends EventEmitter {
   }
 
   /**
+   * Deep Clean:
+   * Keep all keys, account data, wallet maps (name and path).
+   * Dump all TX history and balance state.
+   * A rescan will be required but is not initiated automatically.
+   * @returns {Promise}
+   */
+
+  async deepClean() {
+    const unlock1 = await this.txLock.lock();
+    const unlock2 = await this.writeLock.lock();
+    const unlock3 = await this.readLock.lock();
+    try {
+      return await this._deepClean();
+    } finally {
+      unlock3();
+      unlock2();
+      unlock1();
+    }
+  }
+
+  /**
+   * Deep Clean (without locks):
+   * Keep all keys, account data, wallet maps (name and path).
+   * Dump all TX history and balance state.
+   * A rescan will be required but is not initiated automatically.
+   * @returns {Promise}
+   */
+
+  async _deepClean() {
+    this.logger.warning('Initiating Deep Clean...');
+
+    const b = this.db.batch();
+    const removeRange = (opt) => {
+      return this.db.iterator(opt).each(key => b.del(key));
+    };
+
+    this.logger.warning('Clearing block map, tx map and outpoint map...');
+    // b[height] -> block->wid map
+    await removeRange({
+      gte: layout.b.min(),
+      lte: layout.b.max()
+    });
+
+    // o[hash][index] -> outpoint->wid map
+    await removeRange({
+      gte: layout.o.min(),
+      lte: layout.o.max()
+    });
+
+    // T[hash] -> tx->wid map
+    await removeRange({
+      gte: layout.T.min(),
+      lte: layout.T.max()
+    });
+
+    this.logger.warning('Clearing all tx history for all wallets...');
+    // t[wid]* -> txdb
+    await removeRange({
+      gte: layout.t.min(),
+      lte: layout.t.max()
+    });
+
+    await b.write();
+
+    this.logger.warning('Deep Clean complete. A rescan is now required.');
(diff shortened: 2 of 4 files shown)


pinheadmz:
I wrote a bash script here: https://gist.github.com/pinheadmz/a25a58a4bea13a88035cf1899f7c3e15

- Checks out master branch, transfers a name and checks the balance (locked coins are still there)
- Stop hsd and checkout the finalize-test branch which includes #511 and #512 
- Execute DEEP RESCAN
- Check balance, locked coins are cleared out.

lib/wallet/http.js
```diff
@@ -200,6 +200,26 @@ class HTTP extends Server {
...
+
+      const valid = Validator.fromRequest(req);
+      const disclaimer = valid.bool('I_HAVE_BACKED_UP_MY_WALLET', false);
+
+      enforce(
+        disclaimer,
+        'Deep rescan requires I_HAVE_BACKED_UP_MY_WALLET=true'
+      );
+
+      res.json(200, { success: true });
```
RevCBH, line 218: Why is the response being set before the `deepRescan()` call? Is it to avoid caller timeout since it's a long-running operation?

lib/wallet/walletdb.js
```diff
@@ -503,6 +503,105 @@ class WalletDB extends EventEmitter {
...
+
+  /**
+   * Deep rescan (without locks)
+   * Keep all keys, account data, wallet name and path maps.
+   * Dump all TX history and balance state, rescan from genesis.
+   * @param {Boolean?} rescan -- only used for testing
+   * @returns {Promise}
+   */
+
+  async _deepRescan(rescan) {
```
RevCBH, line 535: Could break `_deepRescan` into two parts, drop and scan, then call those directly from the test, instead. 'Test only' params are just asking to be passed wrong.

test/wallet-deeprescan-test.js
```diff
@@ -0,0 +1,259 @@
...
+    aliceHistory = await alice.getHistory();
+    aliceAcct0Info = await alice.getAccount(0);
+
+    bobBalance = await bob.getBalance();
+    bobNames = await bob.getNames();
+    bobHistory = await bob.getHistory();
+    bobAcct0Info = await bob.getAccount(0);
+  });
+
+  it('should DEEP RESCAN - wipe only', async () => {
```
RevCBH, line 156: It might be worth calling out that this is where the test **really** starts, since everything before it seems like setup, right?

RevCBH (commented):
This makes sense to me and looks generally reasonable. Not sure what the approve process is, so just leaving the review as 'Comment', but there's nothing blocking as far as I can tell.

lib/wallet/http.js
pinheadmz: Exactly. You see the same logic in the regular `/rescan` endpoint.

lib/wallet/walletdb.js
pinheadmz: Yeah this is a good point, I'll think about how it could be done and if that's worth it.
You caught me in that I'm just using the param for testing but you could also think of it like the API just doesn't expose that option. I don't want the user to have to execute two different commands to fix their wallet state, and after the data is deleted a rescan is really required.

test/wallet-deeprescan-test.js
pinheadmz: That's true but this is the pattern in all the other tests. You'll notice a lot of the setup steps are repeated throughout the suite like `should fund wallets`, etc. I think using ALL CAPS for DEEP RESCAN is pretty loud but I'm open to suggestions ...?

lib/wallet/http.js
RevCBH: :+1:

lib/wallet/walletdb.js
RevCBH: Probably not worth the work, but the way I would do it is:

- Split `_deepRescan` by moving 590-602 to a new method. I'd call it something like `_rescanFromGenesis` or something.
- Remove the parameter from `_deepRescan` and rename it to '_dropHistory` or something
- Keep `deepRescan` as the top-level entry point, but remove the parameter and have it call `_dropHistory` then `_rescanFromGenesis`
- Tests just call the `_dropHistory` and `_rescanfromGenesis` at the appropriate time.

Might end up just making it harder to read.

test/wallet-deeprescan-test.js
RevCBH: Ahh, yes, didn't realize that was the pattern elsewhere, carry on.

lib/wallet/walletdb.js
pinheadmz: Hm OK yeah I think this may be complicating things, especially if the goal is just to avoid a test-only parameter. But I think its a valid suggestion. I'll leave this open for other reviewers, but it could also be future work. What I'd like to see is the finalize stuff merged along with this so we can release https://github.com/handshake-org/hsd/milestone/2 and users can rebuild their wallets as soon as possible. The longer we wait, the more users will have corrupted wallet state.

pinheadmz: I've been thinking about this and now I agree, this command could be reduced to `deep clean` and users should follow it up with a regular rescan. The only catch is the rescan MUST MUST MUST include the entire lifetime of the wallet. A normal rescan doesn't require this because it rollsback to the given height and rescans from there. `deep clean` wipes everything. 

Hopefully I get some time in the near future to clean this up and we can get another round of review to merge.

RevCBH: Cool. Happy to take a crack at it, too, if it'd be helpful.

turbomaze:
Just to add my 2c, have tested deep rescans somewhat extensively on local regtest wallets, on small mainnet wallets, and also some larger mainnet wallets. Have not encountered any issues!

pinheadmz:
@RevCBH @turbomaze This PR has been updated! (finally!)

It's not `deep rescan` anymore - now its just `deep clean` and it does NOT automatically start a rescan. Meaning this operation is now a two-step process. Changelog has been updated to reflect this as well as the integration test and also the bash script I describe [here](https://github.com/handshake-org/hsd/pull/511#issuecomment-689175694)

pinheadmz:
This is a sample of the debug log output when Deep Clean is run:

```
[warning] (wallet) Initiating Deep Clean...
[warning] (wallet) Clearing block map, tx map and outpoint map...
[warning] (wallet) Clearing all tx history for wallet: primary (0)
[warning] (wallet) Clearing all tx history for wallet: 5V1xGufkA88e1pbPGZyFiy58oKg1pVk (1)
[warning] (wallet) Clearing all tx history for wallet: 5V1ybvoPa4vWJA17ftSKrvbpiVNm3AX (2)
[warning] (wallet) Deep Clean complete. A rescan is now required.
```

TLD-O:
Just tried it on my mainnet wallet, with a lot of transfer and finalize in it and negative balance. Seems to work well.
Before : 
```
"balance": {
        "account": -1,
        "tx": 1768,
        "coin": 185,
        "unconfirmed": 9116650351,
        "confirmed": 9116650351,
        "lockedUnconfirmed": 9513310000,
        "lockedConfirmed": 6106000011
    }
```

after
```
"balance": {
        "account": -1,
        "tx": 1768,
        "coin": 185,
        "unconfirmed": 9116650351,
        "confirmed": 9116650351,
        "lockedUnconfirmed": 6106000011,
        "lockedConfirmed": 6106000011
    }
```

lib/wallet/walletdb.js
```diff
@@ -503,6 +503,93 @@ class WalletDB extends EventEmitter {
...
+      this.logger.warning(
+        'Clearing all tx history for wallet: %s (%d)',
+        wallet.id, wid
+      );
+
+      // remove all txdb data *except* blinds ('v')
+      const key = 'v'.charCodeAt();
+      const prefix = layout.t.encode(wid);
+      await removeRange({
+        gte: Buffer.concat([prefix, Buffer.alloc(1)]),
```
tynes, line 579: Is this supposed to be a `1` or a `0`?

pinheadmz: It "should" be a single `0x00` byte buffer. This code will remove everything in the txdb _up to_ the blinds. The next snippet removes everything in the txdb _after_ the blinds:

Removed: `0x00 -> v` and `v -> 0xff`

PR #512, opened 2020-09-08
txdb: fix lockedUnconfirmed if FINALIZE is inserted with block

pinheadmz:
This is an edge case we should have addressed in #464 especially after dealing with basically the same behavior in #387. And actually if I'd written better test coverage in #387 to actually include TRANSFER and FINALIZE, we probably would've fixed all the bugs six months ago 🤷‍♂️ 😬 

The issue is that in bcoin and hsd, `unconfirmed` includes all confirmed amounts in addition to unconfirmed amounts, they are not mutually exclusive. Normally when we send a TX, it is inserted into the database as unconfirmed. Then when we see it in a block, we update the confirmed balance but we leave the unconfirmed balance alone (it was already updated when the TX was first seen in the mempool).

During a rescan (or more explicitly, an import-from-seed or [deep rescan](https://github.com/handshake-org/hsd/pull/511)) we never see the TX as unconfirmed -- it is inserted into the database with the block that confirms it. So to keep the `unconfirmed` balance in sync, we need to update that value first, and then continue with the confirmation process that updates the confirmed balance.

### DIFF
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -1933,6 +1933,8 @@ describe('Wallet', function() {
     let start;
     // The main test wallet, and wallet that will receive the FINALIZE.
     let wallet, recip;
+    // Store balance data before rescan to ensure rescan was complete
+    let recipBalBefore, senderBalBefore;
 
     // Hack required to focus test on txdb mechanics.
     // We don't otherwise need WalletDB or Blockchain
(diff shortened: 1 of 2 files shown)


test/wallet-test.js
```diff
@@ -2153,21 +2155,21 @@ describe('Wallet', function() {
...
-      assert.strictEqual(senderBal.clocked, secondHighest);
-
-      let recipBal = await recip.getBalance();
-      assert.strictEqual(recipBal.tx, 0);
-      assert.strictEqual(recipBal.coin, 0);
-      assert.strictEqual(recipBal.confirmed, 0);
-      assert.strictEqual(recipBal.unconfirmed, 0);
-      assert.strictEqual(recipBal.ulocked, 0);
-      assert.strictEqual(recipBal.clocked, 0);
+      const senderBal1 = await wallet.getBalance();
```
tynes, line 2158: Good idea to make these very explicitly different variables

test/wallet-test.js
```diff
@@ -2420,6 +2442,135 @@ describe('Wallet', function() {
...
+      };
+      await wallet.txdb.add(register.toTX(), block);
+
+      // Check
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 5);
+      // Wallet coin count doesn't change:
+      // REVEAL + fee money -> REGISTER + change
+      assert.strictEqual(bal.coin, 4);
+      assert.strictEqual(bal.confirmed, 10e6 - (4 * fee));
```
tynes, line 2501: Maybe it would be more scalable to have a global "transaction count" variable and a helper function `feeDiff` or `spentFees` or something like that, so that each section doesn't need to do `(n * fee)` and instead you could just call a function that increments the global variable while returning the required number. What do you think?

test/wallet-test.js
pinheadmz: 👍

test/wallet-test.js
pinheadmz: Ok, that's what we did in #464 but not in #387. I think you're right though its a good precedent to set and looks cleaner, will update.

pinheadmz:
rebase to 0e7865ca1ba33aaeeb00bfcfcf2ec869697f0f45:
clean up test re: https://github.com/handshake-org/hsd/pull/512#discussion_r485247084

PR #514, opened 2020-09-16
Mining: set minimum block weight to 5000, parse as configuration option

pinheadmz:
Addresses https://github.com/handshake-org/hsd/issues/430 -- [solution 1 & 2](https://github.com/handshake-org/hsd/issues/430#issuecomment-614611881)

(see also #515)

# Goal

Prevent low- or no-fee transactions from being stuck in the mempool forever, by adjusting mining policy to occasionally include them.

This PR is mainly changing a constant in policy.js from a value that used to be `0`:
```
/**
 * Minimum block size to create. Block will be
 * filled with free transactions until block
 * reaches this weight.
 * @const {Number}
 * @default
 */

exports.MIN_BLOCK_WEIGHT = 5000;
```

What this means is, in miner.js `assemble()` when a block is being put together to be hashed, if the current block template has a total weight under 5000, the miner will insert "free" transactions from the mempool (very low fee transactions).

One such TX might be this poor guy:

https://shakescan.com/transaction/a239970210b948d462adf8a07fb2e8bf9b301ba34942cb01b86eb96aeec2ad30

...that has been patiently pending for SIX MONTHS.

5000 weight units is tiny by the way. An "empty" block with just a coinbase is already `1308` weight.

I ran a few heuristics to determine how often a block with <5000 weight ever gets mined. Parsing blocks 2016-33054. ([all data](https://gist.github.com/pinheadmz/e7257136a60325f705458a8cb71185bc))
```
weight || count
   1000   2834
   2000   1391
   3000   1175
   4000   840
   5000   829
   6000   727
   7000   708
   8000   719
   9000   598
   10000  4361
   20000  2867
   30000  2128
   40000  1595
   50000  1206
   60000  979
   70000  848
   80000  724
   90000  628
   100000 502
```

and another test to see how big the average "gap" in blocks is between blocks of a certain weight:

```
wt.   || average gap
1000   10.942484121383204
2000   22.249640287769783
3000   26.313458262350938
4000   36.84642857142857
5000   37.39492753623188
6000   42.57575757575758
7000   43.74115983026874
8000   43.17941585535466
9000   51.88609715242881
10000   7.109736540664375
20000  10.815202231520223
30000  14.49906015037594
40000  19.337092731829575
50000  25.409958506224065
60000  31.163265306122447
70000  36.560802833530104
80000  42.881051175656985
90000  48.346092503987244
100000 60.439121756487026
```

# So,

[...]

Issue #430, opened 2020-04-16
Better handling of low-fee TXs


pinheadmz:
Some low-fee transactions are propagating around the network somehow even though they are below the `minRelay` rate for the network.

One such TX has been "stuck" in the mempool for over a month:

https://shakescan.com/transaction/a239970210b948d462adf8a07fb2e8bf9b301ba34942cb01b86eb96aeec2ad30

```
Fee   0.000062 HNS
Time  2020-03-05 01:49
```
(Posting this issue on `2020-04-16`)

Another such TX was sent to me by a user and I wrote a script to check its standardness and fee rate: https://gist.github.com/pinheadmz/406c0d28751c969e46d4ed57823a26df

```
check undefined
stdinputs true
min fee 270
tx  fee 2
is free false
```

With default settings, nodes should allow a small rate of "free" TXs into their mempool:

https://github.com/handshake-org/hsd/blob/2d5ab277374ef751565caeaac84166c713ef9f4f/lib/mempool/mempool.js#L1576-L1595

**However, these low-fee TXs are either not getting relayed to miners, or simply not included in blocks.** Since there is no RBF in hsd, they are stuck forever (or until enough network nodes reset with clear mempools). This could also be an issue with the mining code not including low-fee TXs even if there is room in the block.

pinheadmz:
## Solution 1

We should consider bumping up the minimum block weight value:

https://github.com/handshake-org/hsd/blob/2d5ab277374ef751565caeaac84166c713ef9f4f/lib/protocol/policy.js#L133-L141

The policy of `0` minimum weight comes from bcoin in November 2016: https://github.com/bcoin-org/bcoin/commit/bc00697adbe6ae3c4ea5960b1cd2c00279fb23a7
So that would be just after SegWit adoption and a very busy time for Bitcoin, definitely no need to allow low-fee transactions.

## Solution 2

Expose `minWeight` as a Full Node option and socially convince miners to set it to some value like at least `1000` maybe even `10000`

---

I ran test locally using `rpc settxfee 0.000001` and by adjusting these values was able to confirm low-fee TXs.

HarukaMa:
I think with solution 1 and 2, if the chain is busy enough and the parameter is low like 1000, there will still be issues that free txs are not being mined for quite a while until there are some lucky block streaks and without RBF there will still be no way to boost it.

randomlogin:
[...]

PR #515, opened 2020-09-16
Mempool: evict expired transactions regardless of current mempool size

pinheadmz:
Port of https://github.com/bcoin-org/bcoin/pull/997 with explanation [here](https://github.com/bcoin-org/bcoin/issues/996)

(see also #514 )

Addresses https://github.com/handshake-org/hsd/issues/430 -- [solution 4](https://github.com/handshake-org/hsd/issues/430#issuecomment-614611881)

Summary: We already have logic to remove ["old"](https://github.com/handshake-org/hsd/blob/7826128ff72c20fdf42c659309dcffea7a28f6cc/lib/protocol/policy.js#L116-L123) transactions from the mempool, but that logic is never executed unless the mempool is full. This PR checks for and evicts old transactions every time a new transaction is added to the mempool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ the miner will add transactions up to the minimum weight that would normally be
 ignored for being "free" (paying a fee below policy limit). The default value is
 raised from `0` to `5000` (a 1-in, 2-out BID transaction has a weight of about `889`).
 
+- Transactions that have sat unconfirmed in the mempool for 3 days will be evicted.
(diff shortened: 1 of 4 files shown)


chjj:
Is there perf overhead here? Inserting everything into a heap is going to be costly.

pinheadmz:
> Is there perf overhead here? Inserting everything into a heap is going to be costly.

@chjj thanks for taking a look, I refactored the patch so we just screen and evict expired mempool entries before getting to the Heap, but still before checking overall mempool size for excess and returning `false`.

PR #519, opened 2020-10-14
wallet: fix missing tx when rescan with filter update

chikeichan:
**CHANGE:**
- add migration-id `1` to change lookahead value from `uint8` to `uint32`
- when either `changeDepth` or `receiveDepth` exceed current lookahead value during rescan, increment lookahead value by 20, and then send updated filter and block height to to `node/http.js`
- when filter is updated during rescan, `node/http` will halt the current rescan and start a new one with the new filter from the new height

**Summary**

This borrow heavily from [this issue in bcoin](https://github.com/bcoin-org/bcoin/commit/c6859b347bd6989c1eb87fc58e05b6fcc4b91c8b) but with a simplified approach in `node/http`. 

This version of `hsd` and rescan functionality can be tested in this [branch](https://github.com/kyokan/bob-wallet/tree/bump-hsd) in Bob. I have tested this with a wallet with 5k+ transactions and lookahead value at 3k+ and have no issue.

### DIFF
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -80,6 +80,7 @@ class WalletDB extends EventEmitter {
 
     // Address and outpoint filter.
     this.filter = new BloomFilter();
+    this.filterUpdated = false;
 
     this.init();
   }
@@ -195,6 +196,8 @@ class WalletDB extends EventEmitter {
       this.state.height,
       this.state.startHeight);
 
+    await this.migrateChange();
+
     const wallet = await this.ensure({
       id: 'primary'
     });
@@ -206,8 +209,6 @@ class WalletDB extends EventEmitter {
       wallet.id, wallet.wid, addr.toString(this.network));
 
     this.primary = wallet;
-
-    await this.migrateChange();
   }
 
   /**
@@ -597,12 +598,15 @@ class WalletDB extends EventEmitter {
    * @returns {Promise}
    */
 
-  syncFilter() {
+  async syncFilter() {
+    const height = this.state.height;
     this.logger.info('Sending filter to server (%dmb).',
       this.filter.size / 8 / (1 << 20));
 
     this.filterSent = true;
-    return this.client.setFilter(this.filter);
+    this.filterUpdated = false;
+
+    return this.client.setFilter(this.filter, height);
   }
 
   /**
@@ -2152,8 +2156,15 @@ class WalletDB extends EventEmitter {
           total += 1;
       }
 
-      // Sync the state to the new tip.
-      await this.setTip(tip);
+      // Make sure to update the filter with the node
+      // server and the p2p network.
+      if (this.filterUpdated) {
+        await this.syncFilter();
+        await this.rollback(this.state.height - 1);
+      } else {
+        // Sync the state to the new tip.
+        await this.setTip(tip);
+      }
     } finally {
       this.confirming = false;
     }
@@ -2309,11 +2320,17 @@ class WalletDB extends EventEmitter {
 
       assert(wallet);
 
-      if (await wallet.add(tx, block)) {
+      const {details, derived} = await wallet.add(tx, block);
+
+      if (details) {
         this.logger.info(
           'Added transaction to wallet in WalletDB: %s (%d).',
           wallet.id, wid);
         result = true;
+      }
+
+      if (derived.length > 0) {
+        this.filterUpdated = true;
       }
     }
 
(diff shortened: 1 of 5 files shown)


lib/wallet/walletdb.js
```diff
@@ -206,8 +210,62 @@ class WalletDB extends EventEmitter {
...
+          account.wid = wid;
+          account.accountIndex = accountIndex;
+          account.name = name;
+
+          await this.saveAccount(b, account);
+          this.logger.info(
+            'migrated lookahead (wid=%d, name=%s)',
+            wid, name);
+        } catch (e) {
+          //
```
tynes, line 262: Skipping this on purpose?

tynes (changes requested):
It would be nice to have some tests and also squash the commits so that they follow the conventions of the codebase

chikeichan:
> Are you sure we need all this work on the lookahead value?
> I was hoping that if we merge #506 (bump lookahead to 200, still within the U8 value) and the filter update commit from bcoin we'd basically be done with this. There is a test as well in this other commit: [bcoin-org/bcoin@46967f8](https://github.com/bcoin-org/bcoin/commit/46967f8da604412051d0eff223c45ff6ecc15e3a)


I see what you mean now - after some digging i believe there is a bug in the [syncDepth](https://github.com/handshake-org/hsd/blob/master/lib/wallet/wallet.js#L3819) method where some new account keys are not being saved. To workaround, I basically call [setLookahead] with account depth to force account keys to be generated. 

Let me try to fix the bug in syncDepth instead -- that should allow us to leave lookahead value unchanged.

chikeichan:
> Are you sure we need all this work on the lookahead value?

So after more digging, I can confirm that there is at least one user wallet with a lookahead value of 500, testing using the rescan filter fix in this PR + hardcoding lookahead value.

@pinheadmz This [issue](https://github.com/handshake-org/hsd/issues/407#issuecomment-668625559) from you helped lot! I will open a separate PR for lookahead value migration and see if there are any community support for it. For Bob, I will just maintain a simple fork with a hardcode lookahead value of `1000` for now so that lookahead datatype stays the same.

@tynes squashed and added unit test!

test/node-sync-test.js
```diff
@@ -0,0 +1,164 @@
...
+      workersSize: 2,
+      bip37: true,
+      plugins: [require('../lib/wallet/plugin')],
+      listen: true,
+      port: ports.p2p,
+      httpPort: ports.node,
+      rsPort: ports.rs,
+      nsPort: ports.ns,
+      env: {
+        'BCOIN_WALLET_HTTP_PORT': ports.wallet.toString()
```
pinheadmz, line 55: This might need to be changed to `HSD_WALLET_HTTP_PORT` :-)

test/node-sync-test.js
```diff
@@ -0,0 +1,164 @@
...
+
+    // Prepare the miner and wallet.
+    const {miner, chain} = node;
+    wdb = node.require('walletdb').wdb;
+    wallet = await wdb.create();
+    miner.addAddress(await wallet.receiveAddress());
+
+    // Mature the initial coins to use for the
+    // use in generating the test case.
+    for (let i = 0; i < 200; i++) {
```
pinheadmz, line 99: I think in hsd the coinbase maturity for regtest is much smaller so we can reduce this here.

https://github.com/handshake-org/hsd/blob/c85d9b4c743a9e1c9577d840e1bd20dee33473d3/lib/protocol/networks.js#L745

pinheadmz:
note, Braydon doesn't like to be tagged in Handshake-org stuff but this is mostly his work, I'm not sure if we need to be explicit about the license or anything but you might consider just adding his name to the commit message.

pinheadmz:
@chikeichan finally getting to more deeply review this. I'm playing with the test locally, converting it more hsd-friendly parameters etc. Looks like the main patch isn't just Braydon's commit but some extra work as well, including calling `rollback()` -- were you able to solve the issue here? Can you provide a test script with bash or anything to evaluate the changes? This issue is going to be my next top priority, I may start a new branch with my modified test and Braydon's bcoin commit and then add back in some of your changes.

chikeichan:
hello @pinheadmz - nice to see you active again!

For this PR, the general changes to scan is:
- wallet client will call `setFilter` with current height
- wallet node receive `set filter` in the socket event
- wallet node stop current scan progress (kinda hacky now by throwing an error)
- wallet db rollback to current height minus 1
- wallet node restart new scan from new filter height minus 1

I do not have test for this changes unfortunately; I have been using my personal wallet with a lookahead value of around 500 to test during development.

I am more than happy to review new branch + merge back to Bob when it's ready!

Detetivepro1 (approved):

Issue #93, opened 2019-01-25
Argument 'force' is unused in `wallet.makeOpen()`, others...


pinheadmz:
https://github.com/handshake-org/hsd/blob/master/lib/wallet/wallet.js#L1493

I see this param `force` get passed around in wallet, and checked `typeof force === 'boolean'` but never really used? Am I missing something or is this for an old or future feature?

pinheadmz:
Following up on this a bit... the `force` param is consumed in some downstream functions to bypass the database locks:

https://github.com/handshake-org/hsd/blob/505b6f76834ef93ee883b8052780936c92f199fa/lib/wallet/wallet.js#L1134-L1141

But we still aren't actually using the argument to `makeOpen` this way.

PR #529, opened 2021-01-05
Proposal : add the possibility to create auction-related transactions in advance

HDardenne-HNS:
At the moment, if I bid on a name now, I can't create the reveal and redeem transactions before the the name gets respectively in the reveal period and close period. That is, unless I use a web wallet (Namebase for example), I have get back X blocks later to create the reveal (and send it) and get back again Y blocks later for the redeem (and send it). I think we can keep the REGISTER out of the scope. 

The goal of this PR is to add the possibility to somewhat automate the auction process. The user will still have to either keep a personal hsd open to send the transactions at the right time or trust someone else to do it (sometimes referred as a kind of watchtower).

If it is accepted, a user could create a bid (and broadcast it immediately or not), and prepare the corresponding reveal and redeem transactions in advance. Then he could either send them at the right time or delegate this action to a third party. It could make the bidding experience on personal wallet less tricky (the principal concern I heard being the reveal that is not made automatically by the blockchain and people being afraid to forget to do it)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # HSD Release Notes & Changelog
 
+## unreleased
+
+- Adds new wallet HTTP endpoint `/wallet/:id/auction` based on `POST /wallet/:id/bid`.
+It requires an additional parameter `broadcastBid` set to either true or false.
+This action returns a bid and its corresponding reveal, the reveal being prepared in advance.
+The bid will be broadcasted either during the creation (`broadcastBid=true`) or at a later time
+(`broadcastBid=false`).
+The reveal will have to be broadcasted at a later time, during the REVEAL phase.
+The lockup must include a blind big enough to ensure the BID will be the only input of the REVEAL
+transaction. 
+
 ## v2.3.0
 
 ### Node changes
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -1084,6 +1084,45 @@ class HTTP extends Server {
       return res.json(200, mtx.getJSON(this.network));
     });
 
+    // Create auction-related transactions in advance (bid and reveal for now)
+    this.post('/wallet/:id/auction', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+      const bid = valid.u64('bid');
+      const lockup = valid.u64('lockup');
+      const passphrase = valid.str('passphrase');
+      const sign = valid.bool('sign', true);
+      const broadcastBid = valid.bool('broadcastBid');
+
+      assert(name, 'Name is required.');
+      assert(bid != null, 'Bid is required.');
+      assert(lockup != null, 'Lockup is required.');
+      assert(broadcastBid != null, 'broadcastBid is required.');
+      assert(broadcastBid ? sign : true, 'Must sign when broadcasting.');
+
+      const options = TransactionOptions.fromValidator(valid);
+      const auctionTxs = await req.wallet.createAuctionTxs(
+        name,
+        bid,
+        lockup,
+        options
+      );
+
+      if (sign) {
+        if (broadcastBid) {
+          auctionTxs.bid = await req.wallet.sendMTX(auctionTxs.bid, passphrase);
+        } else {
+          await req.wallet.sign(auctionTxs.bid, passphrase);
+        }
+        await req.wallet.sign(auctionTxs.reveal, passphrase);
+      }
+
+      return res.json(200, {
+        bid: auctionTxs.bid.getJSON(this.network),
+        reveal: auctionTxs.reveal.getJSON(this.network)
+      });
+    });
+
     // Create Reveal
     this.post('/wallet/:id/reveal', async (req, res) => {
       const valid = Validator.fromRequest(req);
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -37,6 +37,8 @@ const {states} = require('../covenants/namestate');
 const {types} = rules;
 const {Mnemonic} = HD;
 const {BufferSet} = require('buffer-map');
+const Coin = require('../primitives/coin');
+const Outpoint = require('../primitives/outpoint');
 
 /*
  * Constants
@@ -1195,12 +1197,15 @@ class Wallet extends EventEmitter {
     if (rate == null)
       rate = await this.wdb.estimateFee(options.blocks);
 
-    let coins;
+    let coins = options.coins || [];
+    assert(Array.isArray(coins));
     if (options.smart) {
-      coins = await this.getSmartCoins(options.account);
+      const smartCoins = await this.getSmartCoins(options.account);
+      coins = coins.concat(smartCoins);
     } else {
-      coins = await this.getCoins(options.account);
-      coins = this.txdb.filterLocked(coins);
+      let availableCoins = await this.getCoins(options.account);
+      availableCoins = this.txdb.filterLocked(availableCoins);
+      coins = coins.concat(availableCoins);
     }
 
     await mtx.fund(coins, {
@@ -1818,6 +1823,76 @@ class Wallet extends EventEmitter {
     } finally {
       unlock();
     }
+  }
+
+  /**
+   * Create and finalize a bid & a reveal (in advance)
+   * MTX with a lock.
+   * @param {String} name
+   * @param {Number} value
+   * @param {Number} lockup
+   * @param {Object} options
+   * @returns {{ bid: MTX; reveal: MTX;}} {bid, reveal}
+   */
+
+  async createAuctionTxs(name, value, lockup, options) {
+    const unlock = await this.fundLock.lock();
+    try {
+      return await this._createAuctionTxs(name, value, lockup, options);
+    } finally {
+      unlock();
+    }
+  }
+
+  /**
+   * Create and finalize a bid & a reveal (in advance)
+   * MTX without a lock.
+   * @param {String} name
+   * @param {Number} value
+   * @param {Number} lockup
+   * @param {Object} options
+   * @returns {{ bid: MTX; reveal: MTX;}} {bid, reveal}
+   */
+
+  async _createAuctionTxs(name, value, lockup, options) {
+    const bid = await this._createBid(name, value, lockup, options);
+
+    const bidOuputIndex = bid.outputs.findIndex(o => o.covenant.isBid());
+    const bidOutput = bid.outputs[bidOuputIndex];
+    const bidCoin = Coin.fromTX(bid, bidOuputIndex, -1);
+
+    // Prepare the data needed to make the reveal in advance
+    const nameHash = bidOutput.covenant.getHash(0);
+    const height = bidOutput.covenant.getU32(1);
+
+    const coins = [];
+    coins.push(bidCoin);
+
+    const blind = bidOutput.covenant.getHash(3);
+    const bv = await this.getBlind(blind);
+    if (!bv)
+      throw new Error('Blind value not found.');
+    const { nonce } = bv;
+
+    const reveal = new MTX();
+    const output = new Output();
+    output.address = bidCoin.address;
+    output.value = value;
+    output.covenant.type = types.REVEAL;
+    output.covenant.pushHash(nameHash);
+    output.covenant.pushU32(height);
+    output.covenant.pushHash(nonce);
+    reveal.addOutpoint(Outpoint.fromTX(bid, bidOuputIndex));
+    reveal.outputs.push(output);
+
+    await this.fill(reveal, { ...options, coins: coins });
+    assert(
+      reveal.inputs.length === 1,
+      'Pre-signed REVEAL must not require additional inputs'
+    );
+
+    const finalReveal = await this.finalize(reveal, options);
+    return { bid, reveal: finalReveal };
   }
 
   /**
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -846,6 +846,71 @@ describe('Wallet HTTP', function() {
     matchTxId(auction.reveals, state.reveals[1].hash);
   });
 
+  it('should create a bid and a reveal (reveal in advance)', async () => {
+    const balanceBeforeTest = await wallet.getBalance();
+    const lockConfirmedBeforeTest = balanceBeforeTest.lockedConfirmed;
+    const lockUnconfirmedBeforeTest = balanceBeforeTest.lockedUnconfirmed;
+
+    await wallet.createOpen({ name: name });
+
+    await mineBlocks(treeInterval + 2, cbAddress);
+
+    const balanceBeforeBid = await wallet.getBalance();
+    assert.equal(balanceBeforeBid.lockedConfirmed - lockConfirmedBeforeTest, 0);
+    assert.equal(
+      balanceBeforeBid.lockedUnconfirmed - lockUnconfirmedBeforeTest,
+      0
+    );
+
+    const bidValue = 1000000;
+    const lockupValue = 5000000;
+
+    const auctionTxs = await wallet.client.post(
+      `/wallet/${wallet.id}/auction`,
+      {
+        name: name,
+        bid: 1000000,
+        lockup: 5000000,
+        broadcastBid: true
+      }
+    );
+
+    await mineBlocks(biddingPeriod + 1, cbAddress);
+
+    let walletAuction = await wallet.getAuctionByName(name);
+    const bidFromWallet = walletAuction.bids.find(
+      b => b.prevout.hash === auctionTxs.bid.hash
+    );
+    assert(bidFromWallet);
+
+    const { info } = await nclient.execute('getnameinfo', [name]);
+    assert.equal(info.name, name);
+    assert.equal(info.state, 'REVEAL');
+
+    const b5 = await wallet.getBalance();
+    assert.equal(b5.lockedConfirmed - lockConfirmedBeforeTest, lockupValue);
+    assert.equal(b5.lockedUnconfirmed - lockUnconfirmedBeforeTest, lockupValue);
+
+    await nclient.broadcast(auctionTxs.reveal.hex);
+    await mineBlocks(1, cbAddress);
+
+    walletAuction = await wallet.getAuctionByName(name);
+    const revealFromWallet = walletAuction.reveals.find(
+      b => b.prevout.hash === auctionTxs.reveal.hash
+    );
+    assert(revealFromWallet);
+
+    const b6 = await wallet.getBalance();
+    assert.equal(b6.lockedConfirmed - lockConfirmedBeforeTest, bidValue);
+    assert.equal(b6.lockedUnconfirmed - lockUnconfirmedBeforeTest, bidValue);
+
+    await mineBlocks(revealPeriod + 1, cbAddress);
+
+    const ns = await nclient.execute('getnameinfo', [name]);
+    const coin = await wallet.getCoin(ns.info.owner.hash, ns.info.owner.index);
+    assert.ok(coin);
+  });
+
   it('should create a redeem', async () => {
     await wallet.createOpen({
       name: name
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -2869,4 +2869,344 @@ describe('Wallet', function() {
       assert.equal(claim.covenant.isClaim(), true);
     });
   });
+
+  describe('Create auction-related TX in advance', function () {
+    const network = Network.get('regtest');
+    const workers = new WorkerPool({ enabled });
+    const wdb = new WalletDB({ network, workers });
+    // This test executes a complete auction for this name
+    const name = 'satoshi-in-advance';
+    // There will be two bids. Both from our wallet, one made in advance and
+    // another created and broadcasted right away
+    const value = 1e6;
+    const lockup = 2e6;
+    const secondHighest = value - 1;
+    // All TXs will have a hard-coded fee to simplify the expected balances,
+    // along with counters for OUTGOING (un)confirmed transactions.
+    const fee = 10000;
+    let uTXCount = 0;
+    let cTXCount = 0;
+    // Initial wallet funds
+    const fund = 10e6;
+    // Store height of auction OPEN to be used in second bid.
+    // The main test wallet, and wallet that will receive the FINALIZE.
+    let wallet;
+    let unsentReveal;
+
+    // Hack required to focus test on txdb mechanics.
+    // We don't otherwise need WalletDB or Blockchain
+    wdb.getRenewalBlock = () => {
+      return network.genesis.hash;
+    };
+
+    before(async () => {
+      await wdb.open();
+      wallet = await wdb.create();
+      // rollout all names
+      wdb.height = 52 * 144 * 7;
+    });
+
+    after(async () => {
+      await wdb.close();
+    });
+
+    it('should fund wallet', async () => {
+      const addr = await wallet.receiveAddress();
+
+      // Fund wallet
+      const mtx = new MTX();
+      mtx.addOutpoint(new Outpoint(Buffer.alloc(32), 0));
+      mtx.addOutput(addr, fund);
+      const tx = mtx.toTX();
+
+      // Dummy block
+      const block = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+
+      // Add confirmed funding TX to wallet
+      await wallet.txdb.add(tx, block);
+
+      // Check
+      const bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 1);
+      assert.strictEqual(bal.coin, 1);
+      assert.strictEqual(bal.confirmed, fund);
+      assert.strictEqual(bal.unconfirmed, fund);
+      assert.strictEqual(bal.ulocked, 0);
+      assert.strictEqual(bal.clocked, 0);
+    });
+
+    it('should send and confirm OPEN', async () => {
+      const open = await wallet.sendOpen(name, false, { hardFee: fee });
+      uTXCount++;
+
+      // Check
+      let bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 2);
+      assert.strictEqual(bal.coin, 2);
+      assert.strictEqual(bal.confirmed, fund);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, 0);
+      assert.strictEqual(bal.clocked, 0);
+
+      // Confirm OPEN
+      const block = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+      await wallet.txdb.add(open, block);
+      cTXCount++;
+
+      // Check
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 2);
+      assert.strictEqual(bal.coin, 2);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, 0);
+      assert.strictEqual(bal.clocked, 0);
+    });
+
+    it('should send and confirm BID', async () => {
+      // Advance to bidding
+      wdb.height += network.names.treeInterval + 1;
+      const losingBid = await wallet.sendBid(name, secondHighest, lockup, {
+        hardFee: fee
+      });
+      uTXCount++;
+
+      // Check
+      let bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 3);
+      assert.strictEqual(bal.coin, 3);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, lockup);
+      assert.strictEqual(bal.clocked, 0);
+
+      // Confirm BID
+      let block = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+      await wallet.txdb.add(losingBid, block);
+      cTXCount++;
+
+      const losingBlindFromMtx = losingBid.outputs
+        .find(o => o.covenant.isBid())
+        .covenant.getHash(3);
+      let allBids = await wallet.getBids();
+      assert.strictEqual(allBids.length, 1);
+      const losingBlindFromWallet = allBids.find(b =>
+        b.blind.equals(losingBlindFromMtx)
+      );
+      assert(losingBlindFromWallet);
+
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 3);
+      assert.strictEqual(bal.coin, 3);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, lockup);
+      assert.strictEqual(bal.clocked, lockup);
+
+      const auctionsTxs = await wallet.createAuctionTxs(name, value, lockup, {
+        hardFee: fee
+      });
+      const winningBidUnsent = auctionsTxs.bid;
+      unsentReveal = auctionsTxs.reveal;
+
+      const winningBid = await wallet.sendMTX(winningBidUnsent, null);
+      uTXCount++;
+
+      // Check
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 4);
+      assert.strictEqual(bal.coin, 4);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, 2 * lockup);
+      assert.strictEqual(bal.clocked, lockup);
+
+      // Confirm BID
+      block = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+      await wallet.txdb.add(winningBid, block);
+      cTXCount++;
+
+      const winningBlindFromMtx = winningBid.outputs
+        .find(o => o.covenant.isBid())
+        .covenant.getHash(3);
+      allBids = await wallet.getBids();
+      assert.strictEqual(allBids.length, 2);
+      const winningBlindFromWallet = allBids.find(b =>
+        b.blind.equals(winningBlindFromMtx)
+      );
+      assert(winningBlindFromWallet);
+
+      // Check
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 4);
+      assert.strictEqual(bal.coin, 4);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, 2 * lockup);
+      assert.strictEqual(bal.clocked, 2 * lockup);
+    });
+
+    it('should send and confirm REVEAL', async () => {
+      // Advance to reveal
+      wdb.height += network.names.biddingPeriod;
+      const reveal = await wallet.sendMTX(unsentReveal, null);
+      uTXCount++;
+
+      // Check
+      let bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 5);
+      assert.strictEqual(bal.coin, 5);
+      assert.strictEqual(bal.confirmed, fund - cTXCount * fee);
+      assert.strictEqual(bal.unconfirmed, fund - uTXCount * fee);
+      assert.strictEqual(bal.ulocked, lockup + value);
+      assert.strictEqual(bal.clocked, 2 * lockup);
+
+      // Confirm REVEAL
+      const block = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+      await wallet.txdb.add(reveal, block);
+      cTXCount++;
+
+      const revealValueFromMtx = reveal.outputs.find(o => o.covenant.isReveal())
+        .value;
+      let allReveals = await wallet.getReveals();
+      assert.strictEqual(allReveals.length, 1);
+      const revealFromWallet = allReveals.find(
+        b => b.value === revealValueFromMtx
+      );
+      assert(revealFromWallet);
+
+      // Check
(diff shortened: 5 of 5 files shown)


tynes:
This is a good idea, concept ack for sure. I think it would be possible to create invalid transactions if the `force` param was used incorrectly.

pinheadmz:
Great work on this!
Also kinda closes https://github.com/handshake-org/hsd/issues/93 or at least should be linked.

A few questions:

1. You modified bid, reveal and redeem in the rpc module but in the http module, only the corresponding endpoint for reveal was modified. I think we should either just pick one set (rpc or http) and the leave the other alone, or do both equally. Sorry this is an annoying aspect of hsd's multi-API structure. 

2. In the http endpoint we also have `broadcast` and `sign` options, and note how `sign` is required when broadcasting:
```
assert(broadcast ? sign : true, 'Must sign when broadcasting.');
```
...should `force` also require that `sign=true`? Or `broadcast=false`?

3. I think we should not add a `force` to redeem. There is no time constraint on that covenant so it's not urgent enough to need an external agent (watchtower) to broadcast on a user's behalf. In addition, in your PR you correctly bypass the auction-state check but the wallet will still try to determine if it lost the auction which could be funky if the auction isn't over yet.

4. REDEEMs spend from REVEALs - and if we are presigning a reveal before broadcasting that means the UTXO wont actually be in the wallets database yet, and so it wont be able to construct a REDEEM even if there is a `force` option, it just wont have the coin to spend. This could also be an issue if a user tries to pre-sign BIDs and REVEALs both before the auction even starts - the wallet will not be able to construct the REVEAL until the BID has been seen by the wallet (successfully broadcast). This is not impossible to do but I think it will require more tools, maybe outside of hsd, to construct a complete chain of pre signed, non broadcasted covenants

So due to 3 & 4 I kinda think we should just limit this PR to presigned REVEALs. The operation won't work unless the corresponding BID has been broadcast and processed by the wallet and if we create a scenario where the BIDs don't exist in the wallet yet we're going to increase complexity and user confusion. REVEAL is the real crucial one anyway (potential fund loss) and the more urgent issue. I can see a version of Bob where every time you place a BID, the wallet  also returns the pre-signed reveal tx in hex to the user. User can send the blob via SMS or something to a full node service when bidding ends or any time before.

TLD-O:
> 1. You modified bid, reveal and redeem in the rpc module but in the http module, only the corresponding endpoint for reveal was modified. I think we should either just pick one set (rpc or http) and the leave the other alone, or do both equally. Sorry this is an annoying aspect of hsd's multi-API structure.

The only assert I had in mind was for reveal, hence I just modified the reveal endpoint. If there is additional assert, will add them. If there is something wrong in my reasoning, let me know.
Still, I forgot to change the `fromValidator` function to include the force option. Will change this.


> 2. In the http endpoint we also have `broadcast` and `sign` options, and note how `sign` is required when broadcasting:
> ```
> assert(broadcast ? sign : true, 'Must sign when broadcasting.');
> ```
> ...should `force` also require that `sign=true`? Or `broadcast=false`?

Well we could enforce the use of `broadcasting=false` to at least be sure it is not broadcasted right away. For `sign`, requiring it or not is OK with me.

> 3. I think we should not add a `force` to redeem. There is no time constraint on that covenant so it's not urgent enough to need an external agent (watchtower) to broadcast on a user's behalf. In addition, in your PR you correctly bypass the auction-state check but the wallet will still try to determine if it lost the auction which could be funky if the auction isn't over yet.

Didn't see this, will dig a little on that and see if I can find a workaround or will abandon the REDEEM.

> 4. REDEEMs spend from REVEALs - and if we are presigning a reveal before broadcasting that means the UTXO wont actually be in the wallets database yet, and so it wont be able to construct a REDEEM even if there is a `force` option, it just wont have the coin to spend. This could also be an issue if a user tries to pre-sign BIDs and REVEALs both before the auction even starts - the wallet will not be able to construct the REVEAL until the BID has been seen by the wallet (successfully broadcast). This is not impossible to do but I think it will require more tools, maybe outside of hsd, to construct a complete chain of pre signed, non broadcasted covenants

Will make some tests and see what I can come up with.

> So due to 3 & 4 I kinda think we should just limit this PR to presigned REVEALs. The operation won't work unless the corresponding BID has been broadcast and processed by the wallet and if we create a scenario where the BIDs don't exist in the wallet yet we're going to increase complexity and user confusion. REVEAL is the real crucial one anyway (potential fund loss) and the more urgent issue.

The goal I was aiming was to automate the auction as much as possible. If we want to do it covenant by covenant instead of multiple at the same time, I am OK. I agree REVEAL is the more important, I just think if we can add the same possibility to other covenants, we should do it.

> I can see a version of Bob where every time you place a BID, the wallet also returns the pre-signed reveal tx in hex to the user. User can send the blob via SMS or something to a full node service when bidding ends or any time before.

Exactly the kind of of Use Case I am going for, except I was thinking of sending the hex to a website. More or less the same.

pinheadmz:
The only problem with adding another parameter is that the rpc commands expect params to be positional-based (not named, like the JSON 2.0 spec actually requires 😬 ) Which means that if you want to include a `force` value you MUST also include an `account` value:

Usage:
`hsw-rpc createreveal <name> <account> <force>`

If you neglect the `account` parameter, then the word "true" (or whatever you think you put for `force`) will be interpreted as the account you want the wallet to create the TX from.

The nice thing about the HTTP API is the arguments are named not just positional (there are no CLI commands for these endpoints yet):
```
curl http://x:api-key@127.0.0.1:14039/wallet/$id/reveal \
  -X POST \
  --data '{
    "name": "<name>",
    "force": true,
  }'
```

The other thing I'm worried about is the "broadcast anyway" logic in fullnode:

https://github.com/handshake-org/hsd/blob/a1409dc48b41b6f23d389d9448ecbde551502688/lib/node/fullnode.js#L339-L344

This is something I hope we can get rid of ASAP, it only causes problems. And if we are going to make it even easier for users to generate invalid transactions, it really must be taken out.

Also, the `rpc createreveal` endpoint is [not documented yet](https://github.com/handshake-org/handshake-org.github.io/issues/64) but if we update it like this we should add it to the docs ASAP.

And finally, my vote is to just update REVEALs for now. The 52-week name rollout is almost over and so the use case of pre-signing BIDs before an auction starts is going to be over soon anyway. The urgent use-case is preventing funds loss from users who cant open Bob and click a button inside of a 10-day window ;-)

The bigger idea of creating a chain of BID->REVEAL->REGISTER/REDEEM all in advance I think is cool, and I think ideally that can be constructed with more complicated rpc calls based on `createrawtransaction` (see long discussion in https://github.com/handshake-org/hsd/pull/465 as well) or with an external application.

Note that REGISTER (as well as RENEW) covenants must commit to a block hash from the past 6 months, to prevent excessive pre-signing.

TLD-O:
I encountered some difficulties with the creation of the REVEAL transaction based on the BID created before (not able to get the bid blind from txdb). To avoid this, I changed my approach and created a new specific HTTP API endpoint. Down the line, we could use this endpoint to also create the remaining transactions (redeem & register).

The new endpoint will create the BID and the REVEAL based on the created bid. Had to change the way `makeReveal` and `fill` work to let me specify everything I needed. 

Happy to make any changes that could simplify the code.

lib/wallet/wallet.js
```diff
@@ -1820,14 +1824,77 @@ class Wallet extends EventEmitter {
...
+    const rawName = Buffer.from(name, 'ascii');
+    const nameHash = rules.hashName(rawName);
+    const bidBlind = new BlindBid();
+    bidBlind.name = name;
+    bidBlind.nameHash = nameHash;
+    bidBlind.prevout = Outpoint.fromTX(bid, 0);
+    bidBlind.value = bid;
+    bidBlind.lockup = lockup;
+    bidBlind.blind = bid.outputs[0].covenant.getHash(3);
+    bidBlind.own = true;
```
pinheadmz, line 1867: Even though the BID transaction has not been broadcast yet and it's coins are not yet known to the wallet, at this point in time the `bidBlind` should actually already be saved and available in the `txdb`. It is generated inside `makeBid()` which is called by `_createBid()` up on line 1855:

https://github.com/handshake-org/hsd/blob/4042141f1aa28a887179fd0ddbf78489b053dedf/lib/wallet/wallet.js#L1739

What I think might work here is, after you `_createBid()`, loop through the outputs of that MTX and find the actual BID covenant (it could be at any index). When you have the BID covenant you can pull the blind value from it and then:

```js
      const blind = coin.covenant.getHash(3);
      const bv = await this.getBlind(blind);
```

Grabbing the output of the BID TX will also enable you to get the coin you want to pass to `mtx.fund()` (see other comment)

In fact, it may make sense to just leave the existing `makeReveal()` function alone entirely, but copy the output construction logic we need to this new function.

I think if this approach works, it'll mean you don't need to do the extra module export stuff from TXDB that you got above.

lib/wallet/wallet.js
```diff
@@ -1820,14 +1824,77 @@ class Wallet extends EventEmitter {
...
+    const coins = [];
+    const bidCoin = Coin.fromTX(bid, 0, this.wdb.height);
```
pinheadmz, line 1869: See my bigger comment above, there is actually no guarantee that the BID covenant will always be at index `0`, we'll have to find it to be sure.

lib/wallet/wallet.js
```diff
@@ -1820,14 +1824,77 @@ class Wallet extends EventEmitter {
...
+    coins.push(bidCoin);
+
+    const acct = options ? options.account : null;
+    const force = options ? options.force || false : false;
```
pinheadmz, line 1873: I think this can just be `const force = options.force` ?

But actually if we go with my suggestion of copying the reveal logic to this new function and leaving `makeReveal()` alone as-is, we won't need the force option at all. (This function will just always create a pre-signed reveal without doing any context checks).

test/wallet-test.js
```diff
@@ -2869,4 +2869,298 @@ describe('Wallet', function() {
...
+
+      // Confirm REVEAL
+      const block2 = {
+        height: wdb.height + 1,
+        hash: Buffer.alloc(32),
+        time: Date.now()
+      };
+      await wallet.txdb.add(reveal2, block2);
+      cTXCount++;
+
```
pinheadmz, line 3084: Just to be extra meticulous, I think it'd be nice to also see `await wallet.getBids()` after each BID is confirmed and `await wallet.getReveals()` after each reveal is confirmed, and then check that the values are correct and that the presigned BID and presigned REVEAL are definitely coming up just like the regular BID and REVEAL. What do you think? Couldn't hurt to cover that here.

pinheadmz (changes requested):
DUDE. Really really awesome work on this. Thank you for having the patience for discussion and review, and for following through with good tests.

Besides a few nits and maybe bike-shedding over function names, my biggest suggestion is to leave `makeReveal()` alone, and just copy that logic into your new function. You already had to copy/paste it inside an if/else, I think we could have better control if the code path was just brand new rather than having options like `force` in there.

Also see my note about getting the BID covenant and deriving the blind value from there - that should save a lot of work arounds like exporting extra stuff from the `txdb` module.

Again, really great work and feel free to disagree with any of my comments or push back or ask another reviewer for second opinions!

pinheadmz:
Oh one more thing - I think we should have a test in the `wallet-http-test` module as well - and make sure that the raw hex we get back from the new API endpoint can be sent right back to the node HTTP `/broadcast` endpoint, and confirmed as valid TXs.

lib/wallet/http.js
```diff
@@ -1064,6 +1064,35 @@ class HTTP extends Server {
       return res.json(200, mtx.getJSON(this.network));
     });
 
+    this.post('/wallet/:id/auction-txs', async (req, res) => {
```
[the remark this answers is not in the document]
TLD-O: Will do for the comment ! For the name, in fact I was aiming for one single entry point for the whole process down the line, hence a generic name. In the futur, I hope to add redeem & register, just because it would be nice to just create everything at the start, ask for broadcast at good height and then forget about it.

lib/wallet/http.js
```diff
@@ -1064,6 +1064,35 @@ class HTTP extends Server {
...
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+      const bid = valid.u64('bid');
+      const lockup = valid.u64('lockup');
+      const passphrase = valid.str('passphrase');
+      const sign = valid.bool('sign', true);
+
+      assert(name, 'Name is required.');
+      assert(bid != null, 'Bid is required.');
+      assert(lockup != null, 'Lockup is required.');
```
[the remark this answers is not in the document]
TLD-O: Yeah I remember it, but I don't think we should enforce any particular value. Big warning or whatever you can come up with is OK with me though. I think this "1 HNS" blind minimum has to be enforced by apps calling this API.

lib/wallet/wallet.js
```diff
@@ -1195,12 +1197,14 @@ class Wallet extends EventEmitter {
     if (rate == null)
       rate = await this.wdb.estimateFee(options.blocks);
 
-    let coins;
-    if (options.smart) {
-      coins = await this.getSmartCoins(options.account);
-    } else {
-      coins = await this.getCoins(options.account);
-      coins = this.txdb.filterLocked(coins);
+    let coins = specificCoins;
```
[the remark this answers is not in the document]
TLD-O: So basically, when using fund from my initial function, I will pass a copy of options with an additional attribute (as to not alter the current options object) ? Why not

lib/wallet/wallet.js
```diff
@@ -1820,14 +1824,77 @@ class Wallet extends EventEmitter {
     }
   }
 
+  /**
+   * Create and finalize a bid & a reveal (in advance)
+   * MTX with a lock.
+   * @param {String} name
+   * @param {Number} value
+   * @param {Number} lockup
+   * @param {Object} options
```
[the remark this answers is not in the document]
TLD-O: Will do !

lib/wallet/wallet.js
```diff
@@ -1820,14 +1824,77 @@ class Wallet extends EventEmitter {
...
+   */
+
+  async createAuctionTxs(name, value, lockup, options) {
```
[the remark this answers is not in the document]
TLD-O: Same logic behind the HTTP API. Will update the name if we change it

lib/wallet/wallet.js
TLD-O: Damn I know I had to change it before sending it, but forgot ! Nice catch

lib/wallet/wallet.js
TLD-O: The thought behind altering reveal is to keep linked code in the same place. Sure it add complexity to the function, but it limits the odds of a function to be "forgotten" when introducing changes. Will think about it

test/wallet-test.js
TLD-O: Alright, will do !

TLD-O:
> 
> 
> Oh one more thing - I think we should have a test in the `wallet-http-test` module as well - and make sure that the raw hex we get back from the new API endpoint can be sent right back to the node HTTP `/broadcast` endpoint, and confirmed as valid TXs.

In fact I do have one that seems similar (not sure about the broadcast though, I use the `sendrawtransaction` RPC call instead but I think it does not verify it ?). I just didn't pushed it. As far as I can tell, I would need to create a new entry to hs-client to call the new endpoint and as it is another module, well, don't really know what to do ...

TLD-O:
I think I addressed most of your remarks except
- Name of API endpoint / function as to me this endpoint will be used to create the whole chain of transactions sometime in the futur
- Enforcing a blind value as this seems more like something app/client may want to set up (or not), not sure we should enforce this ourself
- `wallet-http-test` as I don't know how to call the new API endpoint, got a prototype of it working but had to change the `hs-client` dependency locally

tynes:
> wallet-http-test as I don't know how to call the new API endpoint, got a prototype of it working but had to change the hs-client dependency locally

You should be able to `.post` directly on the client, see: https://github.com/handshake-org/hs-client/blob/d82611cead27d7f9586d878438a23ad2005ff8af/lib/node.js#L86

lib/wallet/wallet.js
```diff
@@ -1849,37 +1850,40 @@ class Wallet extends EventEmitter {
...
    * @param {Number} lockup
    * @param {Object} options
+   * @returns {{ bid: MTX; reveal: MTX;}} {bid, reveal}
    */
 
   async _createAuctionTxs(name, value, lockup, options) {
     const bid = await this._createBid(name, value, lockup, options);
 
+    const bidOuputIndex = bid.outputs.findIndex(o => o.covenant.isBid());
+    const bidCoin = Coin.fromTX(bid, bidOuputIndex, this.wdb.height);
```
pinheadmz, line 1860: I think `height` should be `-1` for an unconfirmed coin.

lib/wallet/wallet.js
```diff
@@ -1862,40 +1930,40 @@ class Wallet extends EventEmitter {
     const mtx = new MTX();
 
     for (const {prevout, own} of bids) {
-      if (!own)
-        continue;
+        if (!own)
```
pinheadmz, line 1933: what happened to the whitespace in this function? :-)

test/wallet-test.js
```diff
@@ -2869,4 +2869,338 @@ describe('Wallet', function() {
...
+      await wallet.txdb.add(winningBid, block);
+      cTXCount++;
+
+      const winningBlindFromMtx = winningBid.outputs
+        .find(o => o.covenant.isBid())
+        .covenant.getHash(3);
+      let allBids = await wallet.getBids();
+      assert.strictEqual(allBids.length, 1);
+      const winningBlindFromWallet = allBids.find(
+        b => b.blind.toString('hex') === winningBlindFromMtx.toString('hex')
```
pinheadmz, line 3003: since `b.blind` and `winningBlindFromMtx` are both buffers, you should be able to test equality with `b.blind.equals(winningBlindFromMtx)`

pinheadmz:
A rebase on master branch should fix the failed test, which was patched here: #531

lib/wallet/wallet.js
```diff
@@ -1820,6 +1824,70 @@ class Wallet extends EventEmitter {
...
+
+  async _createAuctionTxs(name, value, lockup, options) {
+    const bid = await this._createBid(name, value, lockup, options);
+
+    const bidOuputIndex = bid.outputs.findIndex(o => o.covenant.isBid());
+    const bidCoin = Coin.fromTX(bid, bidOuputIndex, this.wdb.height);
+
+    // Prepare the data needed to make the reveal in advance
+    const rawName = Buffer.from(name, 'ascii');
+    const nameHash = rules.hashName(rawName);
```
pinheadmz, line 1864: the `nameHash` is actually already in the BID covenant as well, `covenant.getHash(0)`

lib/wallet/wallet.js
TLD-O: Well, some leftovers from switching from altering makeReveal to copying it elsewhere. Will change it back !

test/wallet-test.js
TLD-O: Wasn't sure how to test it, thanks for pointing it out

lib/wallet/wallet.js
TLD-O: Alright, will clean it up

lib/wallet/wallet.js
TLD-O: Wasn't sure what to set here, good to know

TLD-O:
> 
> 
> A rebase on master branch should fix the failed test, which was patched here: #531

Rebased, still got the same 6 tests failing (Miner & the 2 `should have the same DB root`)

Made the changes you pointed out in your last review and adding a test for the HTTP endpoint (that may be a little bit overkill btw)

test/wallet-http-test.js
```diff
@@ -278,7 +278,10 @@ describe('Wallet HTTP', function() {
     });
 
     let entered = false;
-    node.mempool.on('tx', () => {
+    node.mempool.on('tx', (tx) => {
+      if(json.hex !== tx.toHex()) {
+        return;
```
pinheadmz, line 283: What's this for? I think this test is designed to fail if the mempool ever emits any `'tx'` event (because we are just creating and not broadcasting).

pinheadmz: Actually I think we should double check this logic, looks like the event listener should be set _before_ the `createOpen()`. @tynes not sure if JS hoists these - but what we want to test is that the event is not emitted, correct?

test/wallet-http-test.js
```diff
@@ -839,6 +845,75 @@ describe('Wallet HTTP', function() {
     matchTxId(auction.reveals, state.reveals[1].hash);
   });
 
+  it('should create a bid and a reveal (reveal in advance)', async () => {
+    const balanceBeforeTest = await wallet.getBalance(null);
```
pinheadmz, line 849: `getBalance()` can be left empty I think, `null` is probably fine but redundant.

test/wallet-http-test.js
```diff
@@ -839,6 +845,75 @@ describe('Wallet HTTP', function() {
...
+    const lockConfirmedBeforeTest = balanceBeforeTest.lockedConfirmed;
+    const lockUnconfirmedBeforeTest = balanceBeforeTest.lockedUnconfirmed;
+    assert.equal(
+      balanceBeforeTest.lockedConfirmed - lockConfirmedBeforeTest,
+      0
+    );
+    assert.equal(
+      balanceBeforeTest.lockedUnconfirmed - lockUnconfirmedBeforeTest,
+      0
+    );
```
pinheadmz, line 859: what are you testing here?

test/wallet-http-test.js
```diff
@@ -839,6 +845,75 @@ describe('Wallet HTTP', function() {
...
+
+    walletAuction = await wallet.getAuctionByName(name);
+    const revealFromWallet = walletAuction.reveals.find(
+      b => b.prevout.hash === auctionTxs.reveal.hash
+    );
+    assert(revealFromWallet);
+
+    const b6 = await wallet.getBalance(null);
+    assert.equal(b6.lockedConfirmed - lockConfirmedBeforeTest, bidValue);
+    assert.equal(b6.lockedUnconfirmed - lockUnconfirmedBeforeTest, bidValue);
```
pinheadmz, line 914: ok awesome up to here, I think another `getnameinfo` check here would be good after generating enough blocks to clsoe the auction, then check the name's `owner`. That way the test asserts that the reveal worked and the auction was won by the bidder.

test/wallet-http-test.js
TLD-O: The problem here was that when I was running my test and tried to broadcast the TX made in advance, it failed because of this check ...

test/wallet-http-test.js
TLD-O: Will try it

test/wallet-http-test.js
pinheadmz: hmmmmmmmmmm thats not good if tests interfere with each other. These tests maybe we can remove the event actually, since we test `assert.ok(!mempool.includes(json.hash));` anyway. OR if the event is really required, we need to remove the listener at the end of the test.

test/wallet-http-test.js
TLD-O: Might be an unwanted extra, just making sure I got the rights number to make comparison down the test without being wrong because of previous tests altering the balance. Could remove it, let me know which you prefer.

test/wallet-http-test.js
TLD-O: Alright, will do

test/wallet-http-test.js
TLD-O: So, should I remove this part of the 2 tests ?

TLD-O:
Thank you for your time, really appreciate it. Cleaned up & improved the test as requested.

lib/wallet/http.js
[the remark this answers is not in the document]
tynes: I think that the name of the endpoint should be `POST /wallet/:id/auction/:name`. This more closely follows the style of the other endpoints, as none of them have dashes in them. It allows for the possibility of additional transactions to be added besides bid and reveal in the future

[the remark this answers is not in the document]
pinheadmz: `POST /wallet/:id/auction/:name` I like this too

lib/wallet/http.js
```diff
@@ -1084,6 +1084,36 @@ class HTTP extends Server {
...
+
+      const options = TransactionOptions.fromValidator(valid);
+      const {
```
tynes, line 1101: Nit: the codebase usually uses arrays when returning multiple things and destructuring them. I think returning an object is fine in this case since it could return more things in the future (additional transactions). I don't think there is a huge need to destructure out `bidMTX` and `revealMTX`, maybe something like this would be more in style:

```js
const auction = await req.wallet.createAuction(name, bid, lockup, options);

if (sign) {
  await req.wallet.sign(auction.bid, passphrase);
  await req.wallet.sign(auction.reveal, passphrase);
}

return res.json(200, {
  bid: auction.bid.getJSON(this.network),
  reveal: auction.reveal.getJSON(this.network)
});
```

lib/wallet/wallet.js
```diff
@@ -1195,12 +1197,14 @@ class Wallet extends EventEmitter {
     if (rate == null)
       rate = await this.wdb.estimateFee(options.blocks);
 
-    let coins;
```
tynes, line 1198: I think there should be a typecheck on `options.coins` to make sure that it is an array

lib/wallet/http.js
TLD-O: Alright, changed it !

lib/wallet/wallet.js
TLD-O: Assertion added

TLD-O:
Thanks @tynes for your review, made the changes you highlighted.

Also fixed how I get the NameState. When testing it earlier I had a null exception, looked at makeBid to find a fix.

lib/wallet/http.js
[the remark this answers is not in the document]
pinheadmz: Ok my final opinion on this is naming the endpoint `POST /wallet/:id/auction`.
I think you can keep the function name `createAuctionTxs()`

@tynes as I look through the rest of the wallet HTTP API, this matches the pattern. Adding `/:name` to queries seems to be used just in `GET` requests, and all the `POST` requests expect the name and all other parameters to be in the query JSON data.

lib/wallet/wallet.js
```diff
@@ -1820,6 +1825,73 @@ class Wallet extends EventEmitter {
...
+
+    const bidOuputIndex = bid.outputs.findIndex(o => o.covenant.isBid());
+    const bidOutput = bid.outputs[bidOuputIndex];
+    const bidCoin = Coin.fromTX(bid, bidOuputIndex, -1);
+
+    // Prepare the data needed to make the reveal in advance
+    const nameHash = bidOutput.covenant.getHash(0);
+    let ns = await this.getNameState(nameHash);
+    if (!ns)
+      ns = await this.wdb.getNameStatus(nameHash);
```
pinheadmz, line 1868: Yeah I think this is right. If the wallet doesn't have a local copy of the namestate that means nothing has been confirmed in the wallet for this name: It doesn't have any bids yet, etc. So it calls `getNameStatus()` which actually reaches out the full node instead and tries to get the name's status from the blockchain.

HOWEVER, I don't think you need this at all, because you already got this info when you made the bid! If you look below, all that `ns` is used for is `ns.height`... which is already in the BID covenant!

You already have this above:
`const nameHash = bidOutput.covenant.getHash(0);`
Pretty sure you can follow that with:
`const height = bidOutput.covenant.getU32(1);`

and then you don't need to get namestate anymore here at all.

lib/wallet/wallet.js
```diff
@@ -1820,6 +1825,73 @@ class Wallet extends EventEmitter {
...
+
+    const coins = [];
+    coins.push(bidCoin);
+
+    const blind = bidOutput.covenant.getHash(3);
+    const bv = await this.getBlind(blind);
+    if (!bv)
+      throw new Error('Blind value not found.');
```
pinheadmz, line 1876: It's good you have this here, but it would be pretty crazy bug since you just made the bid (and the blind) 12 lines up!

test/wallet-test.js
```diff
@@ -2869,4 +2869,338 @@ describe('Wallet', function() {
...
+
+      bal = await wallet.getBalance();
+      assert.strictEqual(bal.tx, 3);
+      assert.strictEqual(bal.coin, 3);
+      assert.strictEqual(bal.confirmed, fund - (cTXCount * fee));
+      assert.strictEqual(bal.unconfirmed, fund - (uTXCount * fee));
+      assert.strictEqual(bal.ulocked, lockup);
+      assert.strictEqual(bal.clocked, lockup);
+
+      const auctionsTxs = await wallet.createAuctionTxs(name, secondHighest, lockup, {hardFee: fee, force:true});
```
pinheadmz, line 3015: do we need `force: true` ? I don't think that does anything.

pinheadmz: Also I think you should swap the values of the two bids. The test would be much more useful if the new function created the WINNING bid, and we can assert that your bids and reveals won the auction.

test/wallet-test.js
```diff
@@ -2869,4 +2869,338 @@ describe('Wallet', function() {
...
+      await wallet.txdb.add(losingBid, block);
+      cTXCount++;
+
+      const losingBlindFromMtx = losingBid.outputs
+        .find(o => o.covenant.isBid())
+        .covenant.getHash(3);
+      allBids = await wallet.getBids();
+      assert.strictEqual(allBids.length, 2);
+      const losingBlindFromWallet = allBids.find(
+        b => b.blind.toString('hex') === losingBlindFromMtx.toString('hex')
```
pinheadmz, line 3046: can do that `Buffer.equals()` here again

test/wallet-http-test.js
```diff
@@ -839,6 +837,73 @@ describe('Wallet HTTP', function() {
     matchTxId(auction.reveals, state.reveals[1].hash);
   });
 
+  it('should create a bid and a reveal (reveal in advance)', async () => {
+    const balanceBeforeTest = await wallet.getBalance();
+    const lockConfirmedBeforeTest = balanceBeforeTest.lockedConfirmed;
+    const lockUnconfirmedBeforeTest = balanceBeforeTest.lockedUnconfirmed;
+
+    await wallet.createOpen({
+      name: name
```
pinheadmz, line 846: nit: could also just be `createOpen({name})`

pinheadmz (changes requested):
Great work on this as always, we are getting close!
Also if possible it would be great to squash a few of these commits together so the commit history is a bit cleaner and more direct. I'll leave that up to you, I can help or we can squash when we merge.

pinheadmz:
I did a test of this endpoint using regtest. The first time I ran the test, I actually didn't set the lockup high enough (honest mistake actually) and ended up with an invalid REVEAL, because it spent the exact same coin as the BID.

SO... I am worried about users doing this without providing sufficient fee in the lockup for the REVEAL. They're not getting a warning even, and they're not going to realize their REVEAL will be invalid. Because this endpoint is designed for "lazy" users, it's unlikely they will even realize they need to create a whole new REVEAL before time runs out.

There is one other complication we could add which is, when creating the REVEAL -- we LOCK the coins spent in the BID. Check out the function `lib/wallet/wallet.lockCoin()`, that could be used here to help in some cases. We might even want to LOCK the coins spent into the REVEAL as well... although the locked coins list gets reset when a node restarts... so there still will be a possibility that the user sends a TX that invalidates the REVEAL by double spending an input they need to pay the fee.

In any case, when talking about pre-signing transactions, we really need to keep track of which coins are being spent. The REVEAL may not be broadcast for several days, and the user's wallet **MUST** keep track of any coins spent in that REVEAL and remember not to use them before the REVEAL is confirmed.

The absolute best way to deal with this is to broadcast the BID immediately (wallet won't double spend those inputs) and require that there be sufficient lockup to fund the REVEAL.

Might even want to `assert()` that the REVEAL only spends one input (the BID itself). 



Here's my script for testing: https://gist.github.com/pinheadmz/709d6db66a946a9ae274fc793fb4b8d4
If you set the BID too high or LOCKUP too low, the script will fail because of the double-spend issue

test/wallet-http-test.js
TLD-O: Format done

test/wallet-test.js
TLD-O: Changed !

test/wallet-test.js
TLD-O: Done

lib/wallet/wallet.js
TLD-O: As discussed, left it

lib/wallet/wallet.js
TLD-O: Thanks for the tip, changed it

test/wallet-http-test.js
```diff
@@ -278,9 +278,8 @@ describe('Wallet HTTP', function() {
     });
 
     let entered = false;
-    node.mempool.on('tx', () => {
+    node.mempool.on('tx', (tx) => {
```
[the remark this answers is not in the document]
TLD-O: Rebased, all good now

test/wallet-test.js
```diff
@@ -2869,4 +2869,338 @@ describe('Wallet', function() {
...
+      assert.strictEqual(bal.clocked, value + secondHighest);
+    });
+
+    it('should send and confirm REDEEM', async () => {
+      // Advance to close
+      wdb.height += network.names.revealPeriod;
+      const redeem = await wallet.sendRedeem(name, {hardFee: fee, force:true});
+      uTXCount++;
+
+      // // Check
```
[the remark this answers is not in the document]
TLD-O: Nice catch !

lib/wallet/http.js
```diff
@@ -1084,6 +1084,45 @@ class HTTP extends Server {
...
+      const bid = valid.u64('bid');
+      const lockup = valid.u64('lockup');
+      const passphrase = valid.str('passphrase');
+      const sign = valid.bool('sign', true);
+      const broadcastBid = valid.bool('broadcastBid');
+
+      assert(name, 'Name is required.');
+      assert(bid != null, 'Bid is required.');
+      assert(lockup != null, 'Lockup is required.');
+      assert(broadcastBid != null, 'broadcastBid is required');
```
pinheadmz, line 1100: micro-nit: missing period at the end (all the other error statements have it)

test/wallet-test.js
```diff
@@ -2869,4 +2869,343 @@ describe('Wallet', function() {
...
     });
   });
+
+  describe('Create auction-related TX in advance', function () {
+    const network = Network.get('regtest');
+    const workers = new WorkerPool({ enabled });
+    const wdb = new WalletDB({ network, workers });
+    // This test executes a complete auction for this name
+    const name = 'satoshi-in-advance';
+    // There will be one bid. One from our wallet with a lockup and bid value made in advance
```
pinheadmz, line 2879: nit: there are actually two bids in this test.

pinheadmz:
Awesome dude. Just one more commit and I'll approve:
- fix the two nits above
- describe your new api endpoint in the changelog! See https://github.com/handshake-org/hsd/pull/548/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed for an example. Doesn't have to be complete documentation, just mention the new endpoint and 1-2 lines describing.
 
maybe this:

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 9d52a155d..8e0713ddb 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # HSD Release Notes & Changelog
 
+## unreleased
+
+### Wallet API changes
+
+- Adds new endpoint `POST /auction` that produces BID and REVEAL transactions for
+a given name, optionally pre-signs them and optionally broadcasts the BID. The
+pre-signed REVEAL can be given to a trusted party to broadcast during the REVEAL
+period without further user interaction.
+
 ## v2.3.0
 
 ### Node changes
```

lib/wallet/http.js
TLD-O: Had to re-read it two times to spot it --"

test/wallet-test.js
TLD-O: Reworded the whole comment to make things clearer

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # HSD Release Notes & Changelog
 
+## unreleased
+
+- Adds new wallet HTTP endpoint `/wallet/:id/auction` based on `POST /wallet/:id/bid`.
+It requires an additional parameter `broadcastBid` set to either true or false.
+This action returns a bid and its corresponding reveal, the reveal being prepared in advance.
+The bid will be broadcaster either during the creation (`broadcastBid=true`) or at a later time
```
pinheadmz, line 8: s/broadcaster/broadcasted

TLD-O: changed

pinheadmz (approved):
ACK c969e8f89f29a83d4a45d2dacea18bc7ebc9f627

```
Hash: SHA256

ACK c969e8f89f29a83d4a45d2dacea18bc7ebc9f627
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmAZy3gACgkQ5+KYS2KJ
yTprnA//XxWZJ0OSCOo6NIZW6x8pjzdyli51Scr6r2+G4o7wIbhqIa4wciepUyQl
u14ls/UmwMRCYd0nm45hEf2YC/rC+U1pqD6dFxSTNqQYfJLVcReCMxQguZFgL+WI
XH7VOUVz6LiczjRqQfKzYgumeGT9ee4hnw0iNkdFUu6HJj03UBNZvqukhvNVZsE8
KodpH7hN/4hBxe/kyhcsZ+BbCc3i3tZLoBWNUQaD6BqQhluYeNHHGqV0ORY8UC16
qLec7n/dJgUwHbGZG7oVP3RXIPWGD98e8P2xCeDLHBSuCZ3fmnkXdlihsXGiGUWo
76tt3qEHSNiH68An9EcFenA7dfoSXVRKwP06loVtA1QpU+Gu1ci4ddzeXF9tfnqd
USi2Hy7YZDw1R2e/wA8zNWPNPaPheO5MK2xxXfO34aZiut8TAgfJHThT4PzUAzXp
CDL+4gHbq+VRg/X8xwjSjJjiEnNbnKkpZ0RHOLmc3tHMeqYLRaj7/VKKp6H4YIRo
/yA/PFIBnXlxB+SHVyCTXwtoMEUHPafoD9o2BwO1LqNlGzSwNHMbOuRw6fzB4UGb
Pw529HH9LKvCJ/YQ1qxkxXx8E0c788Ix0MjaW5N4b+MGo0tpSeRsmUgA7pwG5wn2
6tDqKbBfQEG+PdLhDtYnh0x2XREdhqSOztUA97au0UK5+qyeU+8=
=cLPv
-----END PGP SIGNATURE-----
```

</details>

tynes (approved):
Good work! Thank you for your contributions

PR #533, opened 2021-01-08
Avoid conflicting chain locks when rescan is triggered by new block

pinheadmz:
Fix for https://github.com/bcoin-org/bcoin/issues/1006 from upstream in bcoin.

Issue summary:
- WalletDB somehow falls behind chain (i.e. botched rescan attempt, maybe due to one of the txdb assertion errors)
- A new block is added to the chain by the network
- The chain block-adding process locks the chain
- WalletDB realizes its behind and initiates a rescan
- Rescan also requires a lock on chain, but chain is still locked and will never unlock until the rescan is done!

Bug fix summary:
Emit the `'block connect'` event synchronously so the original `'connect'` event from chain can resolve. The chain `'connect'` event is emitted _asynchronously_ meaning it waits for all its listeners to resolve before program flow continues. The chain event is proxied by `nodeclient` to the wallet as `'block connect'` and this is what ultimately triggers the reorg. 

Note that this is only an issue when the wallet is run as a plugin. If the wallet is being run as a separate node, it gets the `'block connect'` event from the node's HTTP websocket:

https://github.com/handshake-org/hsd/blob/a1409dc48b41b6f23d389d9448ecbde551502688/lib/node/http.js#L666-L680

The websocket event is obviously synchronous because it is just a one-way firing, it doesn't wait for a response at all. Indeed! The fact that one mode of operation uses synchronous events should help justify this pull request.

The side effect of this change is that in a few tests we always just expect the walletDB to be up to date with the chain when we add new blocks. This is no longer the case and probably we should always consider wallet and node operations to be asynchronous anyway. For those tests I've added a wait-er to ensure wallet is synced to chain before continuing.

Reviewers can run the new test `test/wallet-rescan-test.js` after reverting the one-line change in `lib/wallet/nodeclient.js` and observe that the test will timeout -- this is because the two conflicting locks will wait for each other forever.

### DIFF
--- a/lib/wallet/nodeclient.js
+++ b/lib/wallet/nodeclient.js
@@ -40,7 +40,7 @@ class NodeClient extends AsyncEmitter {
       if (!this.opened)
         return;
 
-      await this.emitAsync('block connect', entry, block.txs);
+      await this.emit('block connect', entry, block.txs);
     });
 
     this.node.chain.on('disconnect', async (entry, block) => {
--- a/test/auction-rpc-test.js
+++ b/test/auction-rpc-test.js
@@ -150,6 +150,11 @@ describe('Auction RPCs', function() {
     const address = (await wallet.createAddress(account)).address;
     const hashes = await util.nrpc('generatetoaddress', [num, address]);
     await util.confirmBlock(hashes.pop());
+    await common.forValue(
+      util.node.plugins.walletdb.wdb,
+      'height',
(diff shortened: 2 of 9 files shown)


pinheadmz:
Closing for now, although we still may need a fix for this. Turns out we use `emitAsync` here for a good reason (memory):

https://github.com/bcoin-org/bcoin/pull/932

PR #444, opened 2020-05-06
Fix RootServer resolution of SYNTH and GLUE records

pinheadmz:
SYNTH4 and SYNTH6 records are very special: they act like NS records that contain their own IP addresses, encoded in base32. They can therefore be resolved without any lookup whatsoever. It's a neat trick to save blockchain space and avoid using useless NS+glue.

This PR fixes three bugs in the hsd RootServer resolution of these records:

#### 1) `binet.isIPv4(raw)` expects `raw` to be 16 bytes

The server does not currently map IPv4 to IPv6 addresses before proceeding with response.

Without patch:

```
$  dig @127.0.0.1 -p 25349 _fs0000g._synth.

[error] (ns) Assertion failed.
    at Object.isMapped (/Users/matthewzipkin/Desktop/work/hsd/node_modules/binet/lib/ip.js:734:3)
    at Object.isIPv4 (/Users/matthewzipkin/Desktop/work/hsd/node_modules/binet/lib/ip.js:745:16)
    at RootServer.response (/Users/matthewzipkin/Desktop/work/hsd/lib/dns/server.js:303:14)
    at RootServer.resolve (/Users/matthewzipkin/Desktop/work/hsd/lib/dns/server.js:398:28)
    at RootServer.answer (/Users/matthewzipkin/Desktop/work/hsd/node_modules/bns/lib/server/dns.js:249:28)
    at RootServer.handle (/Users/matthewzipkin/Desktop/work/hsd/node_modules/bns/lib/server/dns.js:316:24)
    at Server.<anonymous> (/Users/matthewzipkin/Desktop/work/hsd/node_modules/bns/lib/server/dns.js:72:20)
    at Server.emit (events.js:210:5)
    at Socket.<anonymous> (/Users/matthewzipkin/Desktop/work/hsd/node_modules/bns/lib/internal/net.js:73:12)
    at Socket.emit (events.js:210:5)
```

#### 2) cache does not properly detect the `_synth.` pseudo-TLD

The root server caches the first `_synth.` response and returns the same answer for all further queries, even though the design of the synth record is to decode the "subdomain" every time the pseudo-TLD is resolved:

Without patch:

```
$ dig @127.0.0.1 -p 25349 _fs0000g._synth.
...
;; ANSWER SECTION:
_fs0000g._synth.        21600   IN      A       127.0.0.2
...

$ dig @127.0.0.1 -p 25349 _00000000000000000000000008._synth.
...
;; ANSWER SECTION:
[...]

PR #534, opened 2021-01-08
rpc dumpzone: dump HNS root zone to file

pinheadmz:
File output is standard zone file format: https://tools.ietf.org/html/rfc1035#section-5.1

Replaces https://github.com/handshake-org/hsd/pull/280
Based on https://github.com/james-stevens/handshake-bridge

Major differences from these two predecessors:
- Reads from the Urkel Tree, not the `txn` (which include record updates between the 36-block commit cycle)
- Includes `TXT` records (handshake-bridge didn't either by design or by accident by calling `Resource.toDNS()` which results in a referral-only answer)


pinheadmz:
sample:

```
melanie. 21600 IN NS ns1.melanie.
melanie. 21600 IN NS ns1.melanie.
xn--u86c8154b. 21600 IN NS ns1.xn--u86c8154b.
xn--u86c8154b. 21600 IN NS ns1.xn--u86c8154b.
cybermage. 21600 IN NS ns1.cybermage.
cybermage. 21600 IN NS ns1.cybermage.
proofofconcept. 21600 IN NS ns.proofofconcept.
proofofconcept. 21600 IN TXT "A simple social network for Handshake"
proofofconcept. 21600 IN TXT "support: hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x"
proofofconcept. 21600 IN DS 17552 8 2 BBBE70AF5CD965360442CBEF40E3299344AF493D339592B93DAA29F7 839C1D58  ; alg = RSASHA256 ; hash = SHA256
chostner. 21600 IN NS ns1.chostner.
chostner. 21600 IN NS ns1.chostner.
n08. 21600 IN NS ns1.n08.
n08. 21600 IN NS ns1.n08.
nongcun. 21600 IN NS ns1.nongcun.
nongcun. 21600 IN NS ns1.nongcun.
xn--mgbkl3fffp. 21600 IN TXT "dummy"
xn--h1alfddcd4f2a. 21600 IN TXT "dummy"
```

lib/node/rpc.js
```diff
@@ -2533,6 +2538,67 @@ class RPC extends RPCBase {
...
+      fd.end();
+      fd = null;
+      fileErr = err;
+    });
+
+    const tree = this.chain.db.tree;
+    const iter = tree.iterator(true);
+
+    let count = 0;
+    while ((await iter.next()) && (fd != null)) {
```
tynes, line 2566: Is this `fd` check redundant with the check on 2571?

tynes:
Handling a shutdown gracefully during this is also something to consider

james-stevens:
Excluding `TXT` wasn't intentional, I copied the code from `#280`, but if the name is a delegated subdomain, there should *NOT* be any `TXT` records in the parent zone for it anyway. An RFC DNS server would *NEVER* serve those records, they belong in the zone file for the sub-domain *NOT* the parent.

The parent owns the `DS` - and `NS` & any necessary IPv4 or IPv6 would be served, but all other records for a delegated subdomain would be ignored. `named-checkzone` will probably fail & moan to you about it.

So having these `TXT` and expecting them to be served & used is quite a big break of RFC :-1:

If there are `TXT` in the zone file for the sub-domain (extremely common, e.g for email SPF records), how do you expect a `TXT` request to be handled?
1. Give out ONLY the ones in the ROOT (breaks SPF)
2. Give out ONLY those in the zone file (RFC)
3. Merge both sets and give out that


By reading the data from the Urkel Tree, you probably solve a few race conditions if data is updated when the dump is run (:+1:), but it would also be nice if the SOA Serial given, when polled over UDP (say), reflected the Urkel Tree update-time, instead of just being the current date & time ... That way the zone data dumped would be consistent with the SOA Serial ... we did discuss this.

Using the Urkel Tree also means if you dump the zone on two different servers, within the same 36-block commit cycle, you should get the same zone - this is good & highly desirable for failover purposes :+1:

lib/node/rpc.js
james-stevens: probably can remove one or other, just trying to catch if the `fd.on('error',...` got triggered 

JS isn't my strongest skill, so likely all that error handling can all be done better
I think it was to try & catch things like disk-full

pinheadmz:
Added one more commit https://github.com/handshake-org/hsd/pull/534/commits/e80b835b3a05308a71fac7b5a8647573280ad085 which fixes _another_ SYNTH4/SYNTH6 decoding issue that should have been caught in https://github.com/handshake-org/hsd/pull/444

Example output of my own mainnet name `omnitude` with SYNTH records:

```
omnitude. 21600 IN NS _5l6tm80._synth.
omnitude. 21600 IN DS 26614 8 2 A20BC6F9ADA0883326A05374D0D0C0E6290CEF580D00B9B957703014 41733B7F  ; alg = RSASHA256 ; hash = SHA256
omnitude. 21600 IN NS _400hjs000l2gol000fvvsc9cpg._synth.
_5l6tm80._synth. 21600 IN A 45.77.219.32
_400hjs000l2gol000fvvsc9cpg._synth. 21600 IN AAAA 2001:19f0:5:450c:5400:3ff:fe31:2ccc
```

Also worth noting that https://github.com/handshake-org/hsd/pull/566 may affect this rpc output concerning TXT records in the root zone, but I think we decided that was ok. Currently this branch will print TXT even if NS is present:

```
proofofconcept. 21600 IN NS ns.proofofconcept.
proofofconcept. 21600 IN TXT "A simple social network for Handshake"
proofofconcept. 21600 IN TXT "support: hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x"
proofofconcept. 21600 IN DS 17552 8 2 BBBE70AF5CD965360442CBEF40E3299344AF493D339592B93DAA29F7 839C1D58  ; alg = RSASHA256 ; hash = SHA256
ns.proofofconcept. 21600 IN A 142.93.115.133
```

pinheadmz:
new TODO: some names have unknown resource versions, bypass them:

```diff
diff --git a/lib/node/rpc.js b/lib/node/rpc.js
index 1973aa98e..96f276574 100644
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2629,7 +2629,19 @@ class RPC extends RPCBase {
         continue;
 
       const fqdn = bns.util.fqdn(ns.name.toString('ascii'));
-      const resource = Resource.decode(ns.data);
+
+      let resource;
+      try {
+        resource = Resource.decode(ns.data);
+      } catch (e) {
+        this.logger.warning(
+          'could not process resource for name: %s (%s)',
+          fqdn,
+          e.message
+        );
+        continue;
+      }
+
       const zone = resource.toZone(fqdn);
       for (const record of zone) {
         if (fd == null)
```

result:

```
[debug] (node-rpc) dumpzone names processed: 2230000
[debug] (node-rpc) dumpzone names processed: 2240000
[warning] (node-rpc) could not process resource for name: roivantsciences. (Unknown serialization version: 2.)
```

lib/node/rpc.js
```diff
@@ -2533,6 +2538,68 @@ class RPC extends RPCBase {
...
+        break;
+
+      const {value} = iter;
+      const ns = NameState.decode(value);
+
+      if (ns.data.length <= 0)
+        continue;
+
+      const fqdn = bns.util.fqdn(ns.name.toString('ascii'));
+      const resource = Resource.decode(ns.data);
```
rithvikvibhu, line 2581:
```suggestion
      try {
        const resource = Resource.decode(ns.data);
      } catch (e) {
        continue;
      }
```

james-stevens:
TBH: I think `buffrr`'s AXFR plug-in is a better solution than this, so (personally) I'd just drop this

https://github.com/buffrr/hsd-axfr

nglabs42 (commented):
Merged with local pull of master, with rithvikvibhu's update at line 2581 of lib/node/rpc.js:
let resource;
try {
resource = Resource.decode(ns.data);
if (!resource) continue;
} catch (e) {
continue;
}
I was able to pull the zonefile, 16 GB.

nglabs42 (approved):

nglabs42 (approved):
tested RPC DUMPZONE

nglabs42 (changes requested):
Update lib/node/rpc.js 
Line 2581 'const resource = Resource.decode(ns.data);'
to

let resource;
try {
resource = Resource.decode(ns.data);
if (!resource) continue;
} catch (e) {
continue;
}

PR #537, opened 2021-01-18
test: isolate mempool event listeners in wallet-http-test

pinheadmz:
Also removed `assert.ok(false)` in these handlers since those
assertion errors would be swallowed by the mempool event
emitter and would not be caught by the test suite anyway.

These test modifications can be tested themselves by reviewers
by setting `broadcast: true` inside `wallet.createOpen(...)`

ping @TLD-O I think this will help you add tests in #529

### DIFF
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -252,8 +252,8 @@ describe('Wallet HTTP', function() {
   });
 
   it('should create an open and broadcast the tx', async () => {
-    let entered = false;
-    const handler = () => entered = true;
+    let emitted = 0;
+    const handler = () => emitted++;
     node.mempool.on('tx', handler);
 
     const json = await wallet.createOpen({
@@ -263,13 +263,14 @@ describe('Wallet HTTP', function() {
     // wait for tx event on mempool
     await common.event(node.mempool, 'tx');
 
-    assert.equal(entered, true);
     const mempool = await nclient.getMempool();
 
     assert.ok(mempool.includes(json.hash));
 
     const opens = json.outputs.filter(output => output.covenant.type === types.OPEN);
     assert.equal(opens.length, 1);
+
+    assert.equal(emitted, 1);
 
     // reset for next test
     node.mempool.removeListener('tx', handler);
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -252,12 +252,13 @@ describe('Wallet HTTP', function() {
   });
 
   it('should create an open and broadcast the tx', async () => {
+    let entered = false;
+    const handler = () => entered = true;
+    node.mempool.on('tx', handler);
+
     const json = await wallet.createOpen({
       name: name
     });
-
-    let entered = false;
-    node.mempool.on('tx', () => entered = true);
 
     // wait for tx event on mempool
     await common.event(node.mempool, 'tx');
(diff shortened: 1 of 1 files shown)


test/wallet-http-test.js
```diff
@@ -305,21 +307,22 @@ describe('Wallet HTTP', function() {
...
     assert.equal(opens.length, 1);
+
+    // reset for next test
+    node.mempool.removeListener('tx', handler);
   });
 
   it('should create an open and not sign the transaction', async () => {
+    let entered = false;
+    const handler = () => entered = true;
+    node.mempool.on('tx', handler);
```
tynes, line 318: It could be possible to use the event handler `once` instead of `on` which would would remove the need to call `removeListener` below. A benefit of using `on` over `once` would be testing that the handler is called exactly once

pinheadmz: True - but the point of the test is the event never fires!

test/wallet-http-test.js
```diff
@@ -269,20 +270,21 @@ describe('Wallet HTTP', function() {
...
+
     const json = await wallet.createOpen({
       name: name,
       broadcast: false
     });
 
-    let entered = false;
-    node.mempool.on('tx', () => {
```
tynes, line 281: By creating the event listener after sending the transaction, it is subject to a race condition where it is possible that the event is never triggered. Good call in moving it before.

test/wallet-http-test.js
tynes: `assert.equal(entered, false);`

So no need for `once` and the event listener has to be removed no matter what

test/wallet-http-test.js
pinheadmz: Yup thanks.

test/wallet-http-test.js
pinheadmz: Thats what I'm thinking! ...?

test/wallet-http-test.js
```diff
@@ -269,20 +270,21 @@ describe('Wallet HTTP', function() {
...
-      entered = true;
-      assert.ok(false);
-    });
-
     await sleep(500);
```
tynes, line 288: It would be ideal to remove this sleep as well, but I'm not sure of a good way to do so

test/wallet-http-test.js
```diff
@@ -252,13 +252,14 @@ describe('Wallet HTTP', function() {
   });
 
   it('should create an open and broadcast the tx', async () => {
+    let entered = false;
+    const handler = () => entered = true;
```
tynes, line 256: This comment applies here:
https://github.com/handshake-org/hsd/pull/537/files#r559782100

tynes: I think its worth adding test coverage here that it is only called once to prevent applications that rely on the websocket from getting the same notification twice. I don't think that is happening but I don't think any sort of test exists yet that covers that particular case

pinheadmz: Ok interesting idea, so instead of using `once` maybe we'd replace `entered` with a counter and at the end of the test, assert the counter was at `1` ?

tynes: > Ok interesting idea, so instead of using once maybe we'd replace entered with a counter and at the end of the test, assert the counter was at 1 ?

Yup that would work great. Will approve this PR after that change

pinheadmz: added in c36f27026c3dd6322904d4ae8ba20adbe02417a3

tynes (approved):

PR #538, opened 2021-01-18
v2.3.0

pinheadmz:
This release contains a bug fix that affected wallet balances after sending or receiving names. The bug could permanently corrupt the walletDB. It also introduces an API endpoint to clear out this corruption so the wallet can be restored with a rescan (after upgrading). There are also policy improvements to help clear out or confirm transactions accidentally sent with too-low a fee.

Reviewers can install master branch of hsd at this time, and check through the list of closed PRs in the v2.3.0 milestone list: https://github.com/handshake-org/hsd/milestone/2?closed=1

You might grep your log for similar messages to see 3-day-old transactions being evicted from the mempool:

```
[D:2021-01-14T15:46:44Z] (mempool) Added 6cb49aa29103f4f74ab9a9d37e776a827eec26957544eb69f15e13e04db91821 to mempool (txs=46).
...
[D:2021-01-17T15:46:44Z] (mempool) Removing package 6cb49aa29103f4f74ab9a9d37e776a827eec26957544eb69f15e13e04db91821 from mempool (too old).
```

And if you ever transferred names in or out of your wallet you can try the `deepclean` process to restore an accurate balance in the UI (`hsw-cli balance`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # HSD Release Notes & Changelog
 
-## unreleased
-
+## v2.3.0
 
 ### Node changes
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ $ npm install -g https://github.com/handshake-org/hsd.git
 
 A `git` ref can be used to install a particular version by appending
 a `#` and the name of the `git` ref to the URL. For example,
-`https://github.com/handshake-org/hsd.git#v2.2.0`. It is recommended
+`https://github.com/handshake-org/hsd.git#v2.3.0`. It is recommended
 to use the [latest tagged release](https://github.com/handshake-org/hsd/releases).
 
 If adding `hsd` as a dependency to a project, use the command:
--- a/lib/pkg.js
+++ b/lib/pkg.js
@@ -70,7 +70,7 @@ pkg.cfg = `${pkg.core}.conf`;
  * @default
  */
 
-pkg.version = '2.2.0';
+pkg.version = '2.3.0';
 
 /**
  * Repository URL.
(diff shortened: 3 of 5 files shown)


tynes:
The updates to the following files all look good to me:
- `package.json`
- `package-lock.json`
- `lib/pkg.js`
- `CHANGELOG.md`

It may be worth updating the `README.md` so that it doesn't need to be changed for each release, something like: 

```
For example,	a `#` and the name of the `git` ref to the URL. For example,
https://github.com/handshake-org/hsd.git#<tag> where <tag> is a tagged
release of hsd.
```

It may also be worth documenting the release process someplace so that a checklist can be followed, helping to prevent a release from ever forgetting anything. This could go in the `README` or in a `docs` directory.

Will you be able to release this on `npm`?

pinheadmz:
@tynes thanks for asking, its important to note that only @chjj can push updates to npm. That means that although we can bump versions on github, we can not bump versions on npm without his permission. I'm not in love with npm anyway and I think it's probably best if we eventually migrate away from npm [as bcoin has](https://github.com/bcoin-org/bcoin/issues/905). We will have to instruct new users to clone from github and not npm for the time being. Most important is that the new bug fixes and corruption-fixes get merged into Bob ASAP before more users experience the bug.

Relay nodes and miners hopefully will upgrade via github once we announce to get the mempool policy rules deployed. My experience chatting with project leaders on telegram is that they usually clone from github anyway (not npm install).

Regarding the README note and future release processes: I have been thinking about this lately and after this bug fix release I want to adopt a policy like [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/master/doc/release-process.md). Specifically that master branch will always have a `v.X.Y.Z.99` version to indicate it is unstable master branch, and as we get close to release, a new branch will created with the updated version number so we can feature-freeze and finalize that release while ongoing work continues merging to master branch.

Issue #541, opened 2021-01-22
DNS server should be more flexible to plugins


pinheadmz:
If we want to add an ENS resolver for example like https://github.com/tynes/hsd-ens-resolution we can make a plugin much easier.

The blacklist object in `lib/dns/server.js` can be made a property of the class so a plugin can modify it.
Also the lookup function could be modified to check plugins first:

https://github.com/handshake-org/hsd/blob/1a086e44b5b60e1ebb01955b0e1b2a0ebd00b42d/lib/dns/server.js#L323-L326

Issue #455, opened 2020-05-17
Handshake TLDs .0 and .1 ... .255 confusion with IPv4 Namespace


dnsguru:
(Migrated from [hs-names#9](https://github.com/handshake-org/hs-names/issues/9) as advised by @pinheadmz)

It is problematic to allow resolution for TLDs that are entirely numeric, as IPv4 addresses and DNS use the period as delimiters and this will confuse applications.

This refers to any digit between “0” and “9” (Unicode code points U+0030 to U+0039) 

Specifically problematic are the .0 and then .1 through .255, but purely numeric TLDs of any size are also problematic when used as the entire host component of a URL, as many BSD networking stacks still convert Quad A record IP addresses from pure numeric entries.  

Example:   HNS TLD = .8, one could give a hostname 8.8.8 and put it in the .8 TLD, which would make that host 8.8.8.8 in the DNS, which could cause a lot of mayhem within HNS resolvers that might fall back to DNS.Google for lookups.

Example:  HNS TLD = .1, one could establish a host 127.0.0 and put it in the .1 TLD, which would confuse certain software attempting to connect to 127.0.0.1, which is typically "localhost"

Other examples in the .0 and .1-.255 range are hopefully self-explanatory

The example of a purely numeric string outside of that range is largely with the use of the TLD itself as a resolution point...

See [This Article which articulates the dotted quad binary to integer computation](https://security.stackexchange.com/questions/108616/using-a-single-number-instead-of-an-ip-address-or-a-dns-name-which-are-concrete)

chjj:
Hi @dnsguru,

I think this is something we discussed in the past. Typically, most (if not all) applications do not query DNS for a hostname which deserializes to an IP address. I'm not aware of any web browser that does this for example. I've never seen an application query my local resolver for an IP address (ever). In my opinion, an application doing DNS queries on IP addresses is probably already broken in other ways and should not be used.

That said, accidents do happen. I agree there's still a fair case to be made that an application could mistakenly query an IP address and end up with disastrous results. Like @pinheadmz mentioned, I think an acceptable solution would be for resolvers to simply return NXDOMAIN for any [0-255] TLD and not even attempt to resolve via the HNS network.

I think we should maybe start to document edge cases like this to create guidelines for HNS resolver implementation. In the more immediate term we can implement this fix in hsd. Leaving this issue open.

pinheadmz:
[...]

Issue #208, opened 2019-06-28
On Resolving Alternative DNS Systems


tynes:
There are certain Top Level Domains that are blacklisted by policy. In Bitcoin, policy rules are generally more strict than consensus rules and can be changed by users at runtime. They allow for community consensus to change rules over time based on how the network is evolving. See the [bitcoind](https://github.com/bitcoin/bitcoin/blob/c536dfbcb00fb15963bf5d507b7017c241718bf6/src/policy/policy.cpp) policy file for an example.

For the Top Level Domains that are blacklisted by policy, it is possible to resolve these TLDs if specific client code was written and the community agreed on integrations. It would allow a user to be able to run `hsd` as their DNS server and be able to resolve additional custom Top Level Domains that already exist that are not part of the traditional ICANN system. 

The Top Level Domains that are currently blacklisted by policy are:

```js
const blacklist = new Set([
  'bit', // Namecoin
  'eth', // ENS
  'exit', // Tor
  'gnu', // GNUnet (GNS)
  'i2p', // Invisible Internet Project
  'onion', // Tor
  'tor', // OnioNS
  'zkey' // GNS
]);
```

https://github.com/handshake-org/hsd/blob/master/lib/dns/server.js#L51-L60

The code was written in a way where clients for these TLDs can easily be integrated into the codebase. A client for any of the TLDs in the blacklist could be added here and used instead of doing a lookup in the Urkel Tree.

```js
const data = !blacklist.has(tld)
  ? (await this.lookupName(tld))
  : null;
```
---

Are there any additional TLDs that should be blacklisted by policy?  There is Blockstack's [BNS](https://docs.blockstack.org/core/naming/introduction.html) but I am unclear if they only allow certain TLDs or if they are a complete alternative naming system. A naming service built on Zilliqa recently launched called [ZNS](https://github.com/unstoppabledomains/zns) that uses the `.zil` Top Level Domain and another blockchain project called [Emercoin](https://emercoin.com/en/documentation/blockchain-services/emerdns/emerdns-introduction) supports the TLDs `.coin`, `.emc`, `.lib` and `.bazar`. I do not know how much these services are used in practice.

[...]

PR #544, opened 2021-01-27
dns/server: blacklist rfc2606 domains in the resolver

turbomaze:
Reference: https://tools.ietf.org/id/draft-chapin-rfc2606bis-00.html#registry

There are a small handful of domains that are commonly used for internal DNS that ICANN has never delegated. They make up a significant portion of DNS traffic because all those internal DNS queries leak to resolvers they're not supposed to. Handshake has already blacklisted a few of these domains from being registered (see `covenants/rules.js`), but there are others that have not been blacklisted.

Perhaps this was an intentional choice, and if so, I'm curious to discuss it. But assuming it was just an oversight, this PR blacklists the remaining names in the referenced RFC draft.

### DIFF
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -51,6 +51,20 @@ const TYPE_MAP = Buffer.from('000722000000000380', 'hex');
 const RES_OPT = { inet6: false, tcp: true };
 const CACHE_TTL = 30 * 60 * 1000;
 
+// reference: https://tools.ietf.org/id/draft-chapin-rfc2606bis-00.html#registry
+const rfc2606 = [
+  'corp',
+  'domain',
+  'example',
+  'home',
+  'host',
+  'invalid',
+  'lan',
+  'local',
+  'localdomain',
+  'localhost',
+  'test'
+];
 const blacklist = new Set([
   'bit', // Namecoin
   'eth', // ENS
@@ -59,7 +73,8 @@ const blacklist = new Set([
   'i2p', // Invisible Internet Project
   'onion', // Tor
   'tor', // OnioNS
-  'zkey' // GNS
+  'zkey', // GNS
+  ...rfc2606
 ]);
 
 /**


pinheadmz:
some of these strings are already blacklisted at the consensus level:

https://github.com/handshake-org/hsd/blob/d304ff4c6728fe1932920b89dc6ee20e0dc06618/lib/covenants/rules.js#L85-L91

The strings in the dns server can actually still be acquired on chain (I think they are all reserved though) whereas the names in the rules.js blacklist are invalid and can not be opened for auction or claimed, ever.

turbomaze:
Yep definitely can't touch the `consensus/rules.js` file. I think that the ideal version of this is an additional DNS blacklist that gets enabled by switching on a config flag. This way people that run resolvers don't need to get their logs flooded with all the constant errors caused by requests to these names (the linked draft estimates that it's a considerable % of DNS traffic).

pinheadmz:
Out of curiosity I wanted to determine who and how much this change would impact Handshake users. Of the names from the RFC, half are already blacklisted. Two have been sold, three are reserved and one... fantastically... has been assigned by ICANN despite being on this list! Maybe @dnsguru has some insight to that.


| name | status |
|-|-|
|`example`| blacklisted in rules.js |
|`invalid`| blacklisted in rules.js |
|`local`| blacklisted in rules.js |
|`localhost`| blacklisted in rules.js |
|`test`| blacklisted in rules.js |
|`corp`| sold for 10,000 HNS |
|`localdomain`| sold for 111 HNS, has HNS root zone records but they go nowhere |
|`domain`| reserved for `domain.com.au.` |
|`home`| reserved for `home.pl.` |
|`lan`| reserved for `lan.com.` |
|`host`| [reserved for ICANN TLD `host.`](https://www.iana.org/domains/root/db/host.html) |

brandondees:
I think I might have captured localdomain, can't verify atm, but I've been making a little effort to grab these kinds of things when possible to prevent them from becoming vehicles for abuse and to be able to e.g. comment here and say "I am the affected party, block them" or "I have a sane use case to justify unblocking" (I don't but it was a thought)

dnsguru:
> Out of curiosity I wanted to determine who and how much this change would impact Handshake users. Of the names from the RFC, half are already blacklisted. Two have been sold, three are reserved and one... fantastically... has been assigned by ICANN despite being on this list! Maybe @dnsguru has some insight to that.

RFC2606 from the year 1999 was replaced by [RFC6761](https://datatracker.ietf.org/doc/rfc6761) in the year 2013, so if this project is going to work with RFCs on names, it is important to use the most recent versions of them.

As TLDs that had been approved in the 2012 round were starting to be added to the root nameservers, ICANN (intended to mean the whole of the ICANN stakeholder community  + ICANN staff) paid close attention to the potential for things that might break stuff.

One of the areas that had some attention was called '[Name Collision](https://icann.org/namecollision)'.  In plain terms this was careful measurement, assessment and planning for where existing devices and networks might have some namespaces being used before a TLD that matched it gets lit up on the root. 

There was significant work done by ICANN.org and the ICANN community on Name Collision, and they assembled a humbling pantheon of ivory tower-worthy experts from accross a diversity of perspectives, including security professionals and firms.

Without relitigating that stuff, they took some random sources of historic root zone traffic logs from a few of the root name servers to see what kinds of TLD traffic was hitting those nameservers and getting an NXD.

.host was found to fall below a threshold

pinheadmz:
@dnsguru thanks - so looking at RFC6761, the only names I see specified are:
`test`, `localhost`, `invalid`, and `example` -- which are already on the consensus blacklist (along with a bonus `local`, perhaps from a previous draft?)

Jothan what is your opinion on this pull request?

dnsguru:
This should consider #455 #208 and #255 from this repo and would want to also wrap in
https://github.com/handshake-org/hs-names/issues/6 & https://github.com/handshake-org/hs-names/issues/8

turbomaze:
> This should consider #455 #208 and #255 from this repo and would want to also wrap in
> [handshake-org/hs-names#6](https://github.com/handshake-org/hs-names/issues/6) & [handshake-org/hs-names#8](https://github.com/handshake-org/hs-names/issues/8)

Thanks for those thoughts above + all these references to relevant issues! I think a broader refactor of the dns server is warranted that adds plugin functionality + more fine grained control of the DNS blacklist. I don't have capacity for that at the moment but maybe soon

pinheadmz:
I think this goes in a larger category earmarked in https://github.com/handshake-org/hsd/issues/541
There's all kinds of resolver settings that can be set with a plugin, if we just make `server.js` more flexible. All the issues @dnsguru mentioned plus some other greatest hits like "resolve ICANN root before HNS root". Then we can issue a standard child-proof plugin that hackers can easily disable. I'd feel better about that than just expanding the blacklist for everyone using hsd.

dnsguru:
Thanks for the opportunity to provide input.

Wrt ICANN or handshake first...

I was careful not to identify a bias as to which system should be first
attempted (ICANN-handshake vs Handshake-ICANN) as there is a growing number
of alternatives to consider.

The most decentrallized and democratic way to handle that imho is to expose
the order of operation to the user for them to configure.

brandondees:
I concur with the consensus here that plugins/configurability are gonna be necessary longer term and this kind of thing might as well be deferred until it can be resolved with that kind of approach. We definitely have a spectrum of users with different preferences in terms of prioritizing HNS or ICANN or other potentially conflicting systems like ENS, Butterfly, etc. or using greylisting for individual names, and those options will only multiply.

pinheadmz:
Please see https://github.com/pinheadmz/holdmyhand which requires update to hsd (review requested!) https://github.com/handshake-org/hsd/pull/558

pinheadmz:
I'm going to close this now because, as the mathematicians like to say "the solution exists". #558 is merged, and the `holdmyhand` plugin has been released. Users or services interested in applying extra protection can do so in the applicaiton layer (i.e. with a plugin) and so I don't think this is a relevant issue to hsd anymore.

PR #546, opened 2021-01-28
Validates documentation on push

Falci:
This will trigger the jsdoc after each push/merge to master. 
The documentation will be generated and discarded, but this guarantees the docs are valid.


package.json
```diff
@@ -49,7 +49,8 @@
     "urkel": "~0.6.3"
   },
   "devDependencies": {
-    "bmocha": "^2.1.5"
+    "bmocha": "^2.1.5",
+    "jsdoc": "^3.6.6"
```
pinheadmz, line 53: I'm not opposed to including this check in CI, but can we do it without adding a dependency? 

Notice how github actions has to install bslint on its own when it installs hsd. Contributors could be advised to install jsdocs in https://github.com/handshake-org/hsd/blob/master/CONTRIBUTING.md#code-style the same way they are advised to install eslint.

Falci:
IMO, the project should include it since it's literally a "devDependency". 
Since the project already miss the other dependencies (lint and coverage) I'm removing jsdoc to keep it consistent.

pinheadmz:
Looks good, checked a bad-docs PR against my own master branch: https://github.com/pinheadmz/hsd/pull/6

We can merge, but @Falci can you clean up the commits if possible? Even just squash to one commit would work.

pinheadmz:
merged in https://github.com/handshake-org/hsd/commit/9c7bc0a85ced9d3f29fb1a2e961dfcc70c567025

Issue #199, opened 2019-06-17
Add SECURITY.md


tynes:
Adding a `SECURITY.md` file to the root of the repo is becoming increasingly common and I think that it would only help `hsd`.

There is an open pull request to add this to `bcoin` here: https://github.com/bcoin-org/bcoin/pull/799

See `bitcoind` - https://github.com/bitcoin/bitcoin/blob/master/SECURITY.md
See `geth` - https://github.com/ethereum/go-ethereum/blob/master/SECURITY.md

PR #547, opened 2021-01-28
add SECURITY.md

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/199

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,22 @@
+# Security Policy
+
+## Supported Versions
+
+Please see [Releases](https://github.com/handshake-org/hsd/releases).
+We recommend using the [most recently released version](https://github.com/handshake-org/hsd/releases/latest).
+
+## Reporting a Vulnerability
+
+To report security issues send an email to a project maintainer below (not for support).
+
+The following keys may be used to communicate sensitive information to developers:
+
+| Name | Email | Fingerprint | Full Key |
+|------|-------|-------------|----------|
+| Christopher Jeffrey (@chjj) | chjjeffrey@gmail.com | B4B1F62DBAC084E333F3A04A8962AB9DE6666BBD | https://keybase.io/chjj |
+| Matthew Zipkin (@pinheadmz) | pinheadmz@gmail.com  | E61773CD6E01040E2F1BD78CE7E2984B6289C93A | https://keybase.io/pinheadmz |
+
+You can also import a key by running the following command with an individual’s fingerprint:
+
+`gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys "<fingerprint>"`
+


SECURITY.md
```diff
@@ -0,0 +1,20 @@
...
+To report security issues send an email to chjjeffrey@gmail.com (not for support).
+
+The following keys may be used to communicate sensitive information to developers:
+
+| Name | Fingerprint | Full Key |
+|------|-------------|--------|
+| Christopher Jeffrey (@chjj) | B4B1F62DBAC084E333F3A04A8962AB9DE6666BBD | https://keybase.io/chjj |
+
+
+You can also import a key by running the following command with an individual’s fingerprint: `gpg --recv-keys "<fingerprint>"`
```
tynes, line 19: When I run this command, it does not work:

```
$ gpg --recv-keys B4B1F62DBAC084E333F3A04A8962AB9DE6666BBD
```

Note that I do not have any special configuration with the `gpg`, its a rather fresh install. I'm not sure what key servers it is on besides keybase, maybe the command to import from keybase is included as well as the exact `gpg` command needed to import. I believe there is a flag to specify the keyserver to use.

SECURITY.md
```diff
@@ -0,0 +1,20 @@
...
+## Reporting a Vulnerability
+
+To report security issues send an email to chjjeffrey@gmail.com (not for support).
+
+The following keys may be used to communicate sensitive information to developers:
+
+| Name | Fingerprint | Full Key |
+|------|-------------|--------|
+| Christopher Jeffrey (@chjj) | B4B1F62DBAC084E333F3A04A8962AB9DE6666BBD | https://keybase.io/chjj |
+
```
tynes, line 17: I think @pinheadmz should be on this list as well as a fallback in case there is no response from @chjj. This would help to diffuse responsibility so that chjj can take time off without worrying about things if he decides to.

tynes (changes requested):
Besides the two comments, I think another possibility would be to include the pubkeys themselves directly in the `SECURITY.md`. This means that there is a chance they could go stale as it would be another place to have to remember to update them. Having them in git is kind of nice though

SECURITY.md
rithvikvibhu: While my gpg could import it as it is (think it's modified config), searching for it on https://pgp.mit.edu/, http://keys.gnupg.net/, https://keyserver.ubuntu.com/ all returned Not Found until `0x` was prefixed to it. Maybe this hex prefix could be added if someone searches the key id on these sites.

pinheadmz: Does this command work for you both?

```
$ gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys B4B1F62DBAC084E333F3A04A8962AB9DE6666BBD
gpg: key 8962AB9DE6666BBD: public key "Christopher Jeffrey <chjjeffrey@gmail.com>" imported
gpg: no ultimately trusted keys found
gpg: Total number processed: 1
gpg:               imported: 1
```

While we're at we can check mine as well:

```
$ gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys E61773CD6E01040E2F1BD78CE7E2984B6289C93A
gpg: key E7E2984B6289C93A: public key "Matthew Zipkin (GitHub Signing Key) <pinheadmz@gmail.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

rithvikvibhu: Both work :+1:.

SECURITY.md
pinheadmz: added and rebased.

pinheadmz:
Added my own email and key, and specified a keyserver in the example command where both my and JJs key can be downloaded. I think that and keybase are sufficient sources. Both of those systems are well-known PKI and users may have social "connections" or trusted peers in common, etc which I think is a bit more pro than hosting the keys inside the git repo. Keybase in particular also verifies twitter, reddit and github so I think thats a good way to serve the keys.

Issue #528, opened 2021-01-04
Full and SPV nodes should have a --no-dns option


pinheadmz:
When set at launch, `--no-dns=true` or `{noDns: true}` should prevent the node from constructing with Authoritative and Recursive name servers. I mentioned this in a [Bob wallet issue](https://github.com/kyokan/bob-wallet/issues/140#issuecomment-630318788) before but it will be useful:

- Avoid port collisions with hnsd or other HNS resolver
- Simplify integration tests with Full and SPV nodes
- reduce resource usage when a user just needs a wallet / fullnode
- possibly reduce attack surface by closing ports

PR #554, opened 2021-02-05
node: accept --no-dns configuration option to disable NS and RS

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/528

Summary (also included in CHANGELOG):

- `FullNode` and `SPVNode` accept configuration parameter `--no-dns` (or `no-dns: true` in
`hsd.conf`) which launches the node without either DNS server (the root authoritative
server and the recursive resolver). This avoids some port collisions with other HNS resolvers
like hnsd running locally, and generally separates and reduces security concerns around
running unneeded servers when a node is just used for transactions and blocks.
`--no-rs` is also accepted to disable the recursive DNS resolver (but keep the root server).

The default is `false` (meaning by default, hsd runs with the DNS servers).

### DIFF
--- a/lib/node/fullnode.js
+++ b/lib/node/fullnode.js
@@ -147,24 +147,28 @@ class FullNode extends Node {
       cors: this.config.bool('cors')
     });
 
-    this.ns = new RootServer({
-      logger: this.logger,
-      key: this.identityKey,
-      host: this.config.str('ns-host'),
-      port: this.config.uint('ns-port', this.network.nsPort),
-      lookup: key => this.chain.db.tree.get(key),
-      publicHost: this.config.str('public-host')
-    });
+    if (!this.config.bool('no-dns')) {
+      this.ns = new RootServer({
+        logger: this.logger,
+        key: this.identityKey,
+        host: this.config.str('ns-host'),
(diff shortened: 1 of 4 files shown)


pinheadmz:
After discussion with @buffrr I think a better choice is keeping `--no-dns` but also adding `--no-rs` because in some configurations it may desirable to just run hsd as a root name server and use something externally (like BIND) for the recursive.

lib/node/fullnode.js
```diff
@@ -147,24 +151,28 @@ class FullNode extends Node {
...
-    this.rs = new RecursiveServer({
-      logger: this.logger,
-      key: this.identityKey,
-      host: this.config.str('rs-host'),
-      port: this.config.uint('rs-port', this.network.rsPort),
-      stubHost: this.ns.host,
-      stubPort: this.ns.port,
-      noUnbound: this.config.bool('rs-no-unbound')
-    });
+    if (!this.noDNS) {
```
nodar-chkuaselidze, line 154: nit: because this.noDNS and noRS are not used outside you could inline this.config thing here)

nodar-chkuaselidze (approved):
LGTM

nodar-chkuaselidze (approved):

PR #548, opened 2021-01-28
namestate: expose transfer lockup period

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/178

example output:

```
$ hsd-rpc getnameinfo test-ds
{
  "start": {
    "reserved": false,
    "week": 28,
    "start": 56
  },
  "info": {
    "name": "test-ds",
    "nameHash": "c3243b05628fe67bde09bf7bf53752cd446fab5e0434d15c9d04c5db8dcb7ce4",
    "state": "CLOSED",
    "height": 211,
    "renewal": 235,
    "owner": {
      "hash": "5c8beb5651b28c05c10133771933e42fea09130c983ebd27715b414ba30db757",
      "index": 0
    },
    "value": 0,
    "highest": 0,
    "data": "0000e00b08022095a57c3bab7849dbcddf7c72ada71a88146b141110318ca5be672057e865c3e2",
    "transfer": 245,
    "revoked": 0,
    "claimed": 0,
    "renewals": 0,
    "registered": true,
    "expired": false,
    "weak": false,
    "stats": {
      "renewalPeriodStart": 235,
      "renewalPeriodEnd": 5235,
      "blocksUntilExpire": 4990,
      "daysUntilExpire": 34.65,
      "transferLockupStart": 245,
      "transferLockupEnd": 255,
      "blocksUntilValidFinalize": 10,
      "hoursUntilValidFinalize": 1.67
    }
  }
}
```

PR #555, opened 2021-02-09
namestate: improve "stats" object for transfers, expiry and revoke

pinheadmz:
Adds an expired state to the "stats" object, refactors the `toStats()` method in namestate.js.

```
    "stats": {
      "blocksSinceExpired": 21
    }
```

### DIFF
--- a/lib/covenants/namestate.js
+++ b/lib/covenants/namestate.js
@@ -733,7 +733,8 @@ class NameState extends bio.Struct {
       revealPeriod,
       renewalWindow,
       auctionMaturity,
-      transferLockup
+      transferLockup,
+      claimPeriod
     } = network.names;
 
     const openPeriod = treeInterval + 1;
@@ -808,7 +809,9 @@ class NameState extends bio.Struct {
       }
       case states.CLOSED: {
         const start = this.renewal;
-        const end = start + renewalWindow;
+        const end = this.claimed ?
+          claimPeriod :
+          start + renewalWindow;
         const blocks = end - height;
         const days = ((blocks * spacing) / 60 / 60 / 24);
 
@@ -816,6 +819,7 @@ class NameState extends bio.Struct {
         stats.renewalPeriodEnd = end;
 
         stats.blocksUntilExpire = blocks;
+        assert(stats.blocksUntilExpire >= 0);
         stats.daysUntilExpire = Number(days.toFixed(2));
 
         // Add these details if name is in mid-transfer
--- a/lib/covenants/namestate.js
+++ b/lib/covenants/namestate.js
@@ -209,7 +209,7 @@ class NameState extends bio.Struct {
     if (this.isClaimable(height, network))
       return false;
 
-    // If we haven't been renewed in a year, start over.
+    // If we haven't been renewed in two years, start over.
     if (height >= this.renewal + network.names.renewalWindow)
       return true;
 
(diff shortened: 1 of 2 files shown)


lib/covenants/namestate.js
```diff
@@ -740,6 +741,14 @@ class NameState extends bio.Struct {
 
     const stats = {};
 
+    if (this.isExpired(height, network)) {
```
nodech, line 744: I don't FULLY understand all the state transitions but because this `isExpired` is preceding everything I was wondering if all checks do make sense from isExpired.

isExpired has three `true` returns, one of them is:
   1. When name is revoked and then expires (revoked + auction maturity) - Don't we want to include isRevoked information here as well ? (to provide stats whether it was revoked -> expired vs someone forgot to renew?)
   2. If we have not renewed in two years - I guess this PR is concentrated on that case.
   3. Auction failed - no one revealed their bids.
-- Results?
  1. I guess `expired` would be `this.renewal + network.names.auctionMaturity` or `this.revoked + network.names.auctionMaturity` (My understanding, after looking at chain code, is `this.renewal` is always updated when something changes about a name)
  2. I believe this is the case, when this `expire` value makes sense, it just expired w/o renewal.
  3. In this case, I am not sure if we even call this expire. `expired` would probably be the `this.renewal` itself? (last name update when reveal failed?)

lib/covenants/namestate.js
```diff
@@ -209,7 +209,7 @@ class NameState extends bio.Struct {
     if (this.isClaimable(height, network))
       return false;
 
-    // If we haven't been renewed in a year, start over.
+    // If we haven't been renewed in two years, start over.
     if (height >= this.renewal + network.names.renewalWindow)
```
nodech, line 213: Was initial plan to have it expire in a year ?

pinheadmz: haha yeah I guess. There's a bunch of outdated docs and comments we're still finding since mainnet:

https://github.com/handshake-org/hsd/pull/556/files
https://github.com/handshake-org/handshake-web/pull/7
https://github.com/handshake-org/handshake-web/pull/4
https://github.com/handshake-org/handshake-web/pull/4

lib/covenants/namestate.js
pinheadmz: Ok I think you're right I need to add a test for `revoked` and make sure this still makes sense. The revoked period is `auctionMaturity` after which the name is just `CLOSED` and can be reopen for a brand new auction. The point of this PR is that after a name is expired (which is `this.renewal + network.names.renewalWindow`) the JSON stats will indicate a `CLOSED` state but with negative numbers for the renewal, which is invalid after expiration:

```
  stats: {
    renewalPeriodStart: 72,
    renewalPeriodEnd: 5072,
    blocksUntilExpire: -4928,
    daysUntilExpire: -34.22
  }
```

...and makes even less sense if the name was in mid-transfer (this was broken recently in https://github.com/handshake-org/hsd/pull/548)

```
  stats: {
    renewalPeriodStart: 72,
    renewalPeriodEnd: 5072,
    blocksUntilExpire: 1,
    daysUntilExpire: 0.01,
    transferLockupStart: 98,
    transferLockupEnd: 108,
    blocksUntilValidFinalize: -4963,
    hoursUntilValidFinalize: -827.17
  }
```

Note: `this.renewal` is not always updated when something changes about a name. You can see in `chain.verifyCovenants()` that `ns.setRenewal()` is only called on certain covenants: `CLAIM`, `REGISTER`, `RENEW` and `FINALIZE`

nodech: What about failed auction (Even though it should not happen that often) ? - Case when no one revealed.
OPEN -> BID(s) and no one reveals after that. (I believe that's the last case I mentioned)

nodech: You could just use custom version of isExpired - `if (isExpired(..) && !isRevoked(..) && !this.owner.isNull())` to target this specific case. (Or create another helper method)

pinheadmz: An auction that ends with no reveals goes back to null state:

```
{
  "start": {
    "reserved": false,
    "week": 0,
    "start": 0
  },
  "info": null
}
```

pinheadmz:
Added ed85a06aa14290349923c272a01bf72a7900ac4e which checks for REVOKE first before expire. The `stats` field now progresses like below. I decided to keep some of the negative number outputs since I don't want to break the API and there might be applications that check for negative numbers to know when its "ok to finalize", etc.


before transfer:
```
{
  renewalPeriodStart: 72,
  renewalPeriodEnd: 5072,
  blocksUntilExpire: 4975,
  daysUntilExpire: 34.55
}
```

after transfer:
```
{
  renewalPeriodStart: 72,
  renewalPeriodEnd: 5072,
  blocksUntilExpire: 4974,
  daysUntilExpire: 34.54,
  transferLockupStart: 98,
  transferLockupEnd: 108,
  blocksUntilValidFinalize: 10,
  hoursUntilValidFinalize: 1.67
}
```

before expire:
```
{
  renewalPeriodStart: 72,
  renewalPeriodEnd: 5072,
  blocksUntilExpire: 1,
  daysUntilExpire: 0.01,
  transferLockupStart: 98,
  transferLockupEnd: 108,
  blocksUntilValidFinalize: -4963,
  hoursUntilValidFinalize: -827.17
}
```

after expire:
```
{ blocksSinceExpired: 0 }
```


after revoke before reopen:
```
{
  revokePeriodStart: 109,
  revokePeriodEnd: 174,
  blocksUntilReopen: 65,
  hoursUntilReopen: 10.83
}
```

after revoke after reopen:
```
{
  revokePeriodStart: 109,
  revokePeriodEnd: 174,
  blocksUntilReopen: 0,
  hoursUntilReopen: 0
}
```

after revoke after what would be expire:
```
{
  revokePeriodStart: 109,
  revokePeriodEnd: 174,
  blocksUntilReopen: -4935,
  hoursUntilReopen: -822.5
}
```

pinheadmz:
@nodech I did a big refactor on this PR and added a comprehensive test instead of trying to shoe-horn in a few tests here and there in the middle of an auction. So I think all the `"stats"` stuff is right now, for all possible states.

Ironically, the original purpose of this PR was to set `ns.expired` (not the stats object). I'm going to add a commit for that next.

lib/covenants/namestate.js
```diff
@@ -740,96 +740,117 @@ class NameState extends bio.Struct {
...
+
+        stats.revokePeriodStart = start;
+        stats.revokePeriodEnd = end;
+
+        stats.blocksUntilReopen = blocks;
+        stats.hoursUntilReopen = Number(hours.toFixed(2));
+        break;
+      }
+      case EXPIRED: {
+        const expired = this.renewal + network.names.renewalWindow;
```
pbkhrv, line 850: renewalWindow is already imported and being used in line 811, perhaps don't need to reference `network.names.` here, otherwise the question is "are these constants different somehow?"

pinheadmz: You're absolutely right, thanks - actually its defined way up in L734, I'll push an update soon

pinheadmz: thank you! fixed and rebased

test/namestate-test.js
```diff
@@ -0,0 +1,341 @@
...
+  claimPeriod,
+  auctionMaturity,
+  lockupPeriod
+} = network.names;
+
+describe('Namestate', function() {
+  describe('open auction name', function() {
+    const name = 'handshake';
+    const nameHash = rules.hashName(name);
+    let height = 0;
```
nodech, line 29: Can this be rewritten as set of unit tests instead of shared state between tests?

test/namestate-test.js
```diff
@@ -0,0 +1,341 @@
...
+            'revokePeriodStart',
+            'revokePeriodEnd',
+            'blocksUntilReopen',
+            'hoursUntilReopen'
+          ]
+        );
+        heightWithRevoke++;
+      }
+
+      while (heightWithRevoke < renewalWindow + 10) {
```
nodech, line 252: This code does not run.

nodech (commented):
Is it okay to have negative numbers in blocksUntilExpire ? (Maybe have better alias if we want to keep this for backwards compatibility reasons)

test/namestate-test.js
```diff
@@ -0,0 +1,341 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+/* eslint no-unused-vars: "off" */
```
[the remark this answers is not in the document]
pinheadmz: I think I had unused values in here, but I'll fix:

```
const {
  treeInterval,
  biddingPeriod,
  revealPeriod,
  renewalWindow,
  claimPeriod,
  auctionMaturity,
  lockupPeriod
} = network.names;
```

pinheadmz:
@nodech you're right on about negative `blocksUntilExpire`, I had a bug given that claimed names don't expire until the 4-year claim period is over. That is fixed in f82f783e394f06cf7fccbd6c7877ac583314b25e

test/namestate-test.js
pinheadmz: Yeah you got me I'm sloppy. OK this has been updated in b6b55b28599a4392ac1aacae0cd7ef1cd3e836cf

I still used a single `describe` block to run through the single linear auction flow, but then after that the "forked timeline" stuff has been replaced with hard coded start heights for the block the auction ends on.

pinheadmz:
I just realized hsd has never reported expiration date correctly for reserved names. A reserved name does not expire until AFTER the reserved name period if over (four years). The logic can be seen in `namestate.isExpired()`:

https://github.com/handshake-org/hsd/blob/107ed2be9280bb4292d21330e7d1ace62da9a002/lib/covenants/namestate.js#L208-L218

(note the typo in the comment, line 212 should say "two years")

However, this API tells me that the reserved name `torrentfreak` will expire in less than two months:

```
# hsd-rpc getnameinfo torrentfreak
{
  "start": {
    "reserved": true,
    "week": 17,
    "start": 19152
  },
  "info": {
    "name": "torrentfreak",
    "nameHash": "ea821088b72047a89e1defa25f91194760493a6c2070a289fa8c9a73c6428861",
    "state": "CLOSED",
    "height": 5670,
    "renewal": 15104,
    "owner": {
      "hash": "b2e9a0a48438ca58050a8b77cb1ebb6d8c1767c2de24135b33431501f46f3863",
      "index": 0
    },
    "value": 0,
    "highest": 0,
    "data": "0001036e733103646e73046c6976650001036e7332c00601036e7333c00606011a723d68747470733a2f2f746f7272656e74667265616b2e636f6d0601316e61636c3d527774776f3750472b736b6233356d6c6d6f7a32756356457677355063304b495855434a594d5a6833766b3d",
    "transfer": 0,
    "revoked": 0,
    "claimed": 1,
    "renewals": 0,
    "registered": true,
    "expired": false,
    "weak": true,
    "stats": {
      "renewalPeriodStart": 15104,
      "renewalPeriodEnd": 120224,
      "blocksUntilExpire": 8057,
      "daysUntilExpire": 55.95
    }
  }
}
```

nodech:
Fixed expire date as it needs to be: Max(claimPeriodEnd, normal end). If claimed name has been transferred or renewed, it wont expire on claim period end, instead when it should normally expire. But if it has not been transfer/renewed then it will expire on claimPeriod end.

PR #558, opened 2021-02-09
Make Root nameserver more flexible for plugins

### LINKED ISSUE #541: DNS server should be more flexible to plugins

pinheadmz:
Closes #541

TODO: Include example plugins to close...
- [x] dns/server: blacklist rfc2606 domains in the resolver #544
- [x] Handshake TLDs .0 and .1 ... .255 confusion with IPv4 Namespace #455
- [x] Blacklist should include homographs of ICANN TLDs #255
- [x] On Resolving Alternative DNS Systems #208

### DIFF
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -51,17 +51,6 @@ const TYPE_MAP = Buffer.from('000722000000000380', 'hex');
 const RES_OPT = { inet6: false, tcp: true };
 const CACHE_TTL = 30 * 60 * 1000;
 
-const blacklist = new Set([
-  'bit', // Namecoin
-  'eth', // ENS
-  'exit', // Tor
-  'gnu', // GNUnet (GNS)
-  'i2p', // Invisible Internet Project
-  'onion', // Tor
-  'tor', // OnioNS
-  'zkey' // GNS
-]);
-
 /**
  * RootCache
  */
@@ -121,6 +110,19 @@ class RootServer extends DNSServer {
     this.port = 5300;
     this.lookup = null;
     this.publicHost = '127.0.0.1';
+
+    // Plugins can add or remove items from
+    // this set before the server is opened.
+    this.blacklist = new Set([
+      'bit', // Namecoin
+      'eth', // ENS
+      'exit', // Tor
+      'gnu', // GNUnet (GNS)
+      'i2p', // Invisible Internet Project
+      'onion', // Tor
+      'tor', // OnioNS
+      'zkey' // GNS
+    ]);
 
     this.cache = new RootCache(3000);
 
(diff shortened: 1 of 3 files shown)


pinheadmz:
All of the use cases for this modification are exemplified in this WIP hsd plugin: https://github.com/pinheadmz/holdmyhand

The one exception might be "alternative DNS systems" but the plugin can force lookup of certain names to ICANN (before checking the HNS root zone) and it should be clear to developers that the same method can be used to redirect queries to other systems resolvers like ENS, IPFS, Sia, Footnote, etc. They will just have to return a [`Message`](https://github.com/chjj/bns/blob/master/lib/wire.js#L165) object the same way `ns.icann.lookup()` does.

pinheadmz:
Moving the guide to hsd-dev.org, stay tuned for PR over there...

pinheadmz:
Converted to draft while working on more plugin implementations and ensure everything is properly covered.

pinheadmz:
Okay I think this one is ready. It should be a total NO-OP unless a user has a plugin installed, examples include https://github.com/imperviousinc/handover and https://github.com/pinheadmz/holdmyhand.

Docs will be updated when merged, working on that here: https://github.com/handshake-org/handshake-org.github.io/pull/113

buffrr:
Nice! I tried holdmyhand and handover. Having the ability to customize how the root nameserver works with plugins is going to be very useful.

lib/dns/server.js
```diff
@@ -395,6 +398,27 @@ class RootServer extends DNSServer {
     const {name, type} = qs;
     const tld = util.from(name, -1);
 
+    // Plugins can insert middleware here and hijack the
+    // lookup for special TLDs before checking Urkel tree.
+    // We also pass the entire question in case a plugin
+    // is able to return an authoritative (non-referral) answer.
+    if (typeof this.middle === 'function') {
```
nodar-chkuaselidze, line 405: Maybe hook or resolveHook would be more intuitive name?

pinheadmz: Hmmm, appreciate the feedback. I've already written two plugins using this and documented the `middle()` property here: https://hsd-dev.org/guides/plugins.html 🤷‍♂️

Issue #510, opened 2020-09-02
Assertion error when rescanning a block with TRANSFER/FINALIZE for wallet


pinheadmz:
This could be related to #464 but after looking at the actual assertion error I think it is a different bug we need to look in to.
 @HDardenne experienced this issue and brought to my attention on telegram, using Bob wallet. Another user on telegram had the same assertion error using hsd v2.1.5 but claimed the error went away after upgrading to v2.2.0.

The assertion errors are not identical but clearly related: one throws on a FINALIZE when the transfer is not in a transferring state (`ns.transfer === 0`) the other throws on a TRANSFER when the name is *already* in a transferring state (`ns.transfer !== 0`)

### Finalize when name is not in transfer state
```
AssertionError [ERR_ASSERTION]: Assertion failed.
    at TXDB.connectNames (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\txdb.js:2061:13)
    at async TXDB.insert (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\txdb.js:1021:23)
    at async Wallet._add (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\wallet.js:4099:21)
    at async Wallet.add (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\wallet.js:4084:14)
    at async WalletDB._addTX (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\walletdb.js:2252:11)
    at async WalletDB._addBlock (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\walletdb.js:2091:13)
    at async WalletDB.rescanBlock (C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\walletdb.js:2196:7)
    at async C:\Users\Hugo\AppData\Local\Programs\bob-wallet\resources\app.asar\node_modules\hsd\lib\wallet\walletdb.js:150:9
[...]

PR #564, opened 2021-02-24
wallet: save and recover undo data from critical namestate transitions

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/510

This PR fixes a wallet issue that throws an assertion error during a rescan in a shockingly simple scenario:

- Alice OPENs a name. It doesn't matter if she bids or not
- Bob wins the name
- Bob sends a TRANSFER for the name. It doesn't matter if it's ever FINALIZEd or what address the TRANSFER commits to
- Alice rescans --> `assertion error  at TXDB.connectNames`

## Background

There are other related scenarios where Alice doesn't send the OPEN but does bid, and still loses. The crux of the issue has to do with UNDO DATA. Blockchain is generally designed to be append-only and therefore only move forwards. However there are scenarios in which the chain is rewound: a reorganization or a wallet rescan. In these scenarios, the software depends on UNDO data, which are essentially snapshots of state just before a transition happens. To move backwards, the undo data is retrieved and applied. Essentially for every new block and every wallet transaction, a "negative" is stored in the database and used to reverse whatever state update the action caused initially.

The Handshake wallet is a mysterious beast because a lot of its state depends on the actions of OTHER USERS, like tracking bids and determining who won an auction. Bids and Reveals are stored in special buckets in the database so that they can be reverted even if they don't actually directly affect the wallet's _balance_.

TRANSFER, FINALIZE, and REVOKE transactions up until now are _not_ stored in special buckets in the database. In fact, they are not stored at all unless they directly affect the wallet's balance. That's what this PR changes. We still don't need to store the entire TX but we do need to store the UNDO data, and we need to remember to retrieve it if one of the above "rewind" scenarios is triggered.

With all this background I can now explain the bug at the lowest level: When the wallet rescans, it starts by reverting / undoing / unconfirming all the transactions in every block from chain tip backwards to the rescan target (provided as an argument by the user, default `0` aka genesis block). Before this PR, the wallet DID NOT revert transactions like TRANSFERs unless they involved the wallet directly. Because of this, when the wallet rescans the chain it encounters the TRANSFER a second time, but throws an error because it thinks the name is still in a transfer state (`ns.transfer !== 0`)

## To test / fix your own wallet

If you've encountered this bug, you can test this PR by attempting to fix your own wallet:

```sh
# enter hsd repo
cd /path/to/hsd

# download this branch
git remote add pinheadmz https://github.com/pinheadmz/hsd
git fetch pinheadmz
git checkout rescantransfer1

# stop hsd (or killall hsd)
hsd-rpc stop

# back up your wallet
cp -r ~/.hsd/wallet ~/.hsd/wallet-backup-for-rescantransfer1

# restart hsd
hsd

# deep clean the wallet and rescan, may require API key
curl 127.0.0.1:12039/deepclean -X POST --data '{"I_HAVE_BACKED_UP_MY_WALLET":true}'
hsw-cli rescan 0

# watch wallet height and wait until it is fully synced (check height in JSON result)
hsw-rpc getwalletinfo

# rescan a second time
hsw-cli rescan 0

# check
hsw-rpc getwalletinfo
hsw-cli balance
```

## Bonus

Because this PR is urgent and likely to get merged and released quickly, I attached a rider ;-)

I cherry picked https://github.com/handshake-org/hsd/pull/499/commits/4b75f8aa8e45d325f8b7b6501767ff98fceebda9 from https://github.com/handshake-org/hsd/pull/499 which bypasses a too-strict length requirement in `Address` length. (see also https://github.com/bcoin-org/bcoin/issues/989) Without this fix,  assertion errors are still thrown during rescan for wallets (like mine) that have ever sent to a nulldata address. I had to add this bug fix to test this PR against my own wallet.

## TODO

Just by eyeballing the code, I assume that this bug will have the same effect if Bob (see above) sends a FINALIZE or REVOKE -- because these covenant types update namestate properties that are also `assert()`'ed in txdb. So this PR was written to also catch those types but I still need to add explicit tests

- [x] ~Add tests for FINALIZE and REVOKE~ 5013fffa06067bb646f3188a79490acbb3f5746b

## 💰 PR REVIEW BOUNTY: 💰

10 HNS: add a comment to the PR with an original meme expressing how tired we are all of all the damn wallet bugs
100 HNS: add a comment with a screen shot proving you pulled the branch and ran the test locally: `npm run test-file test/wallet-rescan-test.js`
1000 HNS: add a comment that actually influnces me to change the pull request in some way. if its just a typo/ nit ill send you..... oh i dunno 200 HNS

### DIFF
--- a/test/wallet-rescan-test.js
+++ b/test/wallet-rescan-test.js
@@ -1,0 +1,335 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+/* eslint no-implicit-coercion: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const FullNode = require('../lib/node/fullnode');
+const MemWallet = require('./util/memwallet');
+const Network = require('../lib/protocol/network');
+const Address = require('../lib/primitives/address');
+const rules = require('../lib/covenants/rules');
+const {Resource} = require('../lib/dns/resource');
+const {forValue} = require('./util/common');
+
+const network = Network.get('regtest');
+
+const {
+  treeInterval,
+  biddingPeriod,
+  revealPeriod
+} = network.names;
+
+describe('Wallet Rescan with TRANSFER', function() {
+  describe('Only sends OPEN', function() {
+    // Bob runs a full node with wallet plugin
+    const node = new FullNode({
+      network: network.type,
+      memory: true,
+      plugins: [require('../lib/wallet/plugin')]
+    });
+    node.on('error', (err) => {
+      assert(false, err);
+    });
+
+    const {wdb} = node.require('walletdb');
+    let bob, bobAddr;
+
+    // Alice is some other wallet on the network
+    const alice = new MemWallet({ network });
+    const aliceAddr = alice.getAddress();
+
+    // Connect MemWallet to chain as minimally as possible
+    node.chain.on('connect', (entry, block) => {
+      alice.addBlock(entry, block.txs);
+    });
+    alice.getNameStatus = async (nameHash) => {
+      assert(Buffer.isBuffer(nameHash));
+      const height = node.chain.height + 1;
+      const state = await node.chain.getNextState();
+      const hardened = state.hasHardening();
+      return node.chain.db.getNameStatus(nameHash, height, hardened);
+    };
+
+    const NAME = rules.grindName(4, 4, network);
+
+    async function mineBlocks(n, addr) {
+      addr = addr ? addr : new Address().toString('regtest');
+      const blocks = [];
+      for (let i = 0; i < n; i++) {
+        const block = await node.miner.mineBlock(null, addr);
+        await node.chain.add(block);
+        blocks.push(block);
+      }
+
+      return blocks;
+    }
+    before(async () => {
+      await node.open();
+      bob = await wdb.create();
+      bobAddr = await bob.receiveAddress();
+    });
+
+    after(async () => {
+      await node.close();
+    });
+
+    it('should fund wallets', async () => {
+      const blocks = 10;
+      await mineBlocks(blocks, aliceAddr);
+      await mineBlocks(blocks, bobAddr);
+
+      const bobBal  = await bob.getBalance();
+      assert.strictEqual(bobBal.confirmed, blocks * 2000 * 1e6);
+      assert.strictEqual(alice.balance, blocks * 2000 * 1e6);
+    });
+
+    it('should run auction', async () => {
+      // Poor Bob, all he does is send an OPEN but his wallet will
+      // watch all the other activity including TRANSFERS for this name
+      await bob.sendOpen(NAME, true);
+      const openBlocks = await mineBlocks(1);
+      // Coinbase plus open
+      assert.strictEqual(openBlocks[0].txs.length, 2);
+
+      // Advance to bidding phase
+      await mineBlocks(treeInterval);
+      await forValue(alice, 'height', node.chain.height);
+
+      // Alice sends only bid
+      const aliceBid = await alice.createBid(NAME, 20000, 20000);
+      await node.mempool.addTX(aliceBid.toTX());
+      const bidBlocks = await mineBlocks(1);
+      assert.strictEqual(bidBlocks[0].txs.length, 2);
+
+      // Advance to reveal phase
+      await mineBlocks(biddingPeriod);
+      const aliceReveal = await alice.createReveal(NAME);
+      await node.mempool.addTX(aliceReveal.toTX());
+      const revealBlocks = await mineBlocks(1);
+      assert.strictEqual(revealBlocks[0].txs.length, 2);
+
+      // Close auction
+      await mineBlocks(revealPeriod);
+
+      // Alice registers
+      const aliceRegister = await alice.createRegister(
+        NAME,
+        Resource.fromJSON({records:[]})
+      );
+      await node.mempool.addTX(aliceRegister.toTX());
+
+      const registerBlocks = await mineBlocks(1);
+      assert.strictEqual(registerBlocks[0].txs.length, 2);
+    });
+
+    it('should get namestate', async () => {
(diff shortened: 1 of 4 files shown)


pinheadmz:
Requesting reviews from: @brandondees @chikeichan

rozaydin:
i am hoping to get that 100 hns

test/wallet-rescan-test.js
```diff
@@ -0,0 +1,335 @@
...
+    });
+
+    const {wdb} = node.require('walletdb');
+    let bob, bobAddr;
+
+    // Alice is some other wallet on the network
+    const alice = new MemWallet({ network });
+    const aliceAddr = alice.getAddress();
+
+    // Connect MemWallet to chain as minmally as possible
```
rozaydin, line 44: nit, minimally

chjj:
This fixes the assertion error, but I'm now wondering what other actions are potentially affected by this bug. There could be some edge cases where updates/renewals/etc. are not rolled back properly. Granted, it's for a name you probably don't particularly care about (since you don't own it), but there may be other implications down the line.

To be safe we could consider updating the block record/map for damn near everything. Though, that might get funky too.

We should merge/release this but also continue thinking about potential edge cases here.

test/wallet-rescan-test.js
pinheadmz: thanks fixed

pinheadmz:
@rozaydin thank you for reviewing! I sent you 300 HNS

brandondees:
`hs1qexn22kqtlle9nqyzg2pkn3ldkw0ksw4harymq0` 

Screenshot from 2021-02-24 13-10-04

tests passing here, looking at the rest now

brandondees:
@chjj any idea of what exactly might get funky with tracking everything? seems like the sensible and safe thing to do from my limited view

test/wallet-rescan-test.js
```diff
@@ -0,0 +1,516 @@
...
+
+    // Alice is some other wallet on the network
+    const alice = new MemWallet({ network });
+    const aliceAddr = alice.getAddress();
+
+    // Connect MemWallet to chain as minimally as possible
+    node.chain.on('connect', (entry, block) => {
+      alice.addBlock(entry, block.txs);
+    });
+    alice.getNameStatus = async (nameHash) => {
```
brandondees, line 48: this might be a nitpick but why is this function being defined on an alice object when it's a pure function? is it an override?

pinheadmz: yes, `alice` here is a test utility called `MemWallet` which is just a dumb transaction generator. Its not the actual wallet being tested (`Bob` is -- with a fullnode) and so `MemWallet` is missing some functions and doesn't have access to chain data so when it needs to place an open for example, we need to stub this function.

test/wallet-rescan-test.js
```diff
@@ -0,0 +1,516 @@
...
+        NAME,
+        Resource.fromJSON({records:[]})
+      );
+      await node.mempool.addTX(aliceRegister.toTX());
+
+      const registerBlocks = await mineBlocks(1);
+      assert.strictEqual(registerBlocks[0].txs.length, 2);
+    });
+
+    it('should get namestate', async () => {
```
brandondees, line 132: it looks like these test blocks are sequentially order dependent, right? this might should be spelled out somewhere up top or at the repo/contributing.md level since it's a general best practice to write test cases that are strictly independent and randomize the order.

pinheadmz: thats true, we should mention that in contirbuting. In this case I am using separate `describe` blocks for overall test cases, and `it` blocks to verify incrementally. Its a standard we frequently break in this repo.

test/wallet-rescan-test.js
```diff
@@ -0,0 +1,516 @@
...
+    };
+
+    const NAME = rules.grindName(4, 4, network);
+
+    // Block that confirmed the bids
+    let bidBlockHash;
+    // Hash of the FINALIZE transaction
+    let aliceFinalizeHash;
+
+    async function mineBlocks(n, addr) {
```
brandondees, line 289: i feel like all this setup procedure stuff could be a lot more DRY just by moving it up a scope. this could become cumbersome if some of the internal apis change and a lot of test code has to be changed.

brandondees: or if it's special per context it should probably be highlighted what exactly is special in each place

test/wallet-rescan-test.js
brandondees: ok good to know, makes sense

test/wallet-rescan-test.js
brandondees: cool, well then yeah just something newcomers should be able to avoid tripping over by viewing the key docs

brandondees:
Read through all the changes, all looks good to me overall, overlooking style preference stuff.

I think the test cases are probably adequate but could maybe be expanded to prove that balances aren't impacted too, if that's a potential side effect.

test/wallet-rescan-test.js
pinheadmz: true, lots of test modules have a `mineBlocks` function. They are usually slightly different in each case but we could probably add this to a `util` file in a future PR. It would need to accept a chain and miner as parameters though, which are usually local to each test. In this case I need the function to return the blocks, but thats not always the case in other test that mine blocks.

brandondees: gotcha, yeah i could see building up a small library of these sorts of things to raise the level of abstraction for various bob/alice scenarios but you'd need to find the common patterns. obvs all tangential to this branch, just something that sticks out to me as worth improving.

chikeichan:
tested import and rescan in [this branch](https://github.com/kyokan/bob-wallet/tree/test-rescantransfer1) in Bob. My wallet used to run into rescan loop on second rescan after first import. Second and subsequent rescans now can perform without issues on using this PR.

Issue #565, opened 2021-02-25
Should HNS root nameserver serve TXT records?


pinheadmz:
If you ask a DNS expert, they will tell you "no" without hesitation: the root zone is referral only -- it is not authoritative over the top-level domain and therefore can not answer questions about its records, it can only refer you to that zone. This has already produced discrepancies in user experience like this example:

### `hsd` (with unbound as recursive resolver)

```
$ dig @127.0.0.1 -p 5350 proofofconcept TXT +short
"THIS TXT RECORD IS HOSTED ON MY ACTUAL ZONE!!!"
```

### `hsd --rs-no-unbound` (using `bns` as recursive resolver)

```
$ dig @127.0.0.1 -p 5350 proofofconcept TXT +short
"A simple social network for Handshake"
"support: hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x"
```

The blockchain records are what is returned in the `bns` example:

```
$ hsd-rpc getnameresource proofofconcept
{
  ...
    {
      "type": "TXT",
      "txt": [
        "A simple social network for Handshake"
      ]
    },
    {
      "type": "TXT",
      "txt": [
        "support: hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x"
      ]
    },
  ...
}
``` 

`bns` is able to recover the on chain records directly because it sends a complete request to the HNS root:

```
recursive -> root (query):
  proofofconcept. IN TXT
recursive <- root  (response):
  proofofconcept. IN TXT  "A simple social network for Handshake"
  proofofconcept. IN TXT  "support: hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x"
```

However, unbound (at least the way it is configured by default in hsd) is more private about its queries -- it asks for a referral first and always, then once it knows who is authoritative over the zone it asks _them_ for an answer:

```
recursive -> root (query):
  proofofconcept. IN A
recursive <- root  (response):
  proofofconcept. IN NS ns.proofofconcept
  proofofconcept. IN A 142.93.115.133
recursive -> ns.proofofconcept @142.93.115.133 (query):
  proofofconcept. IN TXT
recursive <- ns.proofofconcept @142.93.115.133 (response):
  proofofconcept. IN TXT  "THIS TXT RECORD IS HOSTED ON MY ACTUAL ZONE!!!"
```

### DNSSEC

[...]

PR #566, opened 2021-02-25
dns: fix root zone dnssec

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/565

Based on https://github.com/james-stevens/hsd/commit/18cdabc43d79f88617e52aa56b18239e73a10709

TODO:
- [x] ~Maybe also stop serving TXT (https://github.com/handshake-org/hsd/issues/565)~ (06cf026817dee74c47b2abfb7088afc88f6ead1d)

Changes:
- Root always signs DS, never signs NS, and only signs TXT _if NS is not present_
- Root always sets `aa` (authority) bit for DS, never for NS, and for TXT only _if NS is not present_
- Root doesn't answer a query for TXT if NS is present (sends referral instead)
- Root sets authority bit for NXDOMAIN

### DIFF
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -322,9 +322,6 @@ class Resource extends Struct {
         res.authority = this.toNS(name);
         res.additional = this.toGlue(name);
         break;
-      case types.TXT:
-        res.answer = this.toTXT(name);
-        break;
       case types.DS:
         res.aa = true;
         res.answer = this.toDS(name);
--- a/lib/dns/resource.js
+++ b/lib/dns/resource.js
@@ -285,10 +285,12 @@ class Resource extends Struct {
       res.additional = this.toGlue(name);
 
       // Note: should have nsec unsigned zone proof.
-      if (!this.hasDS())
-        key.signZSK(res.authority, types.NS);
-      else
+      if (this.hasDS()) {
+        // DS records are the only record type
+        // the root zone is actually authoritative over.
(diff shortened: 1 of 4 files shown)


lib/dns/resource.js
```diff
@@ -311,17 +313,24 @@ class Resource extends Struct {
...
         res.additional = this.toGlue(name);
-        key.signZSK(res.authority, types.NS);
         break;
       case types.TXT:
-        res.answer = this.toTXT(name);
-        key.signZSK(res.answer, types.TXT);
+        if (!this.hasNS()) {
+          res.aa = true;
+          res.answer = this.toTXT(name);
+          key.signZSK(res.answer, types.TXT);
```
tuxcanfly, line 329: Should the PR description be updated to say we _do_ sign `TXT`? And `AA` is set for `TXT` as well?

> Root doesn't sign TXT or NS with ZSK, only DS
> Root doesn't set authority bit for TXT or NS, only DS

tuxcanfly (approved):
Tested. I checked against the linked issue and rationale seems to be to avoid conflict by preferring `NS` over `TXT` from the root zone. It maybe a good idea to update the PR with the conclusion from the linked issue and rationale for making this change.

lib/dns/resource.js
pinheadmz: thank you sir! updated the PR description to reflect final changes after #565 discussion

PR #567, opened 2021-02-26
test: provably locked anyone-can-renew/renew-only addresses for names

pinheadmz:
This PR just adds a test to hsd, but like #500 it also acts like a standardizing document for wallet implementations to use and most importantly, proves that the following addresses encode the following script hash. [Note that `8` is the type for RENEW](https://github.com/handshake-org/hsd/blob/a94ce87a886e943a68d4a3e67068f6e473c21156/lib/covenants/rules.js#L52)

```
Script: OP_TYPE OP_8 OP_EQUAL
Serialized script: d05887
Script hash: e466310e566f8f14ac36f7eb7607a5d77a2351ad6bb5aba20a17396c5b18b8c1

main:    hs1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqsw3f8kc
testnet: ts1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqsj8gmfv
regtest: rs1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqs570axm
simnet:  ss1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqs4kzusf
```

If you send a name to one of these addresses and FINALIZE. **No one will ever be able to do anything with that name except RENEW. And anyone can renew it by simply adding a miner fee and broadcasting.** This effectively "locks" the name forever, unless the entire community forgets to renew the name within the 2-year window.

### DIFF
--- a/test/anyone-can-renew-test.js
+++ b/test/anyone-can-renew-test.js
@@ -1,0 +1,341 @@
+/* eslint-env mocha */
+/* eslint prefer-arrow-callback: "off" */
+
+'use strict';
+
+const assert = require('bsert');
+const Network = require('../lib/protocol/network');
+const FullNode = require('../lib/node/fullnode');
+const MTX = require('../lib/primitives/mtx');
+const Address = require('../lib/primitives/address');
+const Output = require('../lib/primitives/output');
+const {Script, Opcode, Stack} = require('../lib/script');
+const rules = require('../lib/covenants/rules');
+const {Resource} = require('../lib/dns/resource');
+const {forValue} = require('./util/common');
+
+// ANYONE-CAN-RENEW address:
+//
+// Script: OP_TYPE OP_8 OP_EQUAL
+// Serialized script: d05887
+// Script hash: e466310e566f8f14ac36f7eb7607a5d77a2351ad6bb5aba20a17396c5b18b8c1
+//
+// main:    hs1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqsw3f8kc
+// testnet: ts1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqsj8gmfv
+// regtest: rs1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqs570axm
+// simnet:  ss1qu3nrzrjkd783ftpk7l4hvpa96aazx5dddw66hgs2zuukckcchrqs4kzusf
+
+const script = new Script([
+  Opcode.fromSymbol('type'),
+  Opcode.fromInt(rules.types.RENEW),
+  Opcode.fromSymbol('equal')
+]);
+const address = new Address().fromScript(script);
+
+const network = Network.get('regtest');
+
+const node = new FullNode({
+  memory: true,
+  network: 'regtest',
+  plugins: [require('../lib/wallet/plugin')]
+});
+
+const {wdb} = node.require('walletdb');
+
+let alice, aliceReceive;
+let bob, bobReceive;
+
+const name = rules.grindName(5, 1, network);
+const nameHash = rules.hashName(name);
+let heightBeforeOpen, heightBeforeRegister, heightBeforeFinalize;
+let coin;
+
+async function mineBlocks(n, addr) {
+  addr = addr ? addr : new Address().toString('regtest');
+  for (let i = 0; i < n; i++) {
+    const block = await node.miner.mineBlock(null, addr);
+    await node.chain.add(block);
+  }
+}
+
+describe('Anyone-can-renew address', function() {
+  before(async () => {
+    await node.open();
+
+    alice = await wdb.create();
+    bob = await wdb.create();
+
+    aliceReceive = await alice.receiveAddress();
+    bobReceive = await bob.receiveAddress();
+  });
+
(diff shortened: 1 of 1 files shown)


test/anyone-can-renew-test.js
```diff
@@ -0,0 +1,331 @@
...
+    mtx.inputs[0].witness.fromStack(witness);
+
+    mtx.check();
+
+    // Bob (aka "anyone") will broadcast
+    const heightBeforeRenewal = node.chain.height;
+    await node.sendTX(mtx.toTX());
+    await mineBlocks(1);
+
+    const ns = await node.getNameStatus(nameHash);
```
tuxcanfly, line 321: Maybe test that the name resource i.e TXT record is preserved here.

test/anyone-can-renew-test.js
```diff
@@ -0,0 +1,331 @@
...
+    }));
+    mtx.output(0).covenant.type = rules.types.RENEW;
+    mtx.output(0).covenant.pushHash(nameHash);
+    mtx.output(0).covenant.pushU32(heightBeforeOpen + 1);
+    mtx.output(0).covenant.pushHash(node.chain.tip.hash);
+
+    await bob.fund(mtx, {coins: [coin]});
+    await bob.finalize(mtx, {coins: [coin]});
+    await bob.sign(mtx);
+
```
tuxcanfly, line 308: Is there a possibility that someone might grief the network by spamming renewals on every name sent here, using dust fee?

test/anyone-can-renew-test.js
```diff
@@ -0,0 +1,331 @@
...
+  it('should not be owned by either wallet', async  () => {
+    assert.rejects(
+      alice.sendTransfer(name, aliceReceive),
+      {message: `Wallet does not own: "${name}".`}
+    );
+
+    assert.rejects(
+      bob.sendTransfer(name, bobReceive),
+      {message: 'Auction not found.'}
```
tuxcanfly, line 174: Just curious why this says 'Auction not found' instead of `Wallet does not own` above.

tuxcanfly (approved):
Tested ACK. On the usage of the script itself, feels like a double-edged sword. Going to lose control over updating the resource record to avoid the inconvenience of renewing seems like a pyrrhic victory. Then again, this is purely opt-in optional and totally allowed by the scripting system so idk.  Seems like it could be useful in a limited number of cases where updates don't matter.

ca98am79:
> On the usage of the script itself, feels like a double-edged sword. Going to lose control over updating the resource record to avoid the inconvenience of renewing seems like a pyrrhic victory. Then again, this is purely opt-in optional and totally allowed by the scripting system so idk. Seems like it could be useful in a limited number of cases where updates don't matter.

We implemented this with our TLD .badass.  You can see the [transfer history here](https://hnsnetwork.com/names/badass).  The use case for us is to lock the TLD so that 2nd-level domain owners do not have to trust the TLD owner.  It is provably owned by no one, but renewable by anyone.

test/anyone-can-renew-test.js
pinheadmz: It's because Alice sent the open, the bid, the transfer - everything. Bob is just an innocent bystander (aka "anyone") so his wallet has no namestate data at all. Alice's wallet has some namestate data because she bid and actually owned the name at some point in the past. So its like, in her "watchlist" (walletDB `nameMap`). But the UTXO that owns the name is not in her wallet or under her control.

You did make me curious though Im going to add a test to see if this action causes bob's wallet to start tracking the name, especially now that #564 has been merged.

test/anyone-can-renew-test.js
pinheadmz: That is a possibility but renewals in general are rate-limited already by two factors:

1) Renewals must be at least 1 tree interval apart (36 blocks): 

https://github.com/handshake-org/hsd/blob/f2702f873928814d6b5b4bb32985e3163b85f55a/lib/blockchain/chain.js#L1262-L1267

2) Renewals are limited to 600 per block:

https://github.com/handshake-org/hsd/blob/f2702f873928814d6b5b4bb32985e3163b85f55a/lib/protocol/consensus.js#L208

Of course this pull request doesn't change anything - the cat's out of the bag. @tynes and I actually spoke with the project founders about this potential rule-bypass during testnet and there was some discussion about altering `OP_TYPE` to more strictly enforce the _intent_ of the renewal requirement, but ultimately it didn't get implemented.

I suppose there is a potential future risk to users that deploy contracts like this, that in the future the community may deploy a fork and break the pattern but right now, and especially in the use case described by @ca98am79, it seems to be a benefit.

test/anyone-can-renew-test.js
pinheadmz: 👍

test/anyone-can-renew-test.js
pinheadmz: ok updated, Alice's wallet still tracks the name but Bob's never does.

chjj:
Just a thought: if people are upset by this behavior being abused on chain, it can be avoided with a soft-fork which terminates script execution if OP_TYPE's corresponding output is a RENEW.

Of course, there is a cost to this: names which utilize OP_TYPE must use TRANSFER/FINALIZE as their RENEW mechanism.

Issue #563, opened 2021-02-19
wallet should migrate to new lookahead value


pinheadmz:
Since https://github.com/handshake-org/hsd/pull/506 was merged, it only affects new wallets. Old wallets still created with the old value of `10` are unaffected, but we should migrate those wallets in the next version. This is probably not an issue with Bob wallet since users are instructed to delete their data directories and rebuild the wallet from seed anyway.

PR #569, opened 2021-03-03
wallet: --check-lookahead to update old wallets

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/563

This probably should have been included in https://github.com/handshake-org/hsd/pull/506

This PR adds an extra wallet option to upgrade wallets created before #506 was merged, that still have a lookahead of `10`. When the option is set, all accounts of all wallets are called with `Account.setLookahead(Account.MAX_LOOKAHEAD
)` which is now `200`. This will set the metadata lookahead value, generate whatever keys/addresses are necessary and write to the database. This only needs to be run once per walletDB, since the update is permanent.

Usage:

`hsd --wallet-check-lookahead`

Log messages:

```
[warning] (wallet) Setting lookahead for wallet primary account 0
[warning] (wallet) Setting lookahead for wallet ten account 0
```

Can be verified by this bash test script: https://gist.github.com/pinheadmz/029bb4ff773b548cf632a5c300e4c17b

output:

```
HEAD is now at 7ca1b71ff v2.1.0
79552

hsd v2.1.0 - initialize
wallet lookahead:
10
wallet getaddressesbyaccount length:
22

Stopping.
Previous HEAD position was 7ca1b71ff v2.1.0
Switched to branch 'synclookahead1'
79564

hsd v2.3.0 #synclookahead1 - open old wallet
wallet lookahead:
10
wallet getaddressesbyaccount length:
22

Stopping.
79575

hsd v2.3.0 #synclookahead1 - check lookahead option set
wallet lookahead:
200
wallet getaddressesbyaccount length:
402

Stopping.
79585

hsd v2.3.0 #synclookahead1 - no extra option set
wallet lookahead:
200
wallet getaddressesbyaccount length:
402
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,12 @@ The bid will be broadcasted either during the creation (`broadcastBid=true`) or 
 (`broadcastBid=false`).
 The reveal will have to be broadcasted at a later time, during the REVEAL phase.
 The lockup must include a blind big enough to ensure the BID will be the only input of the REVEAL
-transaction. 
+transaction.
+
+- Now parses option `--wallet-check-lookahead` (or `--check-lookahead` for standalone
+wallet node) that will check every account of every wallet in the DB and ensure
+the lookahead value is the current default and maximum of `200`. A rescan is
+recommended after this action.
 
 ### Node & Wallet API changes
 
--- a/lib/wallet/node.js
+++ b/lib/wallet/node.js
@@ -51,7 +51,8 @@ class WalletNode extends Node {
       cacheSize: this.config.mb('cache-size'),
       wipeNoReally: this.config.bool('wipe-no-really'),
       spv: this.config.bool('spv'),
-      migrate: this.config.uint('migrate')
+      migrate: this.config.uint('migrate'),
+      checkLookahead: this.config.bool('check-lookahead', false)
     });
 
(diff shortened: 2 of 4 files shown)


pinheadmz:
p.s there was a lot of debate about user-triggered wallet migration here: p.s. there was a lot of debate on user-triggered wallet migration here: https://github.com/handshake-org/hsd/pull/415

That's why I chose this route, setting an optional command line option

nodar-chkuaselidze:
Thanks for linking #415 to the PR, it helped a lot.

I have questions related to previous discussion mostly:
  - Is there reason why this was not written as `migration` ?

  To me this change is similar to `migrateChange`, fixes the
existing database state to something we desire. One big difference
being: migrate-change was critical bug, where this is just recommended setting.

 This works fine as well, it's just is it becuase this is not critical, does not
warrant it to be migration?
 We could reuse `migration=1` in that case.

Also migration is kinda misleading for the previous change to me. It is mostly
fix database state, not change database layout that would make it incompatible.

---

I guess this is note about previous PR: #506
Why not have Account.MAX_LOOKAHEAD = 255 and leave default to 200.
(This would make the value more obvious, maybe with comment?)
I thought 200 was just choses arbitrarily until looked at #506 and saw
that we store it in U8.

---

This works fine though.
LGTM.

PR #578, opened 2021-03-19
Create Automatic Docker deployment

Anunayj:
Adds Automatic docker builds using Github Actions.
Image tags used:

`latest`: For the latest *commit*, only built if npm ci passes. (ran after Build workflow)
`lts`: For latest *tagged* release. 
`*tag*`: Specific version/tag on git, will build whenever a commit with new tag is pushed.

`DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` GitHub secrets need to be set.

Images for the following platforms are made: linux/amd64, linux/arm64, linux/arm using the Dockerfile at root of repo.
(Both Windows and Mac users should be able to run these easily)

### DIFF
--- a/.github/workflows/docker-latest.yml
+++ b/.github/workflows/docker-latest.yml
@@ -1,0 +1,39 @@
+name: docker-latest
+
+# Push on commit if CI passes + Manually
+on:
+  workflow_run:
+    workflows: ["Build"]
+    branches: [master]
+    types: 
+      - completed
+
+  workflow_dispatch:
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    # Run on success and push (unsure if push there is necesarry)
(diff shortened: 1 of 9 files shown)


Anunayj:
Docker builds in their current state look like a security issue, since it's very hard to verify the contents of the docker image to be trustworthy, furthermore Docker Repos like DockerHub provide next to no protection against account compromises. (If account token get's leaked user can push/overwrite any tag).

There are ways to remedy this using [Docker Content Trust](https://docs.docker.com/engine/security/trust/). trusted contributers can sign images to indicate their trust (after verifying it's contents manually) and users can be told to verify signatures when pulling a image. The biggest problem with this method is, both the steps are manual (Unless you set a global docker flag to verify ALL images), and it is very likely that people will skip these steps.

For right now it is best to encourage users to fork this project and make their own Docker builds.

PR #581, opened 2021-03-22
added check for admin permissions on wallet create

randomlogin:
Previously admin token was not required for wallet creation, yet in documentation it was stated. During the update of hints it was discovered: https://github.com/handshake-org/hs-client/pull/35#discussion_r598968378

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -278,6 +278,11 @@ class HTTP extends Server {
 
     // Create wallet
     this.put('/wallet/:id', async (req, res) => {
+      if (!req.admin) {
+        res.json(403);
+        return;
+      }
+
       const valid = Validator.fromRequest(req);
 
       let master = valid.str('master');


nodar-chkuaselidze:
I don't think this is necessary, this API should not be public facing and you need API token anyway.
If malicious user has access to this API, I don't think the API is "ready" for that.

randomlogin:
> I don't think this is necessary, this API should not be public facing and you need API token anyway.
> If malicious user has access to this API, I don't think the API is "ready" for that.

Well, then I have to ask, if there is any reason to keep admin commands distinct from non-admin ones?

nodar-chkuaselidze:
If you consider this API were you expose it internally as a service to your other services (given u have to provide API-key), then you don't need to bother "admin" everytime you want wallet. But to me, this API is not polished for any specific use case so it's hard to argue which usage is more prominent and concentrate on.

You can use for separation of concerns and leave creation of the wallet just for the internal services.
Especially, if they also store private keys there. Even though this is not "secure" in a sense, that wallet service runner can intercept keys with just modified codebase.

If you consider using this as a more "public" service then it makes sense to manage creation of wallets and only allow "xpub" wallets..

So to be honest, I am not sure for what use cases we should improve the API.
This can be a good discussion about wallet in general.

pinheadmz:
Hm, @randomlogin is right, the `hsw-cli` help message lists `mkwallet` as an admin command but in the api docs, it is not in the admin section: https://hsd-dev.org/api-docs/?shell--cli#create-a-wallet

I agree with @nodech as well that the permissions for the API seem all over the place. I can imagine an internal system where employees are allowed to create wallets using just the API key, but admin is still required for rescan, deep clean, reveal master key, and list all wallets. I'm not sure where to go with this. I think I'm leaning more on the side of fixing the hsw-cli help message instead of adding the admin requirement to `mkwallet`

nodech:
Closing this for now - https://github.com/handshake-org/hs-client/pull/44

PR #583, opened 2021-03-31
Major fixes & improvements for getwork

chjj:
It appears some (if not most) miners are using `getwork`/`submitwork` instead of `getblocktemplate`. This is presumably because `getwork` is just easier: it requires minimal implementation and serialization on the part of the mining software. `getwork` was originally only included for testing immature mining clients, however, the community's reliance on it requires us to give it a second look.

There is a nasty bug present in `getwork` which causes blocks to be invalidated when transactions enter the mempool. It seems that miners have been patching this bug themselves to avoid mining invalid blocks. I figure we should make an official fix asap.

This PR fixes `getwork` to allow operation on stale blocks with older merkle roots. 

Some things are also simplified: the RPC no longer auto-generates a mask for the client. This really serves no purpose as mining pools should be doing this in their pool software, generating a mask in accordance with their share target. `submitblock` now takes an optional `mask` argument so the pool can notify hsd of the mask it generated.

To encourage the use of `getblocktemplate`, we make it a bit easier to use and migrate to: `{"capabilities":["coinbasetxn"]}` has been re-enabled in a sane way which should not conflict with popular bitcoin mining software (if it has been modified to support handshake). `merkleroot` and `witnessroot` are also added to the template if `coinbasetxn` is present (this assumes the client will not be modifying the coinbase).

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -3223,7 +3223,7 @@ class Chain extends AsyncEmitter {
       const state = await this.getState(prev, deployment);
 
       if (state === thresholdStates.LOCKED_IN
-          || state === thresholdStates.STARTED) {
+          || (state === thresholdStates.STARTED && deployment.force)) {
         version |= 1 << deployment.bit;
       }
     }
--- a/lib/mining/template.js
+++ b/lib/mining/template.js
@@ -362,29 +362,6 @@ class BlockTemplate {
   }
 
   /**
-   * Generate a random mask (2 bytes less than target).
-   * @returns {Buffer[]}
-   */
-
-  randomMask() {
-    const mask = random.randomBytes(32);
-
-    for (let i = 0; i < this.target.length; i++) {
-      mask[i] = 0;
-
-      if (this.target[i] !== 0) {
-        if (i + 1 < mask.length)
-          mask[i + 1] = 0;
-        break;
-      }
-    }
-
-    const hash = BLAKE2b.multi(this.prevBlock, mask);
-
-    return [mask, hash];
-  }
-
-  /**
    * Create raw block header with given parameters.
    * @param {Buffer} extraNonce
    * @param {Number} time
@@ -468,7 +445,7 @@ class BlockTemplate {
    */
 
   toCoinbase() {
-    return this.coinbase;
+    return this.coinbase.clone();
   }
 
   /**
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -116,6 +116,7 @@ class RPC extends RPCBase {
     this.boundChain = false;
     this.mask = Buffer.alloc(32, 0x00);
     this.merkleMap = new BufferMap();
+    this.merkleList = [];
     this.pollers = [];
 
     this.init();
@@ -2568,10 +2569,11 @@ class RPC extends RPCBase {
 
       this.refreshBlock();
       this.merkleMap.clear();
+      this.merkleList.length = 0;
     };
 
     this.node.on('connect', refresh);
-    this.node.on('disconnect', refresh);
+    this.node.on('reset', refresh);
 
     if (!this.mempool)
       return;
@@ -2610,16 +2612,6 @@ class RPC extends RPCBase {
 
     let attempt = this.attempt;
 
-    // This check is here to make sure a miner does not
-    // accidentally OOM the node. This means the merkle
-    // map is good for 2 hours without a block.
-    if (this.merkleMap.size >= 720) {
-      this.merkleMap.clear();
-
-      if (attempt)
-        this.merkleMap.set(attempt.witnessRoot, attempt);
-    }
-
     if (attempt) {
       if (attempt.address.isNull()) {
         throw new RPCError(errs.MISC_ERROR,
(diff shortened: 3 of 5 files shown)


lib/blockchain/chain.js
```diff
@@ -3223,7 +3223,7 @@ class Chain extends AsyncEmitter {
       const state = await this.getState(prev, deployment);
 
       if (state === thresholdStates.LOCKED_IN
-          || state === thresholdStates.STARTED) {
+          || (state === thresholdStates.STARTED && deployment.force)) {
```
pinheadmz, line 3226: Is this effectively the same fix we did here?

https://github.com/handshake-org/hsd/commit/5e20a5aaaa7d4254ba88b14acd3b85f025330029

lib/mining/template.js
```diff
@@ -468,7 +445,7 @@ class BlockTemplate {
    */
 
   toCoinbase() {
-    return this.coinbase;
+    return this.coinbase.clone();
```
pinheadmz, line 448: Makes sense to return a clone here, but I can't find reference to this function - what's it used for?

lib/node/rpc.js
```diff
@@ -1215,34 +1216,32 @@ class RPC extends RPCBase {
...
+      throw new RPCError(errs.MISC_ERROR, 'getwork');
 
     return this.createWork();
   }
 
   async submitWork(args, help) {
-    if (help || args.length !== 1)
-      throw new RPCError(errs.MISC_ERROR, 'submitwork ( "data" )');
+    if (help || args.length < 1 || args.length > 2)
+      throw new RPCError(errs.MISC_ERROR, 'submitwork ( "data" "mask" )');
```
pinheadmz, line 1227: nit: `"data"` isn't optional, shouldn't be inside parentheses.

pinheadmz:
Should would we be adding this `attempt` from `getTemplate()` to the `merkleMap` as well?

https://github.com/handshake-org/hsd/blob/2d1cbe9c17b0ad4e8858c06a8f85625dbee35ba9/lib/node/rpc.js#L2613-L2619

lib/node/rpc.js
```diff
@@ -1544,14 +1535,10 @@ class RPC extends RPCBase {
...
-      const input = tx.inputs[0];
-
-      // Pop off the nonces.
-      input.witness.pop();
-      input.witness.compile();
+      const tx = attempt.coinbase;
 
-      tx.refresh();
+      json.merkleroot = attempt.merkleRoot.toString('hex');
+      json.witnessroot = attempt.witnessRoot.toString('hex');
```
pinheadmz, line 1541: This is a convenient feature - client can still modify TXs in the block before submitting?

lib/node/rpc.js
```diff
@@ -2626,6 +2610,16 @@ class RPC extends RPCBase {
 
     let attempt = this.attempt;
 
+    // This check is here to make sure a miner does not
+    // accidentally OOM the node. This means the merkle
+    // map is good for 2 hours without a block.
+    if (this.merkleMap.size >= 720) {
```
pinheadmz, line 2616: Just to be clear, this is because we won't create a new template / add to `merkleMap` until at least 10 seconds pass. So if there are 720 items in the map, that means 7,200 seconds have passed (2 hours).

```js
      if (util.now() - this.lastActivity > 10)
        this.refreshBlock();
```

pinheadmz (approved):
Looks good, I just left a few nits and questions but I think this makes sense. I like using a null mask as default, all the pools do already and have had to hack hsd to do so. The `merkleMap` makes sense, only weird edge cases may be a miner switching from `getwork` to `getblocktemplate` or using both for some weird reason. The 2 hour timeout might leave a miner with a stale job but hopefully that is a far out edge case. Maybe in the future we can prune that map without wiping it out completely. 

ACK 6d6dec50062618b33194facdf598c6c3d314681f

```
Hash: SHA256

ACK 6d6dec50062618b33194facdf598c6c3d314681f
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmBks8UACgkQ5+KYS2KJ
yTpJrg//TAFQMiICydLN9ocVmvIf8MsTrg5BDJCEZMzX9ZhBOuz8H2XSs6ZxB2AO
peJ+mHY2gdJJ3W5xy/ZCbFcmZ+dTPwvFK9FeSHB1gpxENpd7EpMFIw3ean0yvP0p
qeel9iywyXpQ9WgTXOcYV8pYOZIXVZOtS007kFlA6HeskrSrVg5yM5u2ABmuK8Uh
0NQcn+gkXlIXNT75UZd129MPg1ivg70B1hcVeigKNrhXpxgQpxUvRJW5DYZR6XIb
LIkEamsTanROlYN2jzCamd/iZ7bYEBoblWMb9LmtfaEJ7voJnhQWQE8KOmjLJ+Fu
piKYCrn/jidN2b4zGEiBQTn2mwA6T5ry6urjGrYL711D1Eh/YQlrLfeuxEJ5VAnf
qrSx12LmhsMzOO2AGgfh1wSkqMB70rUQWmYGGBgMX14Db81Wy/4QMVU1UTQMJlWM
Dyir//fitqVPAc/ROalDDjzsVLg14uo/uT15FdrFrfbQKaOds3mYALJTG7uBEy32
SWKyYHrGBhw6FN19W/hL6MoCpUyduve0z6f9yVUFxsaI/t6+43AbTL7ztjgrVsvE
DkUL258Fjf3sHMVzRVMdzixV56OBSwiddh73DbsPyyEOdzLJpN/jI5c/4j+Ms4S3
uTRkxrv6kt8aZUBIMG6tN4WSqN3S5ZignN+JZxvNTZHnfZDPOpo=
=zK1H
-----END PGP SIGNATURE-----
```

</details>

0racl3z (commented):
awsome!

matuhinal:
Please connect me

PR #585, opened 2021-04-02
Inflation bug fix (covert->overt soft-fork)

chjj:
On March 24th, @pinheadmz discovered an inflation bug pertaining to replacement name claims. Other Handshake contributors were notified and action was quickly taken to deploy a covert soft-fork in collaboration with various miners. As of this moment, ~90% of the network hashrate is signaling the covert soft-fork, and we have confirmation that at least 93% of the network hashrate is _enforcing_ the soft-fork (but perhaps not signaling).

Chain analysis reveals that __this bug has never been exploited on mainnet__, and is now impossible to exploit given the hashrate enforcing this change.

The inflation bug could be exploited as follows:

1. Create a claim for a reserved name with a chosen fee.
2. Create a replacement claim with another fee.
3. Step 2 can now be repeated infinitely.

_All_ fees will be mined by the miner, and only final output value will actually be received by the name claimant.

Only miners stood to benefit financially from this bug. However, a malicious actor could have effectively destroyed the chain by causing hyper-inflation if they so desired.

The soft-fork works as follows:

- Initial claims are now required to commit to block 1 (note that commitments to block 0 were never allowed).
- Replacement claims (claimHeight > 1) are required to have the same output value as the previous claim output.
- Fees collected from replacement claims are required to be burned by miners (i.e. not added to the reward output).
- A frequency limit has been introduced for claims: consecutive replacement claims must be mined 288 blocks apart.
- A hard fee limit of 1000 coins has be introduced for claims (this prevents a registrar hacker from stealing absurd amounts of coins in collusion with a miner).
- A covert signaling mechanism was added to coinbase creation to track soft-fork enforcement (_not_ a consensus change).

There may have been a "better" way to do this, but the above fix worked nicely as a covert change since it does not modify the manner in which _initial_ name claims are handled. As such, it required no immediate changes to the wallet. The downside here is that replacement claims are now effectively zero-fee from the miner's perspective.

This PR also adds new re-org logic to revert an invalid re-org if a contextual error was encountered when connecting new blocks. This was an issue for miners, particularly in the context of the covert soft-fork deployment.

There is also a lesser bug here: the calculation of the total chain value was including _replacement_ claims. This resulted in a superficial error -- `gettxoutsetinfo` could return the wrong chain value after replacement claims have been mined. This PR includes a migration for this which should only take 2-3 minutes to run on boot.

Once this is merged and released, it is important that non-upgraded miners (~7% of them) upgrade immediately.

See https://handshake.org/notice/2020-04-02-Inflation-Bug-Disclosure.html for more information.

Much praise to @pinheadmz for discovering this bug! And a thank you to everyone who helped resolve this crisis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## unreleased
 
+### Chain & Consensus changes
+
+- A consensus inflation bug has been fixed. Non-upgraded miners should upgrade
+  as soon as possible. See
+  https://handshake.org/notice/2020-04-02-Inflation-Bug-Disclosure.html for
+  more information.
+- A new chain value migration is necessary (related to the above fix). This
+  migration will automatically run on boot and should only take 2-3 minutes.
+  Pruned nodes _cannot_ run this migration. Note that pruned nodes may have an
+  incorrect chain value until they re-sync.
+
 ### Wallet changes
 
 - Fixes a bug that caused rescans to fail if a name being "watched" was ever
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -58,7 +58,7 @@ class Chain extends AsyncEmitter {
     this.db = new ChainDB(this.options);
 
     this.locker = new Lock(true, BufferMap);
-    this.invalid = new LRU(100, null, BufferMap);
+    this.invalid = new LRU(5000, null, BufferMap);
     this.state = new DeploymentState(this.network.genesis.hash);
 
     this.tip = new ChainEntry();
@@ -641,6 +641,9 @@ class Chain extends AsyncEmitter {
     if (!this.state.hasHardening() && state.hasHardening())
       this.logger.warning('RSA hardening has been activated.');
 
+    if (this.height === this.network.deflationHeight)
+      this.logger.warning('Name claim deflation has been activated.');
+
     this.state = state;
   }
 
@@ -786,7 +789,7 @@ class Chain extends AsyncEmitter {
       throw new VerifyError(block,
         'invalid',
         'bad-cb-amount',
-        100);
+        0);
     }
 
     // Push onto verification queue.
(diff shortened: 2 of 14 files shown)


pinheadmz (approved):
ACK 24100d4dc7970243dd13bb87f0e92a0303848ab8

```
Hash: SHA256

ACK 24100d4dc7970243dd13bb87f0e92a0303848ab8
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmBnYpUACgkQ5+KYS2KJ
yTqVxBAAz8bbVzlRgyScYdrHKrgbD9f1L9vpl22p2KfzCqJKur211mf9nY3g+ieQ
zDCNeN+EG+FcdPO7bJi5YazQK2Nxl1Ck63ce57SE/YMrAORstUSvdbAE1UjLO8Ju
4dJY5lWZs6AP9coyDnU620r54ItFvtRvcuTkTkzGwQFMSAI/6ckdvF0xhizDQWcf
FXPJs55aNyQoihMB+BjLwwzPWw/pkuru+DxoB20iL0waIf3r1NDflp3YCEhkTP0Z
g8jYpNb0mNaLydPZ18DoyGtTsr99u/3ToJdhzH9/uyDODgP8Ogmpv4XrL5iXb+nL
hNaTPMrQ8PBSjFqcHV/GmgqXSyxj4BiQCJ+GKJjcdO7ioJf4wxtaS/7yIt+teIBk
gVFgjIX9lBelfO67o8VCgcT3jwO4DTsGwAjjiIutOig3tkQvISZqTs+ckpjeT3xk
siPZ9i5oA60smNcDFWNxV3eDi+DF0ZzB2IFKvtQbJXpfx4+SfbEQRM0hlKK/lQDN
lKxhRS8T6OrkLPg/N22GxOdfQXKV4ONANfsizd3zdVPCix41/kNXOWX4gzFqt+zC
bwinWW7Hv3UAIX32IMs/guZ49b7VXrJHO3kXBv6Uf1zBTlgZ3Bi94E6SK29Oab4H
RIVEtUJ3mj0EELCRoo7rtSatjc9ijuF4Ed4hLb3fcrEiaTL2Ksk=
=PmiC
-----END PGP SIGNATURE-----
```

</details>

PR #589, opened 2021-04-06
eslint proposal

rozaydin:
Hello,

First of all, HSD is a great project and i really mean it, thank you for putting this amazing effort to keep it running.
I do have a proposal regarding the es-lint rules currently employed at the HSD project. 

How would you feel to extend the ruleset from "airbnb-base", It already have many rules that, i believe will help improving the code-quality and make it easy to maintain the code ?

I did a dry run and disabled some of the rules which i think contradicts a lot with the hsd code base, it complains a lot only for the test files and relatively less complaints for the hsd code base.

This PR by no means is final, i just wanted to use it to bring to your attention, but if you agree i am also willing to apply to the hsd and create a real PR ? 

PS: please discard the systemd files

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,18 +1,18 @@
 {
   "env": {
-    "es6": true,
+    "es2017": true,
     "node": true
   },
-  "extends": "eslint:recommended",
+  "extends": "airbnb-base",
   "globals": {
-    "Atomics": "readable",
-    "BigInt": "readable",
-    "BigInt64Array": "readable",
-    "BigUint64Array": "readable",
-    "queueMicrotask": "readable",
-    "SharedArrayBuffer": "readable",
-    "TextEncoder": "readable",
-    "TextDecoder": "readable"
+    "Atomics": "readonly",
+    "BigInt": "readonly",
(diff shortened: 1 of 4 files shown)


pinheadmz:
I'm open to discussing style improvements like this, although I've been writing in JJ's style for a few years and it may be hard to break some habits 😆 . I haven't tried running this locally yet -- can you give an example of some style changes you think we should make? Or is there something specifically about these `eslint` modules, like, API or functionality-wise that you prefer to what we have?

Also worth noting, this project was carefully built to have essentially zero dependencies. Everything including `mocha` and `eslint` have been re-written and/or vendored for security reasons, especially as JS is known for crazy package-dependency messes. So it might not gain a lot of traction from contributors just because it adds a new package to the build. 

I think back in [this PR](https://github.com/handshake-org/hsd/pull/484) we discussed adding JJ's vendored `bslint` to `devDependencies` and then ultimately decided against it anyway.

Falci:
> So it might not gain a lot of traction from contributors just because it adds a new package to the build.

IMO, this is the best part of this PR.

Those are "devDependencies" and are not needed for a production install (`npm install --production`)

Issue #586, opened 2021-04-03
Redeem all via curl


befranz:
I try to **redeem all** via curl which should work by passing null as name, but whatever I try (also omit name, name as empty string) I get error "No passphrase." Redeeming single names works fine. 

curl http://x:7cc6fba603b1740d03bb43e14f7005e4cd0b8fff@127.0.0.1:12039/wallet/bob1/redeem -X POST --data '{"passphrase":"mypwd","name":null,"broadcast":true,"sign":true}'

{
  "error": {
    "message": "No passphrase."
  }
}

hsd 2.4.0

Additional info:
Like @pinheadmz suggested sending an unlock tx beforehand circumvents this issue.

pinheadmz:
Good catch, thanks. Check it out here in the code -- we aren't sending any `options` (including `passphrase`) to the `sendRedeemAll()` function! Should be an easy fix but I'll add some tests as well...

https://github.com/handshake-org/hsd/blob/9f0361f7e5b0bf0d2de6879c4d149991a45aab32/lib/wallet/http.js#L1163-L1174

PR #590, opened 2021-04-06
wallet: pass passphrase + TX options to sendRedeemAll())

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/586

When a user tries to redeem multiple losing bids for multiple names by calling `sendRedeem` with `name: null`, the http module calls a specific function `sendRedeemAll()`. This function was not getting passed the wallet passphrase, and so encrypted wallets would always fail. 

This PR simply passes the passphrase (and the rest of the TX options like `subtractFee`, etc) to that function.

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -1160,13 +1160,15 @@ class HTTP extends Server {
 
       assert(broadcast ? sign : true, 'Must sign when broadcasting.');
 
-      if (!name) {
-        const tx = await req.wallet.sendRedeemAll();
-        return res.json(200, tx.getJSON(this.network));
-      }
-
       const options = TransactionOptions.fromValidator(valid);
-      const mtx = await req.wallet.createRedeem(name, options);
+
+      let mtx;
+
+      if (!name) {
+        mtx = await req.wallet.createRedeemAll(options);
+      } else {
+        mtx = await req.wallet.createRedeem(name, options);
+      }
 
       if (broadcast) {
         const tx = await req.wallet.sendMTX(mtx, passphrase);
--- a/test/wallet-http-test.js
+++ b/test/wallet-http-test.js
@@ -45,6 +45,7 @@ const wclient = new WalletClient({
   apiKey: 'foo'
 });
 
+const {wdb} = node.require('walletdb');
 const wallet = wclient.wallet('primary');
 const wallet2 = wclient.wallet('secondary');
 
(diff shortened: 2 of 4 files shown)


lib/wallet/http.js
```diff
@@ -1160,12 +1160,13 @@ class HTTP extends Server {
 
       assert(broadcast ? sign : true, 'Must sign when broadcasting.');
 
+      const options = TransactionOptions.fromValidator(valid);
+
       if (!name) {
-        const tx = await req.wallet.sendRedeemAll();
+        const tx = await req.wallet.sendRedeemAll({passphrase, ...options});
```
nodech, line 1166: I think this does not respect `broadcast` option.

We could have `const mtx = await req.wallet.createRedeemAll(options);` and do the same thing as we do for mtx bellow.

Can be done bellow where `mtx` is defined, make it `let mtx` and create that MTX based on `!name`.

nodech (changes requested):

pinheadmz:
Do we need to apply this to the corresponding rpc call?

https://github.com/handshake-org/hsd/blob/9f0361f7e5b0bf0d2de6879c4d149991a45aab32/lib/wallet/rpc.js#L2161-L2163

nodech:
> Do we need to apply this to the corresponding rpc call?
> 
> https://github.com/handshake-org/hsd/blob/9f0361f7e5b0bf0d2de6879c4d149991a45aab32/lib/wallet/rpc.js#L2161-L2163

I believe rpc relies on `walletpassphrase` to do `unlocking` of the master key, so you don't need to pass passphrase for each of the requests. I have not checked if that works properly, but everything still boils down to `this.unlock` call in the wallet so it should work even when passphrase is not present (when walletpassphrase unlock is in play) 

e.g. sendReveal needs to detect if masterKey is unlocked(happens in this.unlock) and not require passphrase in those cases. (in the end `this.sign` will do the check with `this.unlock`).

pinheadmz:
@nodech ah yes. Maybe it should be a different PR but I was looking at how to "send reveal all" from rpc. I think it should allow 0 parameters to do so.

nodech:
If rpc call `sendreveal` does not have `name` passed as parameter, `sendRevealAll` will be called. So no, it does not need additional params.

lib/wallet/http.js
pinheadmz: Great catch, thanks -- will take this suggestion and push

pinheadmz:
> If rpc call `sendreveal` does not have `name` passed as parameter, `sendRevealAll` will be called. So no, it does not need additional params.

@nodech That may be so, but the argument checker does not agree:

```
--> hsw-rpc sendredeem
sendredeem "name" ( "account" )
```

pinheadmz:
Ok, I fixed rpc sendredeem (all) here: https://github.com/handshake-org/hsd/pull/590/commits/961ea8d18ffe2a12285a7df2917f8cbb4d199397

I think this may actually be the first test of wallet  rpc commands for auctions (!?)

pinheadmz:
I added an extra commit here: e4ae933f7052b0c19ac9fa100001f57d51c0b7f7 which I think will fix the sometimes-failing tests.

The test sends a TX, then abandons it so the coins can be re-spent, then sends it again a different way (for maximum coverage!) _Sometimes_, the abandon is not completed by the time the re-spend TX is being funded, and so not all coins have been restored to the database. In particular, we need those unspent BID outputs back so we can spend them into a REVEAL. The occasional error is `"Could not resolve preferred inputs."`

The commit applies both wallet locks (`writeLock` and `fundLock`) during abandon. Why do we have two separate locks in the wallet? That is left up to the reader :-P

pinheadmz:
OK, the race condition is ACTUALLY fixed here: 9c6151c8cc2cb8eb3bb89b85f7ad699722fd5d39

I don't know why it took me so long to figure that out, I solved the exact same problem before in a similar test: https://github.com/handshake-org/hsd/blob/master/test/wallet-accounts-auction-test.js#L21-L25

SO I guess we could question whether or not we need https://github.com/handshake-org/hsd/commit/e4ae933f7052b0c19ac9fa100001f57d51c0b7f7 -- it didn't break anything, and it LOOKS like we should `fundlock` during `abandon` to prevent race conditions.... so I think we should keep it.

test/wallet-http-test.js
```diff
@@ -1299,6 +1300,103 @@ describe('Wallet HTTP', function() {
...
     assert.equal(ns.info.state, 'REVOKED');
   });
+
+  it('should require passphrase for auction TXs', async () => {
+    const passphrase = 'BitDNS!5353';
+    await wclient.createWallet('lockedWallet', {passphrase});
+    const lockedWallet = await wclient.wallet('lockedWallet');
+
+    // Fast-forward through the default 60-second unlock timeout
+    async function lock() {
```
nodech, line 1310: hm.. this async wont  have an effect (because you are not returning promise)
But if you want to wait for this fully you can just use normal `async/await`:

```js
  async function lock() {
    const wallet = await wdb.get('lockedWallet');
    await wallet.lock();
    // or
    return wallet.lock();
  }
 ```

pinheadmz: thanks

nodech (approved):

PR #593, opened 2021-04-14
dns: fix root server dnssec proofs

### LINKED ISSUE #513: respond with servfail if dnssec validation fails
security-aware resolvers like hsd should protect users from forged DNSSEC responses. hsd only omits the "AD" flag from the response but it should return SERVFAIL instead.

example:

`dig @1.1.1.1 dnssec-failed.org a`

this will return SERVFAIL with 1.1.1.1 but hsd will return an A record

### LINKED ISSUE #574: Fix DNSSEC denial of existence proofs
Authenticated denial of existence is a fundamental part of dnssec. It is used to validate whether a domain name or a specific RR exists. Without such proofs, dnssec wouldn't be very useful. We rely on DANE to do certificate pinning. Verifying the absence of a TLSA record is very important otherwise it would be trivial to bypass such pinning by simply removing the record from the response.

DNSSEC aware resolvers also need to verify whether a DS record exists for a domain since the absence of such record means the zone is insecure. Without authenticated denial of existence it would be easy to lie about this to downgrade secure TLDs to insecure ones. HNS resolvers get DS records from the blockchain so this is less of an issue for handshake at least but resolvers don't know about this so we need to provide proofs when there is no DS record. ICANN includes insecure delegation proofs for their domains. This is easy because they know all the TLDs beforehand, and their root zone is sorted.

Example:
```
dig @a.root-servers.net cf. ds +dnssec
cf.			86400	IN	NSEC	cfa. NS RRSIG NSEC
cf.			86400	IN	RRSIG	NSEC 8 1 86400 20210326170000 20210313160000 42351 . Ozsnfs0n......
```
This is a proof that `.cf` doesn't have a DS record. It only has `NS RRSIG NSEC` records.  Since the zone is sorted, the NSEC record points to the next domain, which is `cfa.`. This also proves that no names exist between ".cf" and ".cfa". It was designed this way to allow offline singing. 

Such proofs are not possible with HNS for two reasons. First, the number of TLDs is much larger, so it wouldn't be easy to dump the root zone, and re-sign every time a new TLD is added although some awesome work has been done in this area see [handshake-bridge](https://github.com/james-stevens/handshake-bridge).  Second, SPV nodes don't have the complete root zone, so this isn't possible. 

HNS resolvers have a workaround for this. There is an NSEC proof that makes the root zone look empty. libunbound used by hsd internally is not happy with this, so it marks all insecure TLDs as bogus. Bogus means that the domain has failed dnssec validation in some way, and it's not safe to trust the result.

We can't differentiate between bogus and insecure domains because all domains without a DS record are marked as bogus. So hsd allows serving bogus records to avoid breaking insecure domains. This isn't only an issue for insecure TLDs because the nameservers for secure TLDs can be mimted and their responses can change, so the results turn to bogus but these results are still served even for those domains (with AD bit omitted). 

TL;DR: These issues make HSD incompatible with existing dnssec tools, and we don't have proper authenticated denial of existence. 

# Solution

This solution is based on a technique used by Cloudflare called compact DNSSEC denial of existence (black lies). It is a simple technique to construct on the fly NSEC proofs by using a single NSEC resource record. Cloudflare has been using NSEC black lies since 2016 on a large scale and it has been proven to work with existing DNSSEC software (coredns uses it as the only method for denial of existence proofs).

Some resolvers like Knot (starting from version 4.1) and PowerDNS recognizes these techniques and included improvements to avoid aggressive NSEC caching for minimal NSEC ranges but the awesome thing about black lies is that resolvers work with it out of the box because it follows RFCs with only some slight differences (and they have attempted to standardize it).

Here's how it can be used with HSD root server to fix all of ours proofs and even replace the ICANN proofs to have a root zone that works!

Since we don't know what is the next TLD when constructing NSEC proofs in hsd, we use the [immediate lexicographic successor of the name](https://tools.ietf.org/html/draft-valsorda-dnsop-black-lies-00#section-2) to create a proof that only covers that name (instead of the entire root zone):

For example, to get the immediate successor of the top label `3b.` we add a null octect to get `\000.3b` as the next domain in the NSEC proof. This is important to ensure we don't accidentally prove other names don't exist we have to create the smallest range possible. You can see how NSEC sorts names in [rfc4034 section 6.1](https://tools.ietf.org/html/rfc4034#section-6.1).  

**Note:** the true immediate successor of `3b.` is `\000.3b.` ~~but for the root zone these are not accepted in the next domain name of an NSEC record.  I suspect this is because the root zone is not authoritative over these names as [mentioned here](https://tools.ietf.org/html/rfc3845#section-2.1.1)~~ (hsd incorrectly compresses the next domain field which was causing this issue) 

Update: If the name has an NS record, the successor should be `3b\000.` because we no longer have authority over the name. This doesn't apply for proving a name doesn't exist because we are "authoritative" (we have to send a NODATA proof for any label underneath) 

## Proving a record doesn't exist 

The following is used for insecure delegation proofs

```
nb.			10800	IN	NSEC	nb\000. NS RRSIG NSEC
nb.			10800	IN	RRSIG	NSEC 13 1 10800 20220313210740 20210312210740 60944 . dovH4c88gX....
```

Meaning: There are only NS NSEC and RRSIG records for `nb.` and the next domain is `\000.nb.` 

## Proving a name doesn't exist

Black lies uses NODATA instead of NXDOMAIN (see [section 3](https://tools.ietf.org/html/draft-valsorda-dnsop-black-lies-00#section-3)) to provide such proofs. Example:

```
howtomakepancakes.	10800	IN	NSEC	\000.howtomakepancakes. RRSIG NSEC
howtomakepancakes.	10800	IN	RRSIG	NSEC 13 1 10800 20220313211044 20210312211044 60944 . MMDt665...
```

Meaning: The tld `howtomakepancakes.` only has NSEC and RRSIG records. This proves there are no other record types which effectively makes it "NXDOMAIN". 
  
## Proof of concept implementation

This branch implements NSEC black lies for hsd and fixes icann and hsd proofs. 

https://github.com/buffrr/hsd/tree/blacklies

This fixes issue #513  but the javascript/unbound bindings have to be patched to add a check for bogus. 

Also, you can now run hsd as a standalone root server! and create a root hints file that can be plugged into bind, unbound or any external dnssec resolver using HSD's KSK (we should probably not hard code these). 

The external recursive resolver will do the heavy lifting for you! In this setup, hsd will only be responsible for serving the root zone. Using an external resolver like bind resolves issue #513 because bind takes care of it. See [examples here](https://gist.github.com/buffrr/546a14ebab606f8b602a9183f28b9a3a) for how to run hsd root server with other resolvers.

References:
https://tools.ietf.org/html/draft-valsorda-dnsop-black-lies-00
https://blog.cloudflare.com/black-lies/

buffrr:
This PR uses epsilon functions optimized for handshake root zone based on [RFC4470](https://tools.ietf.org/html/rfc4470) to find minimally covering NSEC records. 

Closes #574 and closes #513 (a minor change is needed in bns to check for bogus but shouldn't require other changes in hsd)

One issue with compact NSEC proofs is that due to qname minimization a compact proof may require multiple NODATA lookups `test.example.doesntexist` may need three lookups. The change to these proofs was adding a wildcard NSEC record to get NXDOMAIN responses. The response is a tiny bit larger (only for NXDOMAIN case) than compact proofs (~400 bytes vs. ~524 bytes for average length names) but covers all names underneath the TLD limits the number of proofs that need to be signed.

This PR fixes dnssec denial of existence in hsd root server (which is important for DANE) and several other inconsistencies related to how DS and referrals are handled necessary to make NSEC/DS answers valid. It makes hsd root server compatible with external recursive resolvers like Bind, Unbound, Knot resolver ... etc without having to disable DNSSEC or ignore bogus errors in libunbound used by hsd.

### Summary of changes:

* Replaces empty zone proofs with minimally covering NSEC records
* Fix ICANN insecure delegation proofs 
* No referrals for negative DS answers since root zone is authoritative (referral answers must only be for delegated sub-trees).
* Fix ICANN DS answers since it used to send referrals for all questions
* NSEC bitmap now shows correct types available for a name like TXT records (if no delegation)
* Clean up SOA and AA bit handling for handshake and ICANN names
* Remove DS from root zone it should only exist in parent zone 

I added tests that have examples of these changes and should cover various cases.

#### Edge cases

For weird queries like `dig @127.0.0.1 -p 5349 example\\000test a` this PR sends REFUSED.
this simplifies the predecessor/successor functions and avoids signing useless NXDOMAIN proofs for invalid names.

I was curious how Cloudflare deals with this and they avoid creating NSEC records for weird names. They send an NXDOMAIN response without signing it resulting in SERVFAIL `dig @8.8.8.8 example\\000test.cloudflare.com a` but I think it's better to send a REFUSED answer than an invalid answer. They do this because they have a special handling for this case in their 1.1.1.1 resolver

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ when users need to call `rpc importnonce` to repair unknown blinds. The complete
 - Wallet RPC `getnames` (and HTTP endpoint `/wallet/:id/name`) now accept a
 boolean parameter "own" (default: `false`) that filters out names the wallet does not own.
 
+### DNS changes
+
+- DNSSEC proofs from the root name server were fixed, particularly around non-existent
+domains. The empty zone proofs were replaced with minimally covering NSEC records.
+
 ### Other changes
 
 - The logging module `blgr` has been updated. Log files will now be rolled over
--- a/bench/nsec.js
+++ b/bench/nsec.js
@@ -1,0 +1,16 @@
+'use strict';
+
+const bench = require('./bench');
+const nsec = require('../lib/dns/nsec');
+const {TYPE_MAP_EMPTY} = require('../lib/dns/common');
+
+const tld = 'dollarydoo';
+const ops = 100000;
+
+const end = bench('nsec');
+for (let i = 0; i < ops; i++) {
+  const prev = nsec.prevName(tld);
+  const next = nsec.nextName(tld);
+  nsec.create(prev, next, TYPE_MAP_EMPTY);
+}
+end(ops);
--- a/lib/dns/common.js
+++ b/lib/dns/common.js
@@ -1,0 +1,50 @@
+/*!
+ * common.js - dns constants for hsd
+ * Copyright (c) 2021, The Handshake Developers (MIT License).
+ * https://github.com/handshake-org/hsd
+ */
+
+'use strict';
+
+/**
+ * @module dns/common
+ */
+
+exports.DUMMY = Buffer.alloc(0);
+
+// About one mainnet Urkel Tree interval.
+// (60 seconds * 10 minutes * 36)
+exports.DEFAULT_TTL = 21600;
+
+// NS SOA RRSIG NSEC DNSKEY
+// Types available for the root "."
+exports.TYPE_MAP_ROOT = Buffer.from('000722000000000380', 'hex');
+
+// RRSIG NSEC
+exports.TYPE_MAP_EMPTY =  Buffer.from('0006000000000003', 'hex');
+
+// NS RRSIG NSEC
+exports.TYPE_MAP_NS = Buffer.from('0006200000000003', 'hex');
+
+// TXT RRSIG NSEC
+exports.TYPE_MAP_TXT = Buffer.from('0006000080000003', 'hex');
+
+exports.hsTypes = {
+  DS: 0,
+  NS: 1,
+  GLUE4: 2,
+  GLUE6: 3,
+  SYNTH4: 4,
+  SYNTH6: 5,
+  TXT: 6
+};
+
+exports.hsTypesByVal = {
+  [exports.hsTypes.DS]: 'DS',
+  [exports.hsTypes.NS]: 'NS',
+  [exports.hsTypes.GLUE4]: 'GLUE4',
+  [exports.hsTypes.GLUE6]: 'GLUE6',
+  [exports.hsTypes.SYNTH4]: 'SYNTH4',
+  [exports.hsTypes.SYNTH6]: 'SYNTH6',
+  [exports.hsTypes.TXT]: 'TXT'
+};
--- a/lib/dns/nsec.js
+++ b/lib/dns/nsec.js
@@ -27,9 +27,9 @@ function nextName(tld) {
   // increment last character by one
   if (tld.length === 63) {
     // Assuming no escaped octets are present
-    const next = Buffer.from(tld, 'ascii');
-    next[next.length-1]++;
-    return next.toString() + '.';
+    let last = tld.charCodeAt(62);
+    last = String.fromCharCode(last + 1);
+    return tld.slice(0, -1) + last + '.';
   }
 
   return tld + '\\000.';
@@ -42,9 +42,9 @@ function prevName(tld) {
 
   // Decrement the last character by 1
   // assuming no escaped octets are present
-  let prev = Buffer.from(tld, 'ascii');
-  prev[prev.length-1]--;
-  prev = prev.toString();
+  let last = tld.charCodeAt(tld.length - 1);
+  last = String.fromCharCode(last - 1);
+  tld = tld.slice(0, -1) + last;
 
   // See RFC4034 6.1 Canonical DNS Name Order
   // https://tools.ietf.org/html/rfc4034#section-6.1
@@ -54,11 +54,11 @@ function prevName(tld) {
   // smaller name is `helln` append `\255`
   // to ensure that helln\255 > hellna
   // while keeping helln\255 < hello
-  if (prev.length < 63) {
-    prev += '\\255';
+  if (tld.length < 63) {
+    tld += '\\255';
   }
 
-  return util.fqdn(prev);
+  return util.fqdn(tld);
 }
 
 exports.create = create;
--- a/lib/dns/nsec.js
+++ b/lib/dns/nsec.js
@@ -1,0 +1,82 @@
+'use strict';
+
+const assert = require('bsert');
+const {wire, util} = require('bns');
+const {Record, NSECRecord, types} = wire;
+
+// NS SOA RRSIG NSEC DNSKEY
+// Types available for the root "."
+const TYPE_MAP_ROOT = Buffer.from('000722000000000380', 'hex');
+
+// RRSIG NSEC
+const TYPE_MAP_EMPTY =  Buffer.from('0006000000000003', 'hex');
+
+// NS RRSIG NSEC
+const TYPE_MAP_NS = Buffer.from('0006200000000003', 'hex');
+
+// TXT RRSIG NSEC
+const TYPE_MAP_TXT = Buffer.from('0006000080000003', 'hex');
+
+function create(name, nextDomain, typeBitmap) {
+  const rr = new Record();
+  const rd = new NSECRecord();
+  rr.name = util.fqdn(name);
+  rr.type = types.NSEC;
+  rr.ttl = 86400;
+
+  rd.nextDomain = util.fqdn(nextDomain);
+  rd.typeBitmap = typeBitmap;
+  rr.data = rd;
+
+  return rr;
+}
+
+// find the successor of a top level name
+function nextName(tld) {
+  tld = util.trimFQDN(tld.toLowerCase());
+
+  // if the label is already 63 octets
+  // increment last character by one
+  if(tld.length === 63) {
+    // assuming no escaped octets are present
+    const next = Buffer.from(tld, 'ascii');
+    next[next.length-1]++;
+    return next.toString() + '.';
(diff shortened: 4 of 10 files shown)


pinheadmz:
#633 has been merged with your improvement to bns, you should rebase this PR on master branch

pinheadmz:
@buffrr pushed a commit instead of leaving a bunch of annoying review comments.

- fixed style nits like spacing etc
- moved all constants to new file `common.js` this is to avoid circular dependencies with the NSEC bitmaps you added, and also sharing of default TTL in several dns modules
- added a test confirming the NSEC bitmaps
- updated changelog!

lmk what you think, otherwise I really like these changes, great work!

pinheadmz:
Also, I think this PR does not close #513 (anymore) - is that right?

buffrr:
> Also, I think this PR does not close #513 (anymore) - is that right?

This PR should fix what's needed in hsd but bns still requires the check for bogus to get fixed. We can keep it open until bns is updated

pinheadmz:
I was able to confirm the effectiveness of this PR by running hsd without its Javascript-wrapped unbound recursive resolver:

```
sudo hsd --no-rs --log-level=spam --ns-port 53
```

...then running unbound in a separate process with trust anchor file and root hints file pointing back at hsd's root ns (localhost) and DS:

*unbound.conf*
```
server:
        port: 50000
        username: matthewzipkin
        do-daemonize: no
        use-syslog: no
        chroot: ""
        trust-anchor-file: /usr/local/etc/unbound/trust.anchor
        root-hints: /usr/local/etc/unbound/root.hints
        verbosity: 1
        do-not-query-localhost: no
```

*root.hints*
```
. 3600000 IN NS .
. 3600000 IN A 127.0.0.1
```

*trust.anchor*
```
. DS 35215 13 2 7C50EA94A63AEECB65B510D1EAC1846C973A89D4AB292287D5A4D715136B57A3
```

Then I could switch between master branch and PR branch to test how unbound **REALLY** resolves insecure domains:

*master*
```
--> dig @127.0.0.1 -p 50000 nb

; <<>> DiG 9.16.10 <<>> @127.0.0.1 -p 50000 nb
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 55217
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;nb.                            IN      A

;; Query time: 603 msec
;; SERVER: 127.0.0.1#50000(127.0.0.1)
;; WHEN: Sat Sep 18 15:37:56 EDT 2021
;; MSG SIZE  rcvd: 31
```

*pull request*
```
--> dig @127.0.0.1 -p 50000 nb

; <<>> DiG 9.16.10 <<>> @127.0.0.1 -p 50000 nb
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 10991
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;nb.                            IN      A

;; ANSWER SECTION:
nb.                     56      IN      A       52.43.158.89

;; Query time: 109 msec
;; SERVER: 127.0.0.1#50000(127.0.0.1)
;; WHEN: Sat Sep 18 15:41:35 EDT 2021
;; MSG SIZE  rcvd: 47
```

pinheadmz (approved):
ACK 4b96d4b286179ff3a7f1dd2c7f675924d9a8299f

```
Hash: SHA256

ACK 4b96d4b286179ff3a7f1dd2c7f675924d9a8299f
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmFGQ+0ACgkQ5+KYS2KJ
yTqWNRAAtcBiil0aNM8XY3/vYbNmvhnIg2uH4pl04hKr0wNCZAUxO20btSmVsuMP
goCNXOBYcDB6DpF7FBllVVNuJxitxEEIE7/LxIZZynTErQRTekBZ+XvApbeBN7+t
Z3Aqs6hrXrvNgeiISYA7uhVvo/4w71YqK5AqH2SJlRrpMv1GZq5AQ/qWQY6llHbX
Q1mm4+Agp20i65sM0iLyQ1XJClDraoi90obLJiCmradmPxy6Zavp2nIJwBAMI8ms
EWE0ZoGwtu1nJYTdJLwhj2Aa3FBDMS/CtDS0k11Zaryn7XNxbv7FpcYnGJnI6Ool
zEDDzpc6nKxu22Dvwa+vSIzQ2iM/HBH0hIehIy8sheY13mSn1RS5RcLLS/EutpNC
83u0Hl+oh7RRLjP7wlAN5X8Xh2sVOF/A9PPqhEBvUYe1bxpiq01t/Qz/r/aNn7i5
mVWnuCktPTj3D+supswDl0cA8XNb99ykcwcTya3kNmUI97JLeLs5gmhIUqcRS/DT
51CoXyU4blsMFRI9DJCKGWiTbqmVQcbbdK+ec9+Kcl5EoktHqmgiHaHsgGoFVyB/
RRDvkfv2r0yK2sj80nId//34vF500gonMDrS5k6wzTq0hYLdc3Q/9f/B+srkLzh+
su81LwK54TampZMzjEP+VsX4t/8pvrnA2uH3yCfLv4gwLOPEKyg=
=VBOa
-----END PGP SIGNATURE-----
```

</details>

PR #594, opened 2021-04-14
wallet-http: do not return until deepclean is done

pinheadmz:
This swaps 2 lines in wallet/http for `deepclean`, so the HTTP response does not return until walletDB is done with deep clean. It should be a relatively fast operation for most wallets (it doesn't have a built-in rescan any more). Worst case is the http request times out, but I think thats preferable to no one ever knowing when deep clean is done.

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -215,9 +215,9 @@ class HTTP extends Server {
         'Deep Clean requires I_HAVE_BACKED_UP_MY_WALLET=true'
       );
 
-      res.json(200, { success: true });
-
       await this.wdb.deepClean();
+
+      res.json(200, { success: true });
     });
 
     // Resend


nodech (approved):
yeah, probably having some wallet db lock that can be handled by the API in general would be good for big operations like this. So you can check walletdb status and check the state. When you get timeout, you have no clue how long you need to wait for it to finish. Even though "In general" it should not be the issue.

This issue also applies to `rescan/scan`, `backup` and migrations (each of them behaves the same way as this PR)

This makes the issue little bit better for small wallets, so approved.

nodech:
We could add event that is also passed to sockets, that way client will have a way to wait properly.

pinheadmz:
> This makes the issue little bit better for small wallets, so approved.

Ok thanks! Will open issue for the remainder, good idea.

PR #599, opened 2021-05-17
net: restore easyhandshake dns seed

### LINKED ISSUE #540: Review the DNS seeds for mainnet
https://github.com/handshake-org/hsd/blob/c85d9b4c743a9e1c9577d840e1bd20dee33473d3/lib/protocol/networks.js#L51-L54

It seems like `hs-mainnet.bcoin.ninja` is not resolving.
While `seed.easyhandshake.com` resolves

pinheadmz:
Closes #540 

The easyhandshake DNS seed was disabled over a year ago in bbac8dd943767da62e46e52f93227b3da1048e17 when the network protocol version was bumbed (easyhandshake stayed on the old code to act like a "bridge" if the network partitioned). I think it's safe to put it back.

Reviewers can test by:
- `rm ~/.hsd/hosts.json`
- clear out the contents of the array in https://github.com/handshake-org/hsd/blob/master/lib/net/seeds/main.js
- start `hsd`

These log messages should be expected:

```
[warning] (net) Could not find enough peers.
[warning] (net) Hitting DNS seeds...
[info] (hostlist) Resolving host: hs-mainnet.bcoin.ninja.
[info] (hostlist) Resolving host: seed.easyhandshake.com.
...
[info] (net) Resolved 24 hosts from DNS seeds.
```

...followed by a healthy peer count and network connection.

Issue #540 also mentions signing the results with DNSSEC. I think that is a cool idea and could be added in a future PR to /lib/node/seeder.js -- However, I am upset that my hosting provider (Namecheap) doesn't allow me to insert a DS record in my zone `easyhandshake.com`. I can add an NS record (that's how the seeder works) and I can set Namecheap to sign my zone with DNSSEC, but there is no way to continue a secure chain of delegation! Bummer. Stay tuned on that one...

### DIFF
--- a/lib/protocol/networks.js
+++ b/lib/protocol/networks.js
@@ -49,8 +49,8 @@ main.type = 'main';
  */
 
 main.seeds = [
-  'hs-mainnet.bcoin.ninja' // Christopher Jeffrey
-  // 'seed.easyhandshake.com' // Matthew Zipkin
+  'hs-mainnet.bcoin.ninja', // Christopher Jeffrey
+  'seed.easyhandshake.com' // Matthew Zipkin
 ];
 
 /**


pinheadmz:
failing test fixed here: https://github.com/handshake-org/hsd/pull/600

Anunayj (approved):
Seems to work with both and either of them alone.
Things that would be nice: 
1. IPv6 support
2. DNSSEC (Not Supported by Matthew's Registrar yet as he mentions)

pinheadmz:
@Anunayj Thanks for reviewing - I tried to connect to the ipv6 peers that were gossiped to my node, but none connected. Note that ipv6 is supported, but you will not get an answer for `AAAA` if the seed node has not successfully connected to an ipv6 peer:

https://github.com/handshake-org/hsd/blob/bedcb3b87a3c26e0c743effc333cd78fb34b0486/lib/node/seeder.js#L137-L156

PR #601, opened 2021-05-18
Update documented reference for Git clone to https: URI.

aef:
The git URI scheme usually implies the unsecured native Git protocol
which is vulnerable to man-in-the-middle attacks. Using TLS based HTTP
should be an obvious security improvement.


pinheadmz:
great, thanks

Issue #606, opened 2021-06-01
[bounties]: fix signmessage + add rpc {sign, verify}messagewithname + API docs + Bob Wallet UI integration


pinheadmz:
Background: hsd has rpc methods to sign and verify messages with the public key of a specified address:

https://hsd-dev.org/api-docs/#signmessage
https://hsd-dev.org/api-docs/#verifymessage

You can try these out with this example: https://hns-contributor.dev/sigs.txt

# Issues

## 1) signmessage should only work for public-key-hash addresses

You can create a 2-of-2 multisig wallet and it will signmessage with (I assume) the wallet's single key:

```
$ hsw-rpc signmessage rs1q6rclu20hhzprprm2kaj0u2my3eunfzsxr624fwr600xas4zvkvds4xyy0j "hello"
ecVXQQTQIzjdv8NIOa6K5X3+r/D2umiDn777kKJAzKpkrbf3V+Ex33NSlM3/zlM0LOuvnTwsBIw84vMdJrCirg==
```

This should throw an error because this is a script hash address and normally spending from this address would require two signatures! In fact, the `verifymessage` rpc quickly invalidates such attempts:

https://github.com/handshake-org/hsd/blob/c861e4fc95f0d023a7da4c90b93dab692b9209d0/lib/node/rpc.js#L2180-L2181

Solution: `rpc signmessage` should match this condition and throw an error otherwise 

## 2) Create new RPCs `signmessagewithname` and `verifymessagewithname`

These should work just like the `signmessage` and `verifymessage` counterparts but begin by pulling the namestate from urkel, then retrieving the coin from chainDB, then using the `coin.address` for the rest. It should also fail if the address is not a version 0, pubkeyhash address.

## 4) Add {sign, verify} message with {address, name} in Bob Wallet GUI

# Bounties

I am willing to pay bounties for this. I am denominating these bounties in US dollars but they will be paid in HNS. I reserve the right to revoke this offer at any time. Bounties will only paid if a PR is merged. If more than one developer contributes to a PR I will divide the bounty. 

- All pull requests must always include good tests.
- These issues MUST be solved in the correct order!

| issue | bounty | completed? |
|-|-|-|
| 1) ~fix rpc sign message~ | $100  | ✅  |
| 2) ~add {sign, verify}messagewithname~ | $100 | ✅  |
| 3) ~document new rpc~ | $50 | ✅  |
| 4) ~add to bob~ | $100| ✅  |

[...]

PR #608, opened 2021-06-06
signmessage rejects multisig wallet

Falci:
#606 item 1: signmessage should only work for public-key-hash addresses.

Note: I'm reusing the `INVALID_ADDRESS_OR_KEY`, not sure if it would be better to have its own error code.

PR #609, opened 2021-06-06
Sign & verify message with name

Falci:
#606 item 2: Create new RPCs `signmessagewithname` and `verifymessagewithname`

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2227,7 +2227,7 @@ class RPC extends RPCBase {
     }
 
     const address = coin.address.toString(this.network);
-    return this.verifyMessage([address, sig, str], help);
+    return this.verifyMessage([address, sig, str]);
   }
 
   async signMessageWithPrivkey(args, help) {
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2208,6 +2208,9 @@ class RPC extends RPCBase {
     const sig = valid.buf(1, null, 'base64');
     const str = valid.str(2);
 
+    if (!name || !rules.verifyName(name))
+      throw new RPCError(errs.TYPE_ERROR, 'Invalid name.');
+
     const nameHash = rules.hashName(name);
     const ns = await this.chain.db.getNameState(nameHash);
 
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -207,6 +207,7 @@ class RPC extends RPCBase {
     this.add('validateaddress', this.validateAddress);
     this.add('createmultisig', this.createMultisig);
     this.add('verifymessage', this.verifyMessage);
+    this.add('verifymessagewithname', this.verifyMessageWithName);
     this.add('signmessagewithprivkey', this.signMessageWithPrivkey);
 
     this.add('setmocktime', this.setMockTime);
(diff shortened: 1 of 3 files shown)


test/wallet-rpc-test.js
```diff
@@ -313,4 +314,93 @@ describe('Wallet RPC Methods', function() {
       assert.strictEqual(info.height, node.chain.height);
     });
   });
+
+  describe('signmessagewithname', () => {
+    const name = rules.grindName(5, 1, network);
+    const nonWalletName = rules.grindName(5, 1, network);
+    const message = 'Decentralized naming and certificate authority';
```
pinheadmz, line 321: nit: missing word "Experimental" :-)

Just kidding.

test/wallet-rpc-test.js
```diff
@@ -313,4 +314,93 @@ describe('Wallet RPC Methods', function() {
...
+
+      const verify = await nclient.execute('verifymessagewithname', [
+        name,
+        signature,
+        message
+      ]);
+
+      assert.strictEqual(verify, true);
+    });
+  });
```
pinheadmz, line 405: This test actually doesn't add anything because the wallet does not verify the message at all, it's a full node rpc call (notice `nclient` instead of `wclient` to verify). I know what you're trying to do here, technically you'd want to sync a second `FullNode` from the original node and then test on that second node, but I don't think you need to test that whole integration.

To get more coverage, you might want to instead test verifying a name that is invalid or doesn't exist.

After #608 gets merged you might even want to acquire a name with a multisig wallet and make sure sign with name fails like it's supposed to.

pinheadmz (changes requested):
Great job on this dude! Very cool and fun to play with on command line as well:

```
--> hsw-rpc signmessagewithname test-txt "Falci is handsome."
CUjgakQfpOG9aTa4DzKaX9heuYZW2D26tZ9pb6VEgco5pLNuU8NljSQn2DtAFGkpEiEugNYNh94W7JZAjKuWkQ==
--> hsd-rpc verifymessagewithname test-txt CUjgakQfpOG9aTa4DzKaX9heuYZW2D26tZ9pb6VEgco5pLNuU8NljSQn2DtAFGkpEiEugNYNh94W7JZAjKuWkQ== "Falci is handsome."
true
--> hsd-rpc verifymessagewithname test-txt CUjgakQfpOG9aTa4DzKaX9heuYZW2D26tZ9pb6VEgco5pLNuU8NljSQn2DtAFGkpEiEugNYNh94W7JZAjKuWkQ== "Falci is not handsome."

false
```

pinheadmz:
Oh one more thing: please add these new RPC methods to the CHANGELOG!

pinheadmz:
As discussed, both sign and verify should check if name is `null`:

```
if (!name || !rules.verifyName(name))
      throw new RPCError(errs.TYPE_ERROR, 'Invalid name.');
```

test/wallet-rpc-test.js
```diff
@@ -313,4 +314,102 @@ describe('Wallet RPC Methods', function() {
...
+          message
+        ]);
+      }, {
+        type: 'RPCError',
+        message: 'Cannot find the name owner.'
+      });
+    });
+
+    ['', null, '\'null\'', 'localhost'].forEach((invalidName) => {
+      it('should fail to verify with invalid name.', async () => {
```
pinheadmz, line 385:
```suggestion
      it('should fail to sign with invalid name.', async () => {
```

test/wallet-rpc-test.js
```diff
@@ -313,4 +314,102 @@ describe('Wallet RPC Methods', function() {
...
+          nonWalletName,
+          message
+        ]);
+      }, {
+        type: 'RPCError',
+        message: 'Cannot find the name owner.'
+      });
+    });
+
+    ['', null, '\'null\'', 'localhost'].forEach((invalidName) => {
```
pinheadmz, line 384: Since you're not trying to include each `invalidName` in the title of the test, I think you can move the array and the loop (`for const...`) to the inside of the test block. You can also set the `const invalidNames = [...]` at the top of the `describe` section so you can just grab it from both tests.

pinheadmz (changes requested):

lib/node/rpc.js
```diff
@@ -2196,6 +2197,39 @@ class RPC extends RPCBase {
...
+
+    if (!coin) {
+      throw new RPCError(
+        errs.DATABASE_ERROR,
+        'Cannot find the owner\'s address.'
+      );
+    }
+
+    const address = coin.address.toString(this.network);
+    return this.verifyMessage([address, sig, str], help);
```
pinheadmz, line 2230: I don't think you need to pass `help` here. If help were `true`, we'd throw an error right away, and we don't want to return help from a different RPC call anyway. Applies to both functions.

lib/wallet/rpc.js
```diff
@@ -1586,6 +1587,40 @@ class RPC extends RPCBase {
...
+    const wallet = this.wallet;
+    const valid = new Validator(args);
+    const name = valid.str(0, '');
+    const str = valid.str(1, '');
+
+    if (!name || !rules.verifyName(name))
+      throw new RPCError(errs.TYPE_ERROR, 'Invalid name.');
+
+    if (!wallet.master.key)
+      throw new RPCError(errs.WALLET_UNLOCK_NEEDED, 'Wallet is locked.');
```
pinheadmz, line 1605: I don't think we need this here, it will be checked in `signMessage()` and we don't need to decrypt the wallet until then anyway.

pinheadmz (changes requested):

0racl3z (commented):
thx!

Issue #598, opened 2021-05-15
Improving the RPC help


gdassori:
Hello.
I have seen that both the hsw-cli and hsd-cli rpc interfaces are not documented:

```
$ ./hsd-cli rpc help
Select a command.
```

And fails badly when called with no arguments:

```
$ ./hsw-cli rpc
AssertionError [ERR_ASSERTION]: Assertion failed.
```

I would love to contribute, and I thought that addressing those issues could be a good start.
However, I see there's the API unification issue (#543), and could be the reason why no one looked at this behaviours yet.

What's your takes?

pinheadmz:
I think this has been mentioned in other issues, we have a solution upstream in bcoin that just needs to be ported over to hsd: https://github.com/bcoin-org/bcoin/pull/705

There were recent updates to hs-client that have not been released yet in a new npm version but are merged into master:

https://github.com/handshake-org/hs-client/pull/38

https://github.com/handshake-org/hs-client/pull/35

The assertion error you have here is a good catch, that's not good UX, and that should be fixed in the https://github.com/handshake-org/hs-client repo, if you feel like opening a new issue or submitting a pull request there!

gdassori:
Ah, I see, sorry for the duplicate and thanks for the prompt response, better to check out the master instead of the last release before heading to the issue tracker :)
So see you on the client repository for the assertion error, I close this one.

pinheadmz:
Thanks for submitting! I'm actually going to re-open this to remind us to port that bcoin rpc help thing over -- its also something you could do if you feel like contributing!

PR #610, opened 2021-06-10
Chain Prune test

nodech:
Add test for chain prune options and pruning itself.
Add node rpc test for pruneblockchain.

Renamed `bip9-chain-test` to `chain-bip9-test`.


test/node-rpc-test.js
```diff
@@ -225,4 +239,148 @@ describe('RPC', function() {
...
+      nclient = new NodeClient(clientOptions);
+
+      await node.open();
+
+      await assert.rejects(async () => {
+        await nclient.execute('pruneblockchain', [1]);
+      }, {
+        code: errs.MISC_ERROR,
+        type: 'RPCError',
+        message: 'pruneblockchain'
```
pinheadmz, line 279: At first I was like "what?" then I realized this is a help/usage message, ugh terrible. We need to fix all these: https://github.com/handshake-org/hsd/issues/598

test/node-rpc-test.js
```diff
@@ -225,4 +239,148 @@ describe('RPC', function() {
...
+      });
+
+      await node.close();
+    });
+
+    it('should fail for short chain', async () => {
+      node = new FullNode(nodeOptions);
+
+      await node.open();
+
```
pinheadmz, line 325: Mine as well include assertion here that the chain is short -- a future dev may try adding blocks to the chain in an earlier test. Alternatively, you could add a `beforeEach()` or something that guarantees each `it` block starts with a fresh chain .

test/node-rpc-test.js
```diff
@@ -225,4 +239,148 @@ describe('RPC', function() {
...
+      // make sure we have all the blocks.
+      for (let i = 0; i < 30; i++) {
+        const block = await nclient.execute('getblock', [blocks[i]]);
+        assert(block);
+      }
+
+      // now prune..
+      await nclient.execute('pruneblockchain');
+
+      // behined height check
```
pinheadmz, line 360:
```suggestion
      // behind height check
```

test/node-rpc-test.js
```diff
@@ -225,4 +239,148 @@ describe('RPC', function() {
...
+      for (let i = 0; i < 10; i++) {
```
pinheadmz, line 361: for these tests instead of using `10` you might want to use the constants so future reviewers understand what the number represents (`PRUNE_AFTER_HEIGHT` and `KEEP_BLOCKS`)

test/chain-prune-test.js
```diff
@@ -0,0 +1,201 @@
...
+        await chain.open();
+      }, {
+        message: 'Cannot retroactively unprune.'
+      });
+      await chain.close();
+    });
+
+    it('should force retroactive prune', async () => {
+      const chain = new Chain({
+        ...chainOptions
```
pinheadmz, line 57: nit: `...` isn't needed here ?

test/chain-prune-test.js
```diff
@@ -0,0 +1,201 @@
...
+
+    afterEach(async () => {
+      network.block.pruneAfterHeight = PRUNE_AFTER_HEIGHT;
+      network.block.keepBlocks = KEEP_BLOCKS;
+
+      if (chain.opened)
+        await chain.close();
+
+      await miner.close();
+      await workers.open();
```
pinheadmz, line 118: should this be `workers.close()`? (this is the `afterEach`)

test/chain-prune-test.js
```diff
@@ -0,0 +1,201 @@
...
+    });
+
+    it('should prune blocks', async () => {
+      chain.options.prune = true;
+
+      await chain.open();
+
+      const hashes = [];
+
+      // 10 behined height check + 10 pruned + 10 keep blocks
```
pinheadmz, line 128:
```suggestion
      // 10 behind height check + 10 pruned + 10 keep blocks
```

test/chain-prune-test.js
```diff
@@ -0,0 +1,201 @@
...
+      for (let i = 0; i < 30; i++) {
+        const block = await cpu.mineBlock();
+        hashes.push(block.hash());
+        assert(block);
+        assert(await chain.add(block));
+      }
+
+      // behined height check
```
pinheadmz, line 136:
```suggestion
      // behind height check
```

test/chain-prune-test.js
```diff
@@ -0,0 +1,201 @@
...
+        const block = await chain.getBlock(hashes[i]);
+        assert(block);
+      }
+
+      await chain.close();
+
+      chain.options.forceFlags = true;
+      chain.options.prune = true;
+
+      // this should call prune on open.
```
pinheadmz, line 178: wanna assert this?

test/node-rpc-test.js
pinheadmz: ☝️ this goes for the chain test too

test/node-rpc-test.js
nodech: Yep, I agree. Maybe this just replicated btc behaviour in the past ?

test/node-rpc-test.js
nodech: There's is new FullNode in each test. These are `unitTest`s in a way.
`afterEach` close is there only if test fails and `await node.close` is not executed in the end of the tests.

test/node-rpc-test.js
nodech: hehe, apparently I don't know how to spell behind xD

test/chain-prune-test.js
nodech: yep !!!
leftover from testing options

test/chain-prune-test.js
nodech: yes it should, nice catch.

test/chain-prune-test.js
nodech: option test for this one is above in `Prune options` section so decided to skip it here. Should we ?

nodech: Check: `should force retroactive prune`

test/node-rpc-test.js
pinheadmz: oh right, you got `new FullNode()` in each `it` nice.

test/chain-prune-test.js
pinheadmz: ok good enough for me.

PR #592, opened 2021-04-12
workers: use fork() instead of spawn()

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/423
Additional discussion: https://github.com/kyokan/bob-wallet/pull/123

This PR replaces the method we use to spawn workers from [`child_process.spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) to [`child_process.fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options). The motivation for this is Bob Wallet. Currently Bob [does not use workers](https://github.com/kyokan/bob-wallet/blob/master/app/background/node/service.js#L89) because the executable command in Bob is not nodejs, its electron. So if we tried to use workers in Bob right now, we get a bunch of new electron windows and the inter-thread communication is broken. By replacing `spawn()` with `fork()` we are guaranteed to spawn nodejs instances even in the electron environment.

Since workers are all about performance, review of this PR will require a good deal of benchmarking.

Testing involves running different configurations of hsd on the same machine in ideally the same environment (no other processes running on same machine, same level of network access, no one watching Netflix on your wifi suddenly in the middle of a run, etc), syncing to 100% and then parsing the output log: 

[I have a log-parsing script available here.](https://gist.github.com/pinheadmz/d9550744caafee9b53d9aaba74574236)

Here are the results I obtained yesterday in a few trials on my mac book pro, 4x CPU:

---

9f0361f7e default hsd master branch: `hsd --index-tx --index-address`
```
Height: 62000 Total block processing time: 3983637.2504359884 (1.1065659028988857 hours)
Total time elapsed: 4092000ms (1.1366666666666667 hours)
```

9f0361f7e hsd master branch, no workers: `hsd --workers=false --index-tx --index-address`
```
Height: 62000 Total block processing time: 3753916.287935985 (1.0427545244266623 hours)
Total time elapsed: 3804000ms (1.0566666666666666 hours)
```

00c78c488 this PR: `hsd --index-tx --index-address`
```
Height: 62000 Total block processing time: 4989106.028164006 (1.3858627856011128 hours)
Total time elapsed: 5045000ms (1.4013888888888888 hours)
```

https://github.com/kyokan/bob-wallet/commit/ff76917 Bob Wallet, master branch
```
Height: 62000 Total block processing time: 13219661.29872208 (3.6721281385339117 hours)
Total time elapsed: 13355000ms (3.7097222222222226 hours)
```

[...]

PR #612, opened 2021-06-11
Blockstore

nodech:
This pull request adds a flat file storage backend for Handshake blocks and undo coins.

This is based on https://github.com/bcoin-org/bcoin/pull/703 and continuation of https://github.com/handshake-org/hsd/pull/469

TODOs:
 - [x] Migration script
 - [x] Benchmarking
 - [x] Batch API for better chaindb integration
 - [x] Add extended merkle test vector


pinheadmz:
Excited about this!

Ran my log-parse script after a full mainnet sync: `hsd --no-wallet`
```
Height: 62000 Total block processing time: 1869648.0351210015 (0.5193466764225004 hours)
Total time elapsed: 2026000ms (0.5627777777777777 hours)
```

Which is actually a tad slower than the previous test I did on master branch which also had both indexers on: https://github.com/handshake-org/hsd/pull/592#issuecomment-822077141

Probably I'm just not testing the right thing here.

pinheadmz:
We might want to try and maintain browser support as well: https://github.com/bcoin-org/bcoin/pull/757

I think a browser full- or pruned- node is kinda silly, but JJ has made extensive effort to maintain browser compatibility, and we don't have to break it here.

lib/blockstore/file.js
```diff
@@ -0,0 +1,789 @@
...
+    const blockrecord = new BlockRecord({
+      file: fileno,
+      position: bposition,
+      length: blength
+    });
+
+    b.put(layout.b.encode(type, hash), blockrecord.encode());
+    b.put(layout.f.encode(type, fileno), filerecord.encode());
+
+    const last = bio.write(4).writeU32(fileno).render();
```
pinheadmz, line 439: Could this be optimized with `Buffer.alloc(4).writeUInt32BE(fileno)` ? I suppose we assume that `bio` has already allocated memory and the `render()` function has negligible overhead?

lib/blockchain/chaindb.js
```diff
@@ -1868,9 +1885,8 @@ class ChainDB {
     if (this.options.spv)
       return undefined;
 
-    // Write actual block data (this may be
-    // better suited to flat files in the future).
-    this.put(layout.b.encode(hash), block.encode());
+    // Write actual block data
+    this.blocksBatch.writeBlock(hash, block.encode());
```
pinheadmz, line 1889: I like how you are preserving the batch API, but since most of our blockstore writes will be single blocks, I wonder if it's worth the overhead to create `new FileBatch` and `new WriteOp`? I guess those objects are pretty minimal but in a few years bootstrapping nodes will have 100,000s of these to write.

pinheadmz:
I'm going to try and play around with benchmarks today as well -- why didn't you include the file from bcoin? https://github.com/bcoin-org/bcoin/blob/master/bench/blockstore.js

pinheadmz:
Ok I copied over the benchmark script from bcoin (had to `write => writeBlock` etc) and confirmed of course that file blockstore is up to 4x faster like across the board. Maybe we could add batching to the benchmark script as well.

pinheadmz:
@nodech ok forgive me but I pushed a commit to your branch to include the benchmark file (you can revert if you want). Take a look at how I added batches to the benchmark. It definitely impacts performance but there should still be a big savings over LevelDB.

### File

```
                                        write                                        
=====================================================================================
length         operations     min            max            average        median         
-------------------------------------------------------------------------------------
256            959            227.24         20298.95       692.38         451.66         
512            178            254.56         21294.45       770.62         478.48         
1024           120            348.78         21635.96       951.96         593.03         
2048           119            268.85         3242.85        585.33         449.76         
4096           239            266.79         12588.55       650.42         478.11         
8192           179            308.24         7378.96        698.97         522.35         
16384          239            309.67         7825.80        670.27         470.92         
32768          300            274.09         9018.24        647.98         497.41         
65536          418            273.03         5880.21        681.87         523.75         
131072         479            296.48         54765.73       872.06         557.08         
262144         776            366.35         25166.01       949.42         704.10         
524288         537            567.31         39728.09       1495.62        914.01         
1048576        1436           746.65         53609.97       2267.80        1449.85        
2097152        418            1136.64        42973.54       4615.27        3110.61        
4194304        59             2669.67        49957.51       8550.91        6349.41        
```

### File +batch

```
                                     write +batch                                    
=====================================================================================
length         operations     min            max            average        median         
-------------------------------------------------------------------------------------
256            959            198.39         73259.30       787.47         464.54         
512            178            270.60         10113.47       706.59         520.72         
1024           120            334.06         9115.72        913.13         595.87         
2048           119            206.28         12267.21       839.67         431.06         
4096           239            283.31         10089.22       829.88         500.58         
8192           179            304.81         12369.71       891.53         535.13         
16384          239            292.16         9926.09        717.80         520.93         
32768          300            270.04         41346.42       928.04         468.16         
65536          418            272.81         84697.29       1190.36        551.22         
131072         479            313.76         39515.39       1116.52        633.54         
262144         776            308.24         201989.78      1548.82        794.35         
524288         537            485.72         79383.84       1779.00        1123.50        
1048576        1436           592.78         112505.27      3480.29        2068.70        
2097152        418            1119.43        95300.99       6288.05        4820.36        
4194304        59             2490.67        66035.28       10813.09       10274.23       


```

### Level

```
                                        write                                        
=====================================================================================
length         operations     min            max            average        median         
-------------------------------------------------------------------------------------
256            959            28.73          3885416.74     23625.33       1186.22        
512            178            35.33          12814.81       1407.86        636.64         
1024           120            91.96          3310665.67     167615.51      9210.43        
2048           119            43.43          12959.77       2080.04        542.74         
4096           239            47.94          16102.69       1982.71        613.92         
8192           179            53.07          3543046.26     70395.08       1684.84        
16384          239            59.66          19425.53       2074.04        689.30         
32768          300            77.92          4197203.82     44292.37       1399.60        
65536          418            134.38         3671786.08     31728.83       1468.19        
131072         479            241.35         3640675.42     23954.57       1645.05        
262144         776            341.32         3028005.12     17575.73       1994.05        
524288         537            632.03         39711.76       7080.98        2802.49        
1048576        1436           1203.68        2955310.34     18572.65       10158.23       
2097152        418            2539.23        2969985.32     44410.79       14319.11       
4194304        59             11472.26       60178.45       26852.13       24301.89       


```

### Level +batch

```
                                     write +batch                                    
=====================================================================================
length         operations     min            max            average        median         
-------------------------------------------------------------------------------------
256            959            36.38          4643719.25     26380.83       644.58         
512            178            45.11          12510.99       1508.79        638.05         
1024           120            88.86          4119987.24     192246.39      2445.59        
2048           119            39.20          13052.45       1471.33        246.24         
4096           239            49.21          24468.99       1852.19        504.36         
8192           179            55.01          4480789.16     76889.15       1200.40        
16384          239            62.96          43815.11       1772.72        184.80         
32768          300            75.41          4166503.90     46463.39       927.30         
65536          418            132.53         3745061.85     28914.15       1024.45        
131072         479            220.69         4533324.25     31449.56       1451.80        
262144         776            302.56         4658502.69     25072.69       1918.63        
524288         537            603.76         58021.06       6958.29        2760.62        
1048576        1436           1121.02        4170794.14     20691.97       8310.49        
2097152        418            2377.14        4627624.92     46444.07       13170.23       
4194304        59             7953.59        48146.32       23250.20       21577.52     
```

Edord99:
...

pinheadmz:
Something else to look in to:

https://github.com/bcoin-org/bcoin/issues/756

lib/blockstore/file.js
```diff
@@ -178,8 +183,8 @@ class FileBlockStore extends AbstractBlockStore {
 
   async index() {
     await this._index(types.BLOCK);
-    await this._index(types.MERKLE);
     await this._index(types.UNDO);
+    await this._index(types.MERKLE);
```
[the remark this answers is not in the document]
nodech: that's the way there were introduced to codebase, nothing big.

bench/blockstore.js
```diff
@@ -21,7 +21,8 @@
  * - `batch`    Bundle write & prune operations into batches.
  *
  * Copyright (c) 2019, Braydon Fuller (MIT License).
- * https://github.com/bcoin-org/bcoin
+ * Copyright (c) 2021, Matthew Zipkin (MIT License).
```
pinheadmz, line 24: lol. I'd rather "The Handshake Developers" I'm not one for vanity ;-)

lib/blockstore/index.js
```diff
@@ -1,7 +1,7 @@
 /*!
- * blockstore/index.js - bitcoin blockstore for bcoin
+ * blockstore/index.js - bitcoin blockstore for hsd
```
pinheadmz, line 2: hm are we gonna keep the word "bitcoin" though?

package-lock.json
```diff
@@ -1,8 +1,474 @@
 {
   "name": "hsd",
   "version": "2.4.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
pinheadmz, line 4: Did you intend to commit package-lock? Usually we don't, right? I've seen this new lockfile version, need to learn about it -- can it be a separate PR?

pinheadmz (commented):
Took a look at new commits and had a few questions. I'll run some e2e tests next on the migration.

pinheadmz:
This PR needs an update to the CHANGELOG as well - with a new `--chain-migrate` number!

pinheadmz:
Im running this branch on top of current master using a datadir that has already been migrated with --chain-migrate=1 and im getting this:

```
[error] (chain-migrations) Database needs migration(s):
  - Migrate ChainDB migrations - ChainDB migration layout has changed.
  - Chain State migration - Chain state is corrupted.
  - BlockStore migration - Move block and undo data to the blockstore from the chainDB.
```

SHouldnt the first two migrations not be listed?

lib/blockchain/migrations.js
```diff
@@ -207,6 +261,134 @@ class MigrateChainState extends AbstractMigration {
...
+
+    this.logger.info('Migrating blocks and undo blocks.');
+    this.logger.info('This may take a few minutes...');
+
+    await this.migrateBlocks();
+    await this.migrateUndoBlocks();
+
+    this.logger.info('Compacting database...');
+    this.logger.info('This may take a few minutes...');
+    await this.ldb.compactRange();
```
pinheadmz, line 313: Are there undesired consequences if the user kills the process while `compactRange` is running?

package-lock.json
nodech: We have package-lock committed and we usually commit (that's some minimal checksum for you), but yeah I used latest node/npm which apparently upgrades package-lock version with new format. So it's better to do in separate PR.

test/chain-migration-test.js
```diff
@@ -664,6 +664,121 @@ describe('Chain Migrations', function() {
...
+    it('should mine 10 blocks', async () => {
+      for (let i = 0; i < 10; i++) {
+        const block = await cpu.mineBlock();
+        assert(block);
+        assert(await chain.add(block));
+        blocks.push(block);
+      }
+    });
+
+    it('should move blocks back to ldb.', async () => {
```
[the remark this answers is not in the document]
nodech: The content itself does not really make a difference. It could be just random bytes (as you can observe in unit tests for the blockstore itself). This test is here to make sure everything around blockstore behaves properly - integration test for chain + blockstore.

nodech:
> Im running this branch on top of current master using a datadir that has already been migrated with --chain-migrate=1 and im getting this:
> 
> ```
> [error] (chain-migrations) Database needs migration(s):
>   - Migrate ChainDB migrations - ChainDB migration layout has changed.
>   - Chain State migration - Chain state is corrupted.
>   - BlockStore migration - Move block and undo data to the blockstore from the chainDB.
> ```
> 
> SHouldnt the first two migrations not be listed?

Fix in https://github.com/handshake-org/hsd/pull/632

lib/blockstore/file.js
```diff
@@ -0,0 +1,789 @@
...
+  /**
+   * Creates indexes from files for a block type. Reads the hash of
+   * the block data from the magic prefix, except for a block which
+   * the hash is read from the block header.
+   * @private
+   * @param {Number} type - The type of block data
+   * @returns {Promise}
+   */
+
+  async _index(type) {
```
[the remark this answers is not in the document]
nodech: That is definitely something to look into, I would love to have block store as flexible as possible with different sources of corruption. I think for this PR keeping it how it is currently is okay and we can have discussion on new issue how to approach those things.

E.g. if 1 blk + blu is missing, should it only download those and skip the rest (How much rewrite would that require)
only blk or only blu is missing and etc. At least we could add check and fail early, instead of node getting stuck at certain height.

test/chain-migration-test.js
```diff
@@ -664,6 +664,121 @@ describe('Chain Migrations', function() {
...
+      const ldb = chain.db.db;
+
+      const ldbBatch = ldb.batch();
+      const blocksBatch = store.batch();
+
+      for (const block of blocks) {
+        const hash = block.hash();
+
+        // we don't actually have undo blocks with those blocks.
+        const undoData = Buffer.alloc(100, 1);
```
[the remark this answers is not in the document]
nodech: I believe I answered this on different comment.

lib/blockchain/migrations.js
nodech: TLDR: It should not create problems.
NOTE: If migration is terminated, state wont increase so on restart that
migration will restart. That's why it's important to write migrations with
restarts in mind (e.g. FileBlockstore doing it in batches)

okay, I am not 100% sure on this one tbh, cause there are many things to
consider. There's not single way to force the process to exit, but we can
discuss KILL and SIGINT.
  bin/node has SIGINT handler, which means that default nodejs handler wont kick
in. It will call `node.close` which will in turn close everything and wait if
some things need time to close. In this scenario compactRange will just finish
executing before everything closes.
  If something else is used to kill the process, then I can not guarantee that
level-db compactRange wont screw things up. I will need to review it deeply to
confidently answer the question. After several tests I could not produce any
corruption, on the next start it just continued doing compacting db. But the
same is true for many other operations, like Batch - what if process is killed
while level-db batch.write is being executed.

lib/blockchain/migrations.js
```diff
@@ -150,10 +153,10 @@ class MigrateChainState extends AbstractMigration {
 
     for (let height = 0; height <= tipHeight; height++) {
       const hash = await this.db.getHash(height);
-      const block = await this.db.getBlock(hash);
+      const block = await this.getBlock(hash);
       assert(block);
 
-      const view = await this.db.getBlockView(block);
+      const view = await this.getBlockView(block);
```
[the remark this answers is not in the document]
nodech: block.txs is also necessary.

lib/blockstore/file.js
nodech: yep, `bio.write(size)` with size will use StaticWriter which allocates all at once instead of BufferWriter which is dynamic. If you don't pass `size` then you end up with BufferWriter and yeah render just returns allocated buffer (only checks for offset for out of bounds write)

lib/blockchain/chaindb.js
```diff
@@ -338,6 +348,9 @@ class ChainDB {
     this.cacheHash.commit();
     this.cacheHeight.commit();
     this.stateCache.commit();
+
+    if (this.blocks)
+      await this.blocksBatch.write();
```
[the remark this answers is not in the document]
nodech: This follows exact logic as in start.

So I think it's not necessary to do that. Commit wont work without start and start will init blacksBatch if this.blocks exists.

lib/blockchain/chaindb.js
nodech: Without preserving batch API the way logic works becomes significantly harder to follow. You can check bcoin for that, it comes to the point that `removeBlock` method in bcoin does not remove the block and it becomes responsibility of the caller.

So I believe performance hit is negligible compared to the things that happen around that method.

lib/blockstore/file.js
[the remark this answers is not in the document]
pinheadmz: Ok cool good plan.

lib/blockchain/migrations.js
pinheadmz: Ok, sounds like this is the best we can do anyway

nodech:
> I like how you are preserving the batch API, but since most of our blockstore writes will be single blocks, I wonder if it's worth the overhead to create new FileBatch and new WriteOp? I guess those objects are pretty minimal but in a few years bootstrapping nodes will have 100,000s of these to write.
@pinheadmz 

I think I misunderstood this comment so I will elaborate further. CPU overhead as I said is pretty much negligible in
this context, only thing that could create BIG issues with BatchAPI would be if Block Batch could get big somehow.
But that is not the case, so I kinda skipped that in my response. So here's some review of the code and where to look
at to understand if Blocks batch can ever be bigger than 2 writes (1 Block + 1 Undo):   

  Blockstore Batch API is integrated with existing chaindb Batch API,
meaning start, commit and drop will also be responsible for managing
Blockstore Batch API. I think this makes review of the BatchAPI easier,
as we can follow these scopes.
  Only places that are critical for the BatchAPI are when we store the Block
or Prune data, because if we can have Batch with many blocks memory would
grow unnecesserily. Only place method storing block data is
`chaindb.saveBlock` - it is responsible for saving alt and main chains, but
in both cases they are called from `chaindb.save` method where we initialize
the Batch.
  `connectBlock` is the only place storing UNDO data and is called by
`saveBlock` and `reconnect`. saveBlock - when we are storing new blocks and
reconnect when we are reorganizing. Similar to saveBlock, reconnect is also
separate instance of BatchAPIs and only results in 1 UNDO write.
  Even if we had multiple blocks or undos to write, they would happen one at
a time in terms of Batches. Of course these are further guarded with locker in
Chain, so we don't have data races.

pinheadmz:
re: aborting process during `compacting database...` I tried this out with a ctrl-C and got great results:

first run, cancelled:
```
[info] (chain-migrations) Migration 2 in progress...
[info] (chain-migration-blockstore) Migrating blocks and undo blocks.
[info] (chain-migration-blockstore) This may take a few minutes...
[info] (chain-migration-blockstore) Migrating blocks...
[info] (chain-migration-blockstore) Migrated all 1289 blocks.
[info] (chain-migration-blockstore) Migrating undo blocks...
[info] (chain-migration-blockstore) Migrated all 276 undo blocks.
[info] (chain-migration-blockstore) Compacting database...
[info] (chain-migration-blockstore) This may take a few minutes...
```

second run, after restart (error message required me to run with `--chain-migrate=2` again)
```
[info] (chain-migrations) Continue progress on migration:  2
[info] (chain-migrations) Migration 2 in progress...
[info] (chain-migration-blockstore) Migrating blocks and undo blocks.
[info] (chain-migration-blockstore) This may take a few minutes...
[info] (chain-migration-blockstore) Migrating blocks...
[info] (chain-migration-blockstore) Migrated all 0 blocks.
[info] (chain-migration-blockstore) Migrating undo blocks...
[info] (chain-migration-blockstore) Migrated all 0 undo blocks.
[info] (chain-migration-blockstore) Compacting database...
[info] (chain-migration-blockstore) This may take a few minutes...
[info] (chain-migrations) Migration 2 is done.
```

Note that this time the number of blocks migrated is 0, but compacting database still took a minute (this node was pruned)
Good job! I like.

lib/node/fullnode.js
```diff
@@ -41,10 +42,20 @@ class FullNode extends Node {
     // SPV flag.
     this.spv = false;
 
+    // Instantiate block storage.
+    this.blocks = blockstore.create({
+      network: this.network,
+      logger: this.logger,
+      prefix: this.config.prefix,
+      cacheSize: this.config.mb('block-cache-size'),
+      memory: this.memory
```
pinheadmz, line 51: I was confused about this at first until I realized `this.memory` is inherited from the `node` parent class. I don't think this needs to be addressed in this PR since we are so close to merging, but we could probably replace all the memory settings with `this.memory` in the other modules like `chain` and `pool`.

pinheadmz:
LGTM, Tested this branch on existing data directories and checked migration process:
- [x] SPV v2.0.1
- [x] SPV v2.0.1 migrated once to master branch
- [x] Full v2.0.1, pruned, then migrated once to master branch
- [x] Full v2.0.1

The longest migration was the last one, a full node that hasn't been upgraded basically ever. This required 3 chain migrations and 2 wallet migrations. 78,804 blocks had to be migrated. The entire operation completed in less than 3 minutes!!!! :rocket:

pinheadmz (approved):
ACK b779dfc0c4bd6ca63733a3a9461528dd697ccc7b

```
Hash: SHA256

ACK b779dfc0c4bd6ca63733a3a9461528dd697ccc7b
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmE2ZG8ACgkQ5+KYS2KJ
yTpWBA//WBnwzJK7SWZwBUu2G1kLixu946aRGroc/6iVVtLGrHHJ25J7BnQRJBLb
p6fyD3a7bQFzFoqQIUMoIqkGflF9W/h6uoACr22pLFxWOzVGVS8QDqx94OJ8HfPF
zejBK1wTWvVP8RWovuc7pFGYOOGxpLffKxBGf8PH8nV0XxQEnnSK1V9Di39WspVW
Gr7g9zHaVQYx+Ts07T9s2MpG+D1iu1XCNdZNoX7D+jBN16FU7eA/fFue0E7TLA/k
ZvE7vRE/5pZhK5Qia81c/jPIhzbZEqu7zIn75o5tP2J5l94zR9PkGvA//ZQbPcBl
9PQrpKbJfPQIadwqbbGF9i9soblV+vpGr7cwGILjhJ9kwiVtdUi5SrLi2+rUJOUu
50YaDGxgJH8W2Z2nkXhgZ6U/xc/yv9HdAHRdsgfjdvHpEiWYDFx+32+8BPvrL+ig
2+eQC0+wGPTjOknxYHbCdmEfPwkhI9/jcebw3plGCFhoCY6uYH65smPwytItawCa
HBlILqU7ITzjGPkZAIKJE5vT4fQkR2t+QbM5E+mYiwzhjRIT9kgH7uGbwtyfXXpL
Gr4gkY4xsClrEeCRVENF05fuJ/7JrS3qpIkgu7CUnXucNf3jEAIw6PLJzA5H0sd0
fZuVQzV43FBoyP0yEO3p0S1PZCWodojtcHYfvM5ft5vNwtWHRh4=
=b1bk
-----END PGP SIGNATURE-----
```

</details>

lib/blockchain/chaindb.js
```diff
@@ -888,34 +902,36 @@ class ChainDB {
...
-      b.put(layout.O.encode(), flags.encode());
-
-      await b.write();
+      await this.db.put(layout.O.encode(), flags.encode());
     } catch (e) {
       options.prune = false;
       throw e;
     }
 
-    await this.db.compactRange();
```
tuxcanfly, line 917: I think this was here to compact the db at open, is this not required anymore?

tuxcanfly: Is see that this has moved to migrations. I think a note might make this more clear.

lib/blockstore/file.js
```diff
@@ -0,0 +1,791 @@
...
+  }
+
+  /**
+   * Opens the file block store. It will regenerate necessary block
+   * indexing if the index is missing or inconsistent.
+   * @returns {Promise}
+   */
+
+  async open() {
+    this.logger.info('Opening FileBlockStore...');
```
tuxcanfly, line 212: nitpick: this log and it's corresponding closing log looks a bit inconsistent. Can it say 'Opening/Closing block store' instead?

lib/blockchain/chaindb.js
```diff
@@ -225,6 +227,9 @@ class ChainDB {
     this.current = this.db.batch();
     this.pending = this.state.clone();
 
+    if (this.blocks)
```
tuxcanfly, line 230: Do we have a test plan for cases where `chain` and `blocks` directories are out-of-sync due to factors outside of the process? For e.g. either directory gets corrupted and has to be `rm`ed and restarted from scratch. This wasn't an issue before as there was only one db.

lib/blockchain/chaindb.js
nodech: This was here for pruning. For some cases, it could prune more than a single block. E.g. with retroactive pruning it could prune many blocks - in that case compactRange was necessary. So migration compactRange is not exactly replacement for this, here we no longer remove data from level db so compactRange loses it's purpose here.

lib/blockstore/file.js
nodech: I don't get it: both say `Opening FileBlockStore...` and `Closing FileBlockStore...`, which I believe was used for consistency with WalletDB and ChainDB logging.

lib/blockchain/chaindb.js
nodech: Well, you could say it was a case previously as well, because ldb has many files and some of them could be removed by outside process.

Currently, there's no such thing - but I think in the future we can do better and this comment by @pinheadmz goes well with this issue: https://github.com/handshake-org/hsd/pull/612#discussion_r660906283
Making blockstore files more flexible.

0xStefan:
Was able to upgrade a Full v2.4.0 node without any issues.

lib/blockchain/chaindb.js
tuxcanfly: Yep sounds good. Something to note for later.

lib/blockstore/file.js
tuxcanfly: I guess that's OK. Was concerned it's a little bit too technical to go into logs as `info`.

lib/blockchain/chaindb.js
tuxcanfly: OK, thanks that makes sense.

tuxcanfly (approved):
OK.

PR #615, opened 2021-06-28
Migrations

nodech:
Move out migrations from ChainDB and WalletDB.

  - [x] Use database Version instead of chainState for initialization.
  - [x] Initialization code does not use batches. Rewrite so it uses batches.
  - [x] Skip to necessary migration step using Backed up Last migration.
  - [x] WalletDB connect/watch should move at the end of the open.


lib/migrations/state.js
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Get size of the encoded migration state object.
+   * @returns {Number}
+   */
+
+  getSize() {
+    // flags + last migration number
+    let size = 4; // flags
+    size += encoding.sizeVarint(this.lastMigration); // last migration ID
```
pinheadmz, line 60: You don't wanna have a little confidence that we'll need less than 4,294,967,295 migrations and lock this in to a slim U32? Same with `skipped`...?

lib/migrations/migrations.js
```diff
@@ -0,0 +1,350 @@
...
+      // move migration entries if we are migrating
+      // from old version with Migrations.
+      // p.s. if there are no migrations:
+      //  1. database is new
+      //  2. we are migrating from even older version.
+      const oldMigrations = await this.ldb.keys({
+        gte: oldLayout.M.min(),
+        lte: oldLayout.M.max(),
+        parse: key => oldLayout.M.decode(key)[0]
+      });
```
pinheadmz, line 284: I find it a little funny that this new Migrations module... requires a migration. Are there any failure states where this migration of `layout.M` gets interrupted or incomplete? I wonder if the new migrations DB prefix should be different so it's always fresh for this upgrade, and we can just mark `layout.M` as `// deprecated`.

lib/migrations/migrations.js
```diff
@@ -0,0 +1,350 @@
...
+
+      let max = -1;
+      for (const id of oldMigrations) {
+        b.del(oldLayout.M.encode(id));
+        max = Math.max(id, max);
+      }
+
+      if (max > -1)
+        state.lastMigration = max;
```
pinheadmz, line 293: This makes me wonder if the default value for `lastMigration` should be `-1`? You have `this.lastMigration = 0;` in the `MigrationState` constructor.

lib/migrations/migrations.js
```diff
@@ -0,0 +1,350 @@
...
+  /**
+   * Get max migration ID from the map
+   * @returns {Number}
+   */
+
+  getLastMigrationID() {
+    const ids = Object.keys(this.migrations);
+
+    if (ids.length === 0)
+      return 0;
```
pinheadmz, line 310: Same here -- since we used migration ID `0` for  https://github.com/handshake-org/hsd/pull/396, maybe "no migrations" should be represented by `-1`.

lib/migrations/migrations.md
```diff
@@ -0,0 +1,146 @@
+## HSD Migration
```
pinheadmz, line 1: I think some version of this file should be left in place in this directory, like the README in blockstore. If for no other reason, then to explain to future developers how to deploy a migration with the class. I think you've made a very elegant abstraction here, all thats needed to deploy is essentially a `check()` and `migrate()` methods.

lib/blockchain/migrations.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  /**
+   * Run all migrations
+   * @returns {Promise}
+   */
+
+  async migrate() {
+    return super.migrate();
+  }
```
pinheadmz, line 161: Is this necessary?

lib/blockchain/migrations.js
```diff
@@ -0,0 +1,253 @@
...
+  Migrations,
+  types
+} = require('../migrations/migrations');
+
+/**
+ * Migrate chain state and correct total supply.
+ * Applies to ChainDB v1
+ */
+
+class MigrateChainState extends AbstractMigration {
```
pinheadmz, line 25: I think we talked about making this name more specific or linking to the github discussions somehow as metadata (this migration is old-migration 0 and 1 which are https://github.com/handshake-org/hsd/pull/396 and https://github.com/handshake-org/hsd/pull/585)

pinheadmz (commented):
I like where this is headed. There is a lot of work to be done on the wallet txdb to optimize coin selection and it might require a few migrations ;-) I have some questions about design decisions but I think it's overall great work. Once this is "ready for review" I'll sync some old hsd versions in various modes and try it out to practice the UI.

lib/blockchain/chaindb.js
```diff
@@ -48,6 +50,8 @@ class ChainDB {
     this.logger = this.options.logger.context('chaindb');
 
     this.db = bdb.create(this.options);
+    this.dbName = 'chain';
+    this.dbVersion = 1;
```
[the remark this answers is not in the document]
nodech: Actually I was thinking of using them, maybe it's a better name )

lib/blockchain/chaindb.js
```diff
@@ -128,99 +135,69 @@ class ChainDB {
...
+  /**
+   * Write chaindb version.
+   * @param {Batch} b
+   */
 
-    batch.put(layout.M.encode(0), null);
-    batch.put(layout.M.encode(1), null);
-    batch.put(layout.R.encode(), pending.encode());
+  writeVersion(b) {
+    const value = Buffer.alloc(this.dbName.length + 4);
```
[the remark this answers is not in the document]
nodech: This part did not actually belong to the chaindb implementation and currently this code is taken from `bdb`: https://github.com/bcoin-org/bdb/blob/master/lib/db.js#L538

lib/migrations/migrations.js
```diff
@@ -0,0 +1,350 @@
...
+  constructor(options) {
+    this.options = options;
+    this.logger = Logger.global;
+
+    this.migrations = {};
+    this.migrateFlag = false;
+
+    this.layout = migrationLayout;
+    this.db = null;
+    this.ldb = null;
```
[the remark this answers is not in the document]
nodech: Hm.. ChainDB is actual DB implementation (with leveldb backend) so I used `db` and `ldb` for leveldb. As walletdb is the db and vice versa.

lib/migrations/state.js
nodech: Actually it's the reverse, I don't think it will get bigger than 253 (which will take 1 byte), even if it does it's highly unlikely to get bigger than 65k - We definitely have different problems at that point.

So up to 253 it will just take a byte instead of 4 and is still flexible for bigger numbers.

lib/migrations/migrations.js
nodech: We can definitely mark layout.M deprecated and use different prefix for this.
We currently have single migration in both wallet and chain dbs, so instead of iteration I can just check if layout.M exists - then we have done THAT ONE migration so skip it. It can make code somewhat cleaner.

I just did not want to leave unused prefix in the layout and thought this was not that much complicated, but it probably is

lib/migrations/migrations.js
nodech: I am thinking about lastMigration as unsigned int, so in the db we can never have -1 state for that and instead we use 0 as default and start counting from 1. Not sure if which one is better though )

lib/migrations/migrations.js
nodech: ChainDB has merged 0 and 1 migrations together, so even if you have migration 0 in there it does not matter you still need to migrate. hm.. 
But walletdb currently uses 0 as "lastMigration".
We will need to change from varint to int32.

Will look into this.

lib/migrations/migrations.md
nodech: yep, we can rewrite this to README.

lib/blockchain/migrations.js
nodech: Nope, not necessary.

lib/blockchain/migrations.js
nodech: Yep, that's one part of this migration that I have not touched so far. How to include information about the migration: whether it should be method so it can be called by API user or not. Have not decided yet.
Maybe properties ?
 ```
  this.issue = 396
  this.pr = 585
```
or have just a method that returns String description with links and small summary?

File: test/util/common.js [hunk 11]
```diff
@@ -85,6 +87,20 @@ common.writeTX = function writeTX(name, tx, view) {
...
+  const uniq = randomBytes(4).toString('hex');
+  return path.join(tmpdir(), `hsd-test-${name}-${uniq}`);
+};
+
+common.rimraf = async function(p) {
+  const allowed = /hsd\-test\-[a-z\-]+\-[a-f0-9]{8}((\\|\/)[a-z]+)?$/;
+  if (!allowed.test(p))
+    throw new Error(`Path not allowed: ${p}.`);
+
+  return await fs.rimraf(p);
```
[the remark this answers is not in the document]
nodech: actually, I just forgot `await` in one of the `assert.rejects`, that was causing the uncertain outcomes.

lib/blockchain/migrations.js
pinheadmz: Probably the `warning()` message is enough, like we did here: https://github.com/handshake-org/hsd/blob/master/lib/wallet/walletdb.js#L265-L268

I don't think `this.issue` or `this.pr` are great, but maybe an optional `this.description = <string>` which can mention issue or PR number.

pinheadmz:
Looking good so far, I synced v2.0.1 and then upgraded to this branch:

```
pinheadmz@pintu:~/Desktop/work/hsd$ hsd --prefix=~/.hsd/v2.0.1
[info] (chain) Chain is loading.
[info] (chain) Checkpoints are enabled.
[info] (chain-migrations) Opening DB...
[error] (chain-migrations) Database needs migration(s):
  - Migrate ChainDB migrations - ChainDB migration layout has changed.
  - Chain State migration - Chain state is corrupted.
Restart with `hsd --chain-migrate`.
[info] (chain-migrations) Closing DB...
Error: Database needs migration(s):
  - Migrate ChainDB migrations - ChainDB migration layout has changed.
  - Chain State migration - Chain state is corrupted.
Restart with `hsd --chain-migrate`.
    at ChainMigrations.checkMigrations (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:331:13)
    at async ChainMigrations._migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:228:5)
    at async ChainMigrations.migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:197:14)
    at async ChainDB.open (/home/pinheadmz/Desktop/work/hsd/lib/blockchain/chaindb.js:79:5)
    at async Chain.open (/home/pinheadmz/Desktop/work/hsd/lib/blockchain/chain.js:86:5)
    at async FullNode.open (/home/pinheadmz/Desktop/work/hsd/lib/node/fullnode.js:259:5)
    at async /home/pinheadmz/Desktop/work/hsd/bin/node:55:3


pinheadmz@pintu:~/Desktop/work/hsd$ hsd --prefix=~/.hsd/v2.0.1 --chain-migrate
[info] (chain) Chain is loading.
[info] (chain) Checkpoints are enabled.
[info] (chain-migrations) Opening DB...
[info] (chain-migrations) Database needs migration(s):
  - Migrate ChainDB migrations - ChainDB migration layout has changed.
  - Chain State migration - Chain state is corrupted.

[debug] (chain-migrations) Last migration -1, last available migration: 1
[info] (chain-migrations) There are 2 migrations.
[info] (chain-migrations) Migration 0 in progress..
[info] (chain-migration-migrate) Migrating migrations..
[info] (chain-migrations) Migration 0 is done.
[info] (chain-migrations) Migration 1 in progress..
[info] (chain-migration-chainstate) Migrating chain state.
[info] (chain-migration-chainstate) This may take a few minutes...
```

pinheadmz:
Nice, then the wallet migration:

```
[info] (wallet-migrations) Opening DB...
[error] (wallet-migrations) Database needs migration(s):
  - Migrate wallet migrations - Wallet migration layout has changed.
  - Change address migration - Wallet is corrupted.
`hsd --wallet-migrate` or `hsw --migrate`
(Full node may be required for rescan)
[info] (wallet-migrations) Closing DB...
Error: Database needs migration(s):
  - Migrate wallet migrations - Wallet migration layout has changed.
  - Change address migration - Wallet is corrupted.
`hsd --wallet-migrate` or `hsw --migrate`
(Full node may be required for rescan)
    at WalletMigrations.checkMigrations (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:331:13)
    at async WalletMigrations._migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:228:5)
    at async WalletMigrations.migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:197:14)
    at async WalletDB.open (/home/pinheadmz/Desktop/work/hsd/lib/wallet/walletdb.js:188:29)
    at async Plugin.open (/home/pinheadmz/Desktop/work/hsd/lib/wallet/plugin.js:86:5)
    at async FullNode.openPlugins (/home/pinheadmz/Desktop/work/hsd/lib/node/node.js:456:9)
    at async FullNode.open (/home/pinheadmz/Desktop/work/hsd/lib/node/fullnode.js:264:5)
    at async /home/pinheadmz/Desktop/work/hsd/bin/node:55:3
```

pinheadmz:
Good error if migrate flag is set again:

```
pinheadmz@pintu:~/Desktop/work/hsd$ hsd --prefix=~/.hsd/v2.0.1 --chain-migrate
[info] (chain) Chain is loading.
[info] (chain) Checkpoints are enabled.
[info] (chain-migrations) Opening DB...
[debug] (chain-migrations) There are no migrations pending.. last id: 1
[debug] (chain-migrations) Last migration 1, last available migration: 1
[info] (chain-migrations) There are 0 migrations.
[info] (chain-migrations) Closing DB...
Error: There are no available migrations.
    at ChainMigrations._migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:300:13)
    at async ChainMigrations.migrate (/home/pinheadmz/Desktop/work/hsd/lib/migrations/migrations.js:197:14)
    at async ChainDB.open (/home/pinheadmz/Desktop/work/hsd/lib/blockchain/chaindb.js:79:5)
    at async Chain.open (/home/pinheadmz/Desktop/work/hsd/lib/blockchain/chain.js:86:5)
    at async FullNode.open (/home/pinheadmz/Desktop/work/hsd/lib/node/fullnode.js:259:5)
    at async /home/pinheadmz/Desktop/work/hsd/bin/node:55:3
```

pinheadmz:
Wallet migration looks good:

```
[info] (wallet-migrations) Opening DB...                                                                                                                                                                   
[info] (wallet-migrations) Database needs migration(s):                                                                                                                                                    
  - Migrate wallet migrations - Wallet migration layout has changed.                                                                                                                                       
  - Change address migration - Wallet is corrupted.                                                                                                                                                        
                                                                                                                                                                                                           
[debug] (wallet-migrations) Last migration -1, last available migration: 1                                                                                                                                 
[info] (wallet-migrations) There are 2 migrations.                                                                                                                                                         
[info] (wallet-migrations) Migration 0 in progress..                                                                                                                                                       
[info] (wallet-migrations-migrate) Migrating migrations..                                                                                                                                                  
[info] (wallet-migrations) Migration 0 is done.                                                                                                                                                            
[info] (wallet-migrations) Migration 1 in progress..                                                                                                                                                       
[info] (wallet) Wallet opened (primary).                                                                                                                                                                   
[info] (change-address-migration) Checking wallet (id=primary, wid=0).                                                                                                                                     
[info] (wallet) Wallet opened (abandon).                                                                                                                                                                   
[info] (change-address-migration) Checking wallet (id=abandon, wid=1).                                                                                                                                     
[info] (wallet) Wallet opened (stolen).
[info] (change-address-migration) Checking wallet (id=stolen, wid=2).
[info] (wallet-migrations) Migration 1 is done.
```

pinheadmz:
Chain migration succeeded from v2.0.1 to master+PR:

```
  "chain": {                                                                                                                                                                                               
    "height": 77275,                                                                                                                                                                                       
    "tip": "00000000000000014e341cea8f1bd3d14afd7a9815c83a9ac95095b6af847d90",                                                                                                                             
    "treeRoot": "e66dd289837d4cd5373865b1ec9a255973bc1f98a349614237165d0779ee6406",                                                                                                                        
    "progress": 1,                                                                                                                                                                                         
    "state": {                                                                                                                                                                                             
      "tx": 4438857,                                                                                                                                                                                       
      "coin": 2977029,                                                                                                                                                                                     
      "value": 381320814155649,                                                                                                                                                                            
      "burned": 22478073560483                                                                                                                                                                             
    }                                                                                                                                                                                                      
  },   
```

pinheadmz:
Also confirmed behavior for "fresh start" (using `memory`)

```
--> hsd --memory=true
[info] (chain) Chain is loading.
[info] (chain) Checkpoints are enabled.
[info] (chain-migrations) Opening DB...
[info] (chain-migrations) Fresh start, saving last migration id: 1
[info] (chain-migrations) Closing DB...
```

lib/migrations/state.js
```diff
@@ -0,0 +1,117 @@
...
+ */
+'use strict';
+
+const assert = require('bsert');
+const bio = require('bufio');
+const {encoding} = bio;
+
+/**
+ * State of database migrations.
+ * because migration IDs are only increasing, we only need
```
pinheadmz, line 13:
```suggestion
 * Because migration IDs are only increasing, we only need
```

lib/migrations/state.js
```diff
@@ -0,0 +1,117 @@
...
+ * to store the last one.
+ * @alias module.migrations.MigrationState
+ * @property {Boolean} inProgress - is/was migration in progress
+ * NOTE: If inProgress is true, we know lastMigration + 1 was the one in
+ *       the progress.
```
pinheadmz, line 18:
```suggestion
 * NOTE: If inProgress is true, we know lastMigration + 1 was the one in progress
```

lib/migrations/state.js
```diff
@@ -0,0 +1,117 @@
...
+   * @returns MigrationState;
+   */
+
+  inject(obj) {
+    assert(obj instanceof MigrationState);
+    this.inProgress = obj.inProgress;
+    this.nextMigration = obj.nextMigration;
+    this.skipped = obj.skipped.slice();
+    return this;
+  }
```
pinheadmz, line 54: Is this ever used? Or just to have general compatibility with `bio.Struct` ?

lib/wallet/migrations.js
```diff
@@ -0,0 +1,301 @@
...
+   * @returns {Promise}
+   */
+
+  async migrate(b) {
+    this.logger.info('Migrating migrations..');
+    const state = new MigrationState();
+    state.nextMigration = 1;
+
+    if (await this.ldb.get(oldLayout.M.encode(0))) {
+      b.del(oldLayout.M.encode(0));
```
pinheadmz, line 59: What are the consequences of this for backwards compatibility? Like if the user for whatever reason upgrades, migrates, then tries to downgrade? I think we can prevent that case with good docs but just curious. (Blockstore wlll probably be hsd v3.0.0 anyway so broken compatibility should be clear)

lib/wallet/migrations.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    if (total > 0)
+      pending.rescan = true;
+  }
+
+  async migrateWallet(b, wallet) {
+    let total = 0;
+    for (let i = 0; i < wallet.accountDepth; i++) {
+      const account = await wallet.getAccount(0);
```
pinheadmz, line 136:
```suggestion
      const account = await wallet.getAccount(i);
```

right?? That's how the original code for this migration looked

lib/wallet/migrations.js
```diff
@@ -0,0 +1,301 @@
...
+
+      for (let i = 0; i < account.changeDepth + account.lookahead; i++) {
```
pinheadmz, line 138: ...then the nested loop should probably use `j` instead of `i`... I'm actually kinda surprised JJ wrote it this way even though I know those variables are scoped correctly in the for loop.

pinheadmz:
Also successfully migrated from v2.0.1 in full, prune and SPV mode (SPV required a change to `SPVNode`) :+1: 

Also good to see the warning in every boot in prune mode when migration is skipped:

```
[debug] (chain-migrations) There are no migrations pending.. last id: 1
[debug] (chain-migrations) Last migration 1, last available migration: 1
[info] (chain-migrations) There are 0 migrations.
[warning] (chain-migration-chainstate) Pruned nodes cannot migrate the chain state.
[warning] (chain-migration-chainstate) Your total chain value may be inaccurate!
```

lib/wallet/walletdb.js
```diff
@@ -207,67 +219,27 @@ class WalletDB extends EventEmitter {
 
     this.primary = wallet;
 
-    await this.migrateChange();
+    if (migrationResult.rescan) {
+      this.logger.info('Rescanning...');
+      await this.scan();
```
pinheadmz, line 224: I'm not sure but we might want to rescan from genesis here (`this.scan(0)`) In walletDB state there's this `startHeight` property but I think it's set every TX or something?

pinheadmz (changes requested):
Only found 2 important changes (SPVNode and the getAccount(i) thing in wallet migration). Everything else here is more style/design questions and just making sure I understand everything. GREAT WORK DOOD.

lib/wallet/walletdb.js
```diff
@@ -176,8 +179,17 @@ class WalletDB extends EventEmitter {
...
+      walletDB: this,
+      dbVersion: this.version
+    });
+
+    const migrationResult = await migrations.migrate();
+
+    this.logger.info('Opening WalletDB...');
     await this.db.open();
-    await this.db.verify(layout.V.encode(), 'wallet', 0);
+    await this.db.verify(layout.V.encode(), this.name, this.version);
```
pinheadmz, line 192: Why are leaving walletdb verify simple like this but in chainDB you replaced this line entirely with get version / initialize / verifyVersion, etc?

lib/migrations/state.js
nodech: This one is not used anymore, it was for better `.clone`. But left it there if we need to clone it again.

lib/wallet/migrations.js
nodech: This is when you are upgrading from the latest version: You have already migrated change addresses. This part makes sure we clean up oldLayout entries as well ass make sure we don't rerun the change address migration. It's similar thing to chain/migrations - there I iterate over oldEntries because there might be 0 and 1. (Could be done similar way tbh)

lib/wallet/migrations.js
nodech: totally !!! Great catch

lib/wallet/walletdb.js
nodech: Chaindb logic depends on the "atomicity" of the fresh start (even though it did not have atomicity there) and difference between fresh and restarting dbs is big. In wallet however, there's no such difference and there's no need to do them separately.

lib/migrations/migrations.js
```diff
@@ -0,0 +1,417 @@
...
+}
+
+/**
+ * class for migrations.
+ * @alias module:migrations.Migrations
+ */
+
+class Migrations {
+  /**
+   * Create AbstractMigrations object.
```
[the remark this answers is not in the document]
nodech: hm.. `class Migrations` was initially called `AbstractMigrations` :)

lib/wallet/walletdb.js
nodech: Hm.. this follows old `migration` code.

`startHeight` is initial height of the walletDB initialization, it does not ever increase. It can decrease though. (e.g. rollback)

lib/migrations/state.js
pinheadmz: ok :+1:

lib/wallet/migrations.js
pinheadmz: ok got it :+1:

pinheadmz:
In the files `{blockchain, wallet}/migrations.js` in the `migrate(b)` function. You have these hard-coded values:

`state.nextMigration = 2` and `this.db.writeVersion(b, 2);` (for example, from chain) -- so if I'm not mistaken in this case hsd should be run with `hsd --chain-migrate=2` and afterwards the database version will be `2` ?

I think in both files, these numbers should be set at the very top of the file as constants outside the class (wherever you think it looks good, but should be easily visible).

Something like `const CHAIN_MIGRATE_VERSION = 2` and `CHAIN_TARGET_DB_VERSION=2` or something?

And while we're at it, in the CHANGELOG I think you should specify "users upgrading to this version of hsd must start for the first time with `--chain-migrate=2 --wallet-migrate=2`"

lib/wallet/walletdb.js
pinheadmz: As discussed I think this should be `scan(0)` and I'll open an issue about `startHeight` in general.

lib/blockchain/migrations.js
```diff
@@ -0,0 +1,342 @@
...
+   * Create ChainMigrator object.
+   * @constructor
+   * @param {Object} options
+   */
+
+  constructor(options) {
+    super(new ChainMigratorOptions(options));
+
+    this.logger = this.options.logger.context('chain-migrations');
+    this.flagError = 'Restart with `hsd --chain-migrate`.';
```
pinheadmz, line 225: If you like what I suggested [here](https://github.com/handshake-org/hsd/pull/615#issuecomment-889318944) then this error message could actually report the exact value (example `--chain-migrate=2`). Same goes for wallet.

test/util/migrations.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    if (options.migrations != null) {
+      assert(typeof options.migrations === 'object');
+      this.migrations = options.migrations;
+    }
+  }
+}
+
+class MockChainDBMitrator extends Migrator {
```
pinheadmz, line 144:
```suggestion
class MockChainDBMigrator extends Migrator {
```

Typo is repeated a few places.... sorry

nodech: haha, rename all problems.

nodech:
> In the files `{blockchain, wallet}/migrations.js` in the `migrate(b)` function. You have these hard-coded values:
> 
> `state.nextMigration = 2` and `this.db.writeVersion(b, 2);` (for example, from chain) -- so if I'm not mistaken in this case hsd should be run with `hsd --chain-migrate=2` and afterwards the database version will be `2` ?
> 
> I think in both files, these numbers should be set at the very top of the file as constants outside the class (wherever you think it looks good, but should be easily visible).
> 
> Something like `const CHAIN_MIGRATE_VERSION = 2` and `CHAIN_TARGET_DB_VERSION=2` or something?
> 
> And while we're at it, in the CHANGELOG I think you should specify "users upgrading to this version of hsd must start for the first time with `--chain-migrate=2 --wallet-migrate=2`"

Those are hard coded because they occurred at that time in the migration history and the state they need to set that specific one. It does not have anything to do with the way we pass the proper number.  
For example, MigrateMigrations for chain will always be executed when dbVersion is 1 and it will increase to two. We could even have assert for that. Next `upgrade` migration will be for specific version as well, e.g. from 2 to 3.

Same for the migrations, MigrationMigrations sets 2 to skip the next one, that's a `hack` to skip next migration if previous state has already executed next migration and that's specific to the current situation only. Setting `nextMigration` manually should not normally happen, unless you are changing how migrations work. (that's what happened there as well)

as for showing proper number in the error, that I will do. Forgot to update that part of the code :)

pinheadmz:
OK! I think this baby is ready to land.

I tested by checking out v2.0.1 of hsd and syncing full and spv nodes. Then I used `rpc invalidateblock` to freeze the chain at height `78000` to check and save the chainstate info. I duplicated the data directories, upgraded hsd to this branch and tried various configurations to restart hsd.

Expected behaviors were all satisfied:
- full node had wrong chainstate before upgrade and migration, correct chainstate after
- full node requires `migrate` flag to continue, with correct value (currently `1`)
- starting pruned node without `prune` flag threw an error before the migration message
- starting pruend node with `prune` flag throws migration error, then with `migrate` flags logs warning on every boot
- spv node requires `migrate` flag to launch, then logs nothing (`there are 0 migrations`) after migration

pinheadmz (approved):
ACK f1f6de676c1e9c487f8bd6bcd938e35b66ee9c19

```
Hash: SHA256

ACK f1f6de676c1e9c487f8bd6bcd938e35b66ee9c19
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmEJV0wACgkQ5+KYS2KJ
yToqqg//Y4YnGS0oes4WJMJIM5Zk1+tOg9r0QjE1Z2uIbHX+G1Anh5j5GtsN4Rrg
R9ieUTa+0SxnnSrIi4ahg/IK9zuJFWfHTp7BYx4nzSQUT3ZZ6gSePwao34RcJhR3
BVbis4qXGDUpWBaGlIEmmdnGySXoCntwgjIx/oRSYy+TzwK4FhSNcIeqNrczdj9v
Oa3Z781RcQsO6B7jAeQPXqp00QqR3pwwZLuHKPxL5YeKx/DOqEgV+I4N+hDtpHh/
b05wKJnFfamDFrzEYiSp2RiAN1QlzHBQ1g7QNtmOOGYjd2+R7lmnth5RctOm0RZl
7kqCbEfiNcgyXa+o3mor1c+sATiFyzikQsdfPhJ8SnycG5tklar7QC6Z4x3JK/Q8
5rCqiKrdQWWw8zjRGvJ2OzKnHlgTtZSu+ZFd5i0sf1/kId8pGdW2WYocAFT50cRw
bu757IfTiiQv53ppOXx5lQ5RtUcJfghNn5XHZOSf0GDF8U+p6TuD9VhUtck35lmK
41yliXBdE8SWyB3VjikJaUhlePzyS/ME9fqKTG6M+zccYs11muWs1KKEndjXkz1u
DtFfXalPp8GuV3wStDzH7Nb3UyawkNUTokzXXI5AY/NVxBW7cWpgAlLj2NXNynJg
MrP8vzV6Wxx0gf3V1UHh//OeyP3WQ8MkH4ESH7leTAMv4jJmAI4=
=ZDQa
-----END PGP SIGNATURE-----
```

</details>

PR #617, opened 2021-07-09
wallet: enable re-claiming of reserved names using new protocol rules from recent soft fork

pinheadmz:
Background: https://handshake.org/notice/2020-04-02-Inflation-Bug-Disclosure.html

Relevant section:

> NEW PROTOCOL RULES
> A CLAIM must commit to a mainnet block hash. By default, all claims commit to block 1. The existing rules already require that subsequent re-claims commit to higher block heights. This makes it easy to determine in the software when a claim is being re-submitted (commitHeight > 1). The following new rules are enforced by the soft-fork code:
> 
> • An initial CLAIM MUST commit to block 1.
> • An initial CLAIM MUST pay a fee less than 1,000 HNS.
> • A re-claim MUST pay the exact same fee as the initial CLAIM.
> • The fees from any re-claim MUST NOT be collected by the miner: they are "burned" or, more accurately, already exist in a previous block.
> • A re-claim can not be mined until 288 blocks have passed since the last CLAIM or re-claim.

This PR modifies the wallet `makeClaim` path to enable this action. The wallet will have to ask the fullnode for the UTXO that currently owns the reserved name in LOCKED state to determine the correct values for the updated claim. That means that this can only be done with a full node (not SPV, but pruned is OK).

### Soft fork considerations

- A miner that has not upgraded and fails to follow the new protocol rules when including a re-claim in a block will produce an invalid block. The covert signaling mechanism from the soft fork deployment indicates that 88% of miners have upgraded, but I suspect it is actually 100%. We know some pools upgraded that are whatever reason not signaling in blocks:

```
--> hsd-rpc getcovertsignaling
{
  "blocks": 1772,
  "total": 2016,
  "perc": 0.878968253968254
}
```

- Non-upgraded relay nodes and Bob Wallets will interpret a valid re-claim in a valid block as generating new coins when in fact it is not. This will only affect that node's chainstate when reporting total chain value, no other functions are affected. The chainstate is fixed automatically by upgrading to v2.4.0 which migrates  on boot. According to my seed node, well over half the network has already upgraded:

```
  '/hnsd:1.0.0/' => 36,
  '/hsd:2.0.1/' => 1,
  '/hsd:2.0.2/' => 6,
  '/hsd:2.0.3/' => 1,
  '/hsd:2.1.1/' => 2,
  '/hsd:2.1.2/' => 1,
  '/hsd:2.1.3/' => 4,
  '/hsd:2.1.5/' => 26,
  '/hsd:2.2.0/' => 33,
  '/hsd:2.3.0/' => 31,
  '/hsd:2.4.0/' => 146,
  '/hsd:2.4.0/hmail.app/' => 1,
  '/hsd:2.4.0/sinpapeles/' => 2
```

TODO:
- [x] add test for `ownership.decodeData()`

### DIFF
--- a/lib/covenants/ownership.js
+++ b/lib/covenants/ownership.js
@@ -14,6 +14,7 @@ const blake2b = require('bcrypto/lib/blake2b');
 const StubResolver = require('bns/lib/resolver/stub');
 const BNSOwnership = require('bns/lib/ownership');
 const consensus = require('../protocol/consensus');
+const Network = require('../protocol/network');
 const reserved = require('./reserved');
 const {Proof: BNSProof} = BNSOwnership;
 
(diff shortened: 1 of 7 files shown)


pinheadmz:
Some user has already been using this branch on mainnet: https://hnsnetwork.com/address/hs1qkcu2rmhqwmwmw8ewf59g0lc4w6522zgx0llvev

178 re-claimed reserved names and counting! So I guess the code works OK, and there are no inflationary consequences.

PR #623, opened 2021-07-31
🎨 Update mempool.js

0xPoet:
nop function doesn't make anything. 

But check again...


0xPoet:
those are placeholders in mempool when you don't need to do anything for each ancestor for example just count ancestors

it could alternatively be `null`able and have if check there. But `nop`s make code cleaner in many cases

e.g. if logger.debug is called and debug.level is higher than debug it will also call nop and etc.

Thanks to: @nodech

PR #597, opened 2021-05-06
Fix and enable checkpoints for mainnet (WIP)

chjj:
Long overdue PR to enable checkpoints on mainnet. Currently untested.

PR #629, opened 2021-08-27
Increase test coverage for chain

pinheadmz:
- [x] Replace `TODO` with actual test for sigops limits
- [x] fix CSV test (see https://github.com/bcoin-org/bcoin/pull/613)
- [x] test checkpoints (see https://github.com/handshake-org/hsd/pull/597)
- [x] improve checkpoints test with airdrop test and coin supply audit


pinheadmz:
Added test for checkpoints. What we do is generate a chain with a variety of transactions (auctions, name claims, undependable outputs, etc) then add each block into a fresh chain object. The second time we do this, the fresh chain has a checkpoint just below the tip. Before rebasing on master branch which includes the checkpoints fix (#597) the test fails with this output:

```
  1) Checkpoints
       Sync with checkpoints
         should sync chain:

      Error: Sync failure at height 204 (register reserved): Database inconsistency.

      throw new Error(
            ^

      at Context.<anonymous> (/Users/matthewzipkin/Desktop/work/hsd/test/chain-checkpoints-test.js:246:17)
```

Note that the failure is labeled! During the initial chain-generation process, we also build an array of strings that label every single block! If sync fails, the label for the failed block is looked up and printed in the error message ;-)

test/chain-checkpoints-test.js
```diff
@@ -0,0 +1,281 @@
...
+    await mineBlock([register1, register2], null, null, 'registers');
+    await mineBlocks(10, 'after registers');
+  });
+
+  it('should confirm airdrop and faucet proofs', async () => {
+    const proof = AirdropProof.decode(rawProof);
+    const fproof = AirdropProof.decode(rawFaucetProof);
+
+    await mineBlock(null, null, [proof, fproof], 'airdrops');
+  });
```
pinheadmz, line 191: this part sucks

test/chain-test.js
```diff
@@ -694,79 +711,144 @@ describe('Chain', function() {
       'bad-txns-txouttotal-toolarge');
   });
 
-  it('should mine 111 multisig blocks', async () => {
-    const flags = common.flags.DEFAULT_FLAGS & ~common.flags.VERIFY_POW;
+  describe('Sigops', function() {
```
nodech, line 714: This test was somewhat hard to follow because of many numbers, maybe general comment would help.

To summarize: we have 80 blocks with 100 outputs and each outputs has 100 checksigs(sigop),
 but to construct test blocks (max and max + 1) we use 80 blocks but only 10 transactions from each so we get (80 * 10 * 100 = 80000 sigops)

Maybe just summary of the process and mention of 80k sigops being max in the beginning would help - or choosing simpler to follow numbers. I was confused when I got 800k sigops from in the generation block (80 block * 100 outputs * 100 sigops = 800k).

For simpler numbers, not sure. Maybe just generate 10 txs or reduce sigops to 10 instead. Maybe leave as it is just do above, add description.

nodech:
I noticed CSV tests are duplicated in `test/node-test.js` as well, I think all related tests for `nSequence` and `nLocktime` could be moved out from both `node-test` and `chain-test.js` into `chain-cltv-csv-test.js` or `chain-locktime-test.js`(I think this is more generic?)

nodech:
- Sigop comments
- More CSV test refactor

Other than these two, everything LGTM !
Great tests !!! :clap:

test/chain-test.js
```diff
@@ -364,89 +349,121 @@ describe('Chain', function() {
...
-    const csvBlock = await mineCSV(tx);
-
-    assert(await chain.add(csvBlock));
-
-    const csv = csvBlock.txs[1];
-
+    // Create a TX that spends the relative-timelocked output.
     const spend = new MTX();
+    spend.addTX(fund, 0);
+    spend.addOutput(await wallet.getReceive(), 9000);
```
nodech, line 423: you can remove `await` from here (and below). MemWallet is sync.

pinheadmz:
@nodech I cleaned up the nits you found and refactored the tests so all RPC tests are in the `node-rpc-test` file, leaving `node-test` empty and deleted it!

I also moved the CSV tests to a new file... for the old file the tests that check chain height and wallet balance. Well, you know those broke and you can probably guess how I fixed them 😬

nodech (approved):

Issue #400, opened 2020-03-24
DNS Resolver: "Couldn't get address for '.': not found" with +trace


pinheadmz:
Something goes wrong when adding the `trace` option to a query to the handshake resolver. I don't think it's a `bns` issue, but I'm not sure because `bin/dig.js` in `bns` doesn't support the `trace` option.

I think this is possibly related to how the hns resolver checks the Urkel tree first, then falls back to ICANN. Note how the resolver can't find any nameservers for the root zone here, which makes sense in the context of HNS:

```
--> dig . +trace

; <<>> DiG 9.14.6 <<>> . +trace
;; global options: +cmd
.			4661	IN	NS	.
.			4661	IN	RRSIG	NS 13 0 518400 20210324122748 20200323122748 60944 . excuNixdKfpaRUjpMCbkcWrKp8sRMVS/btzg/TvdTuQ2I+mcU1ROLCN9 Juf9grYf6L1HeW+7qF5B2KHE80i5ww==
couldn't get address for '.': not found
dig: couldn't get address for '.': no more
```

The resolver works fine without `+trace`:

```
--> dig crypto51.app TXT

; <<>> DiG 9.14.6 <<>> crypto51.app TXT
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 23727
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: 5cd08efc8eb88cb0 (echoed)
;; QUESTION SECTION:
;crypto51.app.                  IN      TXT

;; ANSWER SECTION:
crypto51.app.           300     IN      TXT     "brave-ledger-verification=2f0f1d9999addc94f42dcb69e94a0ecffb5c2506bb7f65dbaa716fd86248fdb3"
crypto51.app.           300     IN      TXT     "hns-claim:aakhjyfr2yt6fhk7mj4s77mdonvoqg6z2yzp4oblaeaaaaaaaaaklzaororjdpl6qzexi75h7ofhv447lownw5btzuxvs4ibaaaaautcqf3q"

;; SIG0 PSEUDOSECTION:
[...]

PR #630, opened 2021-08-31
Use synth address for root server NS

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/400

- [x] TODO: Needs tests

Also imports a fix from hnsd PR https://github.com/handshake-org/hnsd/pull/67 which returns empty SOA if `_synth` is queried by itself (encourages the recursive resolver to ask again with the full synth address)

Third commit adds newlines `\n\n` before DNS messages in the log. These messages are only printed at `--log-level=spam` anyway and improve readability a whole heckuva lot:

Screen Shot 2021-08-31 at 10 01 25 AM

### DIFF
--- a/lib/dns/server.js
+++ b/lib/dns/server.js
@@ -189,8 +189,8 @@ class RootServer extends DNSServer {
     });
 
     this.on('query', (req, res) => {
-      this.logMessage('DNS Request:', req);
-      this.logMessage('DNS Response:', res);
+      this.logMessage('\n\nDNS Request:', req);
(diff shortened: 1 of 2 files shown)


buffrr (approved):
I can confirm this PR makes dig +trace usable if hsd is configured as a system resolver. Using a _synth address for the root NS avoids adding an A record to the zone apex  👍 

We may want to clean up DNSSEC for _synth negative answers eventually but don't think it's a high priority for now since its mainly for correctness. LGTM.

PR #633, opened 2021-09-07
pkg: update dependencies

pinheadmz:
This PR updates the logging module used by hsd to include this (2 year old) PR: https://github.com/bcoin-org/blgr/pull/8

The effect is that log files will now be rotated out when they reach ~20 MB and only the 10 most recent files will be kept on disk. If your current `debug.log` file is over 20 MB (it likely is if you have a long running node) then on next launch with upgraded hsd, that file will be renamed with a timestamp and a new `debug.log` will be opened and written to.

Example:
```
--> ls -lahn  ~/.hsd/*.log
-rw-r--r--@ 1 501  20    72K Sep  7 13:13 /Users/matthewzipkin/.hsd/debug.log
-rw-r--r--@ 1 501  20    20M Sep  7 13:11 /Users/matthewzipkin/.hsd/debug_2021-09-07_17-11-51-907.log
```

This current behavior (before this update) hsd will truncate the log file on boot, deleting potentially a huge amount of data, and begin writing to the same file.

This update is important for another reason we encountered during testnet (can't seem to find an open issue for it though): long-running nodes with `log-level: spam` are at risk of filling their disk with log data! The current hsd log truncate function ONLY runs on boot, so if your node is running full time for a long time, that `debug.log` file might be seriously enormous.

I just checked two of my full time nodes, one of the `debug.log` files was over 3.0 GB!!


rithvikvibhu:
👍
Updated `blgr` and it renamed the existing `debug.log` (~32 MB) to `debug_2021-09-07_17-33-53-047.log` and started a new `debug.log`.

pinheadmz:
I added a `bsocks` update to this as well, our package-lock was holding us back by one bugfix from 2019 😬

flourishEmmanuel:
Nice one

PR #636, opened 2021-09-28
pkg: update bcfg

pinheadmz:
See https://github.com/bcoin-org/bcfg/pull/6

Instead of waiting for @chjj to merge, tag, release, and publish `bcfg` for this fix I'm going to try something new: I forked `bcfg` into `handshake-org` where I can merge, tag and release. @chjj recently also enabled my npm account to publish to the `@handshake-org` namespace, so that's where we can source small dependencies like this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,14 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@handshake-org/bcfg": {
+      "version": "0.1.7",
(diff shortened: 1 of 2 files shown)


pinheadmz:
Ok big problem with this: we need to switch everything to `require('@handshake-org/hsd')` this also means that `hs-client` will still use the old version which is going to start getting messy and annoying, unless we completely move everything to `@handshake-org` on npm.

PR #639, opened 2021-10-02
Wallet.fund() messes up existing inputs

Anunayj:
`wallet.fund()` used to mess up existing inputs for reasons mentioned below, this made it incredibly awkward to create transactions with custom inputs, this should fix those issues.

As I understand this, the issue was
1. The CoinSelector works by making a clone of the MTX and trying to add inputs to it till the output's value is met, and then returning those coins.
https://github.com/handshake-org/hsd/blob/484ede0b45c92249a4af81faf6579dcf1611edfc/lib/primitives/mtx.js#L1420-L1427

2. Because of how MTX cloning is done, the CoinView is not cloned, CoinView contained the input values, without which determining if a inputAmount is adequate is not possible.
https://github.com/handshake-org/hsd/blob/484ede0b45c92249a4af81faf6579dcf1611edfc/lib/primitives/mtx.js#L103-L109

3. So there is this "hack" in place which basically works by checking the prevout in the utxo set, and then "adding them back" and returning their list, but this would not always work cause the prevout could possibly be not in the utxo.
https://github.com/handshake-org/hsd/blob/484ede0b45c92249a4af81faf6579dcf1611edfc/lib/primitives/mtx.js#L1692-L1717

4. After which all the inputs of the MTX are wiped and the inputs from above are added "back"
https://github.com/handshake-org/hsd/blob/484ede0b45c92249a4af81faf6579dcf1611edfc/lib/primitives/mtx.js#L1126-L1133

~~This required changes to the current implementation of `interactive-name-swap` test since it used a "hack" which is no longer needed.~~
Backward compatibility with the previous hack is maintained for now.
Compatibility with Bob and Shakedex needs to be checked before merging, but is likely gonna be a non-issue.

### DIFF
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -158,11 +158,19 @@ class MTX extends TX {
   addCoin(coin) {
     assert(coin instanceof Coin, 'Cannot add non-coin.');
 
-    const input = Input.fromCoin(coin);
+    // Avoid duplicate inputs
+    for (const input of this.inputs) {
+      const {hash, index} = input.prevout;
+      if (hash.equals(coin.hash) && index === coin.index) {
+        this.view.addCoin(coin);
+        return input;
+      }
+    }
 
+    // Coin represents a new input.
+    const input = Input.fromCoin(coin);
     this.inputs.push(input);
     this.view.addCoin(coin);
-
     return input;
   }
 
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1146,11 +1146,13 @@ class MTX extends TX {
     // (pre-signed) inputs OR add new "preferred" inputs for covenants.
     // The logic below was moved here from CoinSelector.fund() in case
     // the old method was used and "preferred" coins are not yet in the view.
-    if (this.inputs.length > 0) {
-      const inputPrevoutKeys = this.inputs.map(input => input.prevout.toKey());
+    for (const input of this.inputs) {
+      const {prevout} = input;
+      if (this.view.hasEntry(prevout))
+        continue;
+
       for (const coin of coins) {
-        const key = coin.toKey();
-        if (inputPrevoutKeys.some(prevout => prevout.equals(key))) {
+        if (prevout.hash.equals(coin.hash) && prevout.index === coin.index) {
           this.view.addCoin(coin);
         }
       }
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1122,11 +1122,19 @@ class MTX extends TX {
     assert(options, 'Options are required.');
     assert(options.changeAddress, 'Change address is required.');
 
+    // Hack in place to ensure backward compataibility with previous hack
+    if (this.inputs.length > 0) {
+      const inputPrevoutKeys = this.inputs.map(input => input.prevout.toKey());
+      for (const coin of coins) {
+        const {hash, index} = coin;
+        const key = Outpoint.toKey(hash, index);
+        if ( inputPrevoutKeys.some(prevout => prevout.equals(key)) ) {
+          this.view.addCoin(coin);
+        }
+      }
+    }
     // Select necessary coins.
     const select = await this.selectCoins(coins, options);
-
-    // Make sure we empty the input array.
-    this.inputs.length = 0;
 
     // Add coins to transaction.
     for (const coin of select.chosen)
@@ -1424,6 +1432,7 @@ class CoinSelector {
    */
 
   constructor(tx, options) {
+    this.parent = tx;
     this.tx = tx.clone();
     this.coins = [];
     this.outputValue = 0;
@@ -1569,6 +1578,12 @@ class CoinSelector {
       for (let i = 0; i < this.tx.inputs.length; i++) {
         const {prevout} = this.tx.inputs[i];
         this.inputs.set(prevout.toKey(), i);
+        const coin = this.parent.view.getCoin(prevout);
+        // If the coin is not in the view, it's value cannot be known,
+        // therefore it can't be funded correctly.
+        if(!coin)
+          throw new Error('Could not resolve input coin value');
+        this.tx.view.addCoin(coin);
       }
     }
   }
@@ -1585,7 +1600,6 @@ class CoinSelector {
     this.chosen = [];
     this.change = 0;
     this.fee = CoinSelector.MIN_FEE;
-    this.tx.inputs.length = 0;
 
     switch (this.selection) {
       case 'all':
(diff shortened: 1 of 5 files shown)


Anunayj:
Will need the [following ](https://github.com/Anunayj/bob-wallet/tree/coinview-fix) [1] changes in bob wallet

[1] https://github.com/Anunayj/bob-wallet/tree/coinview-fix

lib/wallet/wallet.js
```diff
@@ -1839,7 +1838,7 @@ class Wallet extends EventEmitter {
     output.covenant.pushHash(nameHash);
     output.covenant.pushU32(height);
     output.covenant.pushHash(nonce);
-    reveal.addOutpoint(Outpoint.fromTX(bid, bidOuputIndex));
+    reveal.addCoin(bidCoin);
```
Anunayj, line 1841: These are technically not necessary because the Coin would attempted to be added to CoinView if it's already not there. But this is cleaner.

pinheadmz:
Have you looked in to this?

I wonder why we don't clone the `view` here, and how much simpler that will make your PR?

https://github.com/handshake-org/hsd/blob/eea15c2f2d6b3371eefb3bb7a93d08185bc9e25f/lib/primitives/mtx.js#L102-L113

For example you probbaly won't need to add `this.parent = tx` to `CoinSelector` consturctor

test/interactive-swap-test.js
```diff
@@ -208,11 +208,12 @@ describe('Interactive name swap', function() {
...
     // Bob should verify all the data in the MTX to ensure everything is valid,
     // but this is the minimum.
-    const input0 = mtx.input(0).clone(); // copy input with Alice's signature
-    const coinEntry = await node.chain.db.readCoin(input0.prevout);
+    const coinEntry = await node.chain.db.readCoin(mtx.input(0).prevout);
     assert(coinEntry); // ensures that coin exists and is still unspent
 
-    const coin = coinEntry.toCoin(input0.prevout);
+    const coin = coinEntry.toCoin(mtx.input(0).prevout);
+    mtx.view.addCoin(coin);
```
pinheadmz, line 215: Could this just be `mtx.addCoin(coin)` ?

test/interactive-swap-test.js
```diff
@@ -226,12 +227,8 @@ describe('Interactive name swap', function() {
     const changeAddress = await bob.changeAddress();
     const rate = await wdb.estimateFee();
     const coins = await bob.getSmartCoins();
-    // Add the external coin to the coin selector so we don't fail assertions
-    coins.push(coin);
+
```
pinheadmz, line 230: Can we remove this whole chunk now?

```js
    // Fund the TX.
    // The hsd wallet is not designed to handle partially-signed TXs
    // or coins from outside the wallet, so a little hacking is needed.
    const changeAddress = await bob.changeAddress();
    const rate = await wdb.estimateFee();
    const coins = await bob.getSmartCoins();
```

pinheadmz:
I'm going to continue to review this, but I think it would be helpful to have an explicit test case to show what the change is. i.e. a straightforward test that fails without the patch and succeeds with the patch. I know we have the atomic swap example working already but I think something simple in mtx-test.js or wallet-test.js would be helpful too.

lib/wallet/wallet.js
pinheadmz: So I think now with this logic: https://github.com/handshake-org/hsd/pull/639/commits/edc1f253b37aa228ee458ecb70d9b8d41fff39f0 using `addCoin()` is actually an optimization!

lib/primitives/mtx.js
```diff
@@ -100,15 +100,21 @@ class MTX extends TX {
...
 
   inject(mtx) {
     assert(mtx instanceof this.constructor);
     super.inject(mtx);
     this.changeIndex = mtx.changeIndex;
+
+    // CoinView has other properties like UndoCoins and BitView
+    // that an MTX doesn't care about. We just want a copy of the
+    // coins spent by the MTX.
+    for (const [key, value] of mtx.view.map.entries())
```
nodech, line 115: The reason behind not cloning the view is that in most cases it's not necessary and it's complex.
Also we have couple of variations on the view. Notably, the new `WalletCoinView` needs different approach to be properly cloned.

So in most cases `mtx` clone is done like:
```
const nmtx = mtx.clone();
nmtx.view = mtx.view;
```

Referencing the old view. It needs double checking if there's a case where old way messes things up, but I doubt. Most views in the wallet side are short lived.

lib/primitives/mtx.js
```diff
@@ -158,11 +164,19 @@ class MTX extends TX {
   addCoin(coin) {
     assert(coin instanceof Coin, 'Cannot add non-coin.');
 
-    const input = Input.fromCoin(coin);
+    // Avoid duplicate inputs
```
nodech, line 167: `addCoin`, `addInput`, `addOutpoint` and `addTX` are utilities that are meant to be as simple as possible.
They don't do anything extra, pretty dumb methods. So complicating addCoin here I don't think is necessary, as you can always use pretty cheap check: `isSane` or `checkSanity` after you are done with everything.

lib/primitives/mtx.js
```diff
@@ -158,11 +164,19 @@ class MTX extends TX {
...
+    for (const input of this.inputs) {
```
nodech, line 168: Another note is this is inefficient way of checking this `O(n^2)`. If we really wanted to do this(and as I mentioned above, we should not), we have to create new Map for the hash/index pair of each input similar to CoinView (Not coinview is not good case for this, as it may not have some coins)

lib/primitives/mtx.js
```diff
@@ -1122,12 +1136,31 @@ class MTX extends TX {
     assert(options, 'Options are required.');
     assert(options.changeAddress, 'Change address is required.');
 
+    // Ensure backward compatibility with the old API:
```
nodech, line 1139: Do we really want to remove old API? Other than incorrectly managing preferred inputs, was there another thing wrong with it?

As far as I can see, what we want to fix is - when `tx.clone` ing in the coinSelector, CoinSelector should use existing mtx.view as reference lookup table for these coins and don't expect them to be supplied with `coins` array that is coming from the database. That way "preferred" inputs that have coins wont get removed, but be checked against supplied coinview.

lib/primitives/mtx.js
```diff
@@ -1122,12 +1136,31 @@ class MTX extends TX {
...
+    // Inputs to MTX were added with mtx.addOutpoint() and then later
+    // CoinSelector.fund() was expected to match those outpoints with coins
+    // from the wallet. The problem with this is that when CoinSelector
+    // initialized, it wiped out all the inputs, including some that may have
+    // already been signed outside the wallet for custom scripts or CoinJoins.
+    // With the new API, mtx.addCoin() can be used to either fund existing
+    // (pre-signed) inputs OR add new "preferred" inputs for covenants.
+    // The logic below was moved here from CoinSelector.fund() in case
+    // the old method was used and "preferred" coins are not yet in the view.
+    for (const input of this.inputs) {
```
nodech, line 1149: This will become totally unnecessary, if we do the thing mentioned above.

Also Complexity of this will be off charts. We already have issues with the `coins` being too big for the big wallets, and iterating it for every input will just explode )

lib/primitives/mtx.js
```diff
@@ -1688,33 +1720,6 @@ class CoinSelector {
    */
 
   fund() {
-    // Ensure all preferred inputs first.
-    if (this.inputs.size > 0) {
-      const coins = [];
-
-      for (let i = 0; i < this.inputs.size; i++)
```
nodech, line 1695: Here instead of pushing blindly, we check if `view` has it, if it already has it we don't push this. Everything else can remain same.

lib/wallet/wallet.js
nodech: addCoin will still stay an optimization if we use my suggestion.

nodech (changes requested):

Anunayj:
Closing in favour of #666

PR #641, opened 2021-10-05
added getexplicitmempool rpc method

randomlogin:
Adding rpc method for full mempool retrieval. 
It affects internal presentation of blocks and txs, so I'm not sure it's good for the master.

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -174,6 +174,7 @@ class RPC extends RPCBase {
     this.add('getmempoolentry', this.getMempoolEntry);
     this.add('getmempoolinfo', this.getMempoolInfo);
     this.add('getrawmempool', this.getRawMempool);
+    this.add('getexplicitmempool', this.getExplicitMempool);
     this.add('gettxout', this.getTXOut);
     this.add('gettxoutsetinfo', this.getTXOutSetInfo);
     this.add('pruneblockchain', this.pruneBlockchain);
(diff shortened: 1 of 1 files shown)


nodech:
Mempool limits are set pretty high, so given decently sized mempool this request will be HUGE.

Also it's pretty easy to collect necessary mempool transactions with additional requests, if used with decent HTTP client (can reuse connection - keep-alive) you can easily replicate the same thing.

pinheadmz:
I agree with Nodar, the existing `getrawmempool` method along with individual TX requests should be enough for most users. @nodech I wanna point out that this patch has been open sourced along with https://github.com/handshake-labs/blockexplorer which is the hnsnetwork.com back end. I think we can close this since the @randomlogin even states its not suited for master branch.

However while playing with `getrawmempool true` I did notice a JSON encoding error:

https://github.com/handshake-org/hsd/blob/eea15c2f2d6b3371eefb3bb7a93d08185bc9e25f/lib/node/rpc.js#L3026-L3043

The list of TX dependents is not rendered as a string:

```
  "8d5dada918e50c4a3eb288d51da98577f26a31ee70ad8a286491ee85b7104ac0": {
    "size": 192,
    "fee": 0.0192,
    "modifiedfee": 0,
    "time": 1638469789,
    "height": 96647,
    "startingpriority": 0,
    "currentpriority": 0,
    "descendantcount": 0,
    "descendantsize": 192,
    "descendantfees": 19200,
    "ancestorcount": 14,
    "ancestorsize": 0,
    "ancestorfees": 0,
    "depends": [
      {
        "type": "Buffer",
        "data": [
          31,
          229,
          16,
          227,
          221,
          213,
          43,
          197,
          103,
          144,
          143,
          139,
          160,
          30,
          141,
          122,
          76,
          16,
          43,
          255,
          175,
          38,
          172,
          70,
          131,
          82,
          9,
          106,
          30,
          137,
          75,
          164
        ]
      }
    ]
  },
```

Issue #646, opened 2021-10-18
SPV nodes could use Urkel proof requests for RPC getnameinfo and getnameresource


pinheadmz:
Currently hsd in SPV mode returns null / empty JSON for RPCs `getnameinfo` and `getnameresource`. This is because even though it technically has an Urkel Tree, it is always empty. What we should probably do at the very least is return an error if these RPCs are called on an SPV node - an Error would be more helpful than empty / null.

But we could also go one step further and make a network request for the Urkel data (with proof) using the same mechanism as a DNS request. I started working on this, but it will require a network upgrade (I think).

What we can do right now is basically this:

```js
    let ns;
    if (this.chain.options.spv) {
      const data = await this.pool.resolve(nameHash);
      ns = NameState.decode(data);
    } else {
      ns = await this.chain.db.getNameState(nameHash);
    }
```

The problem with this is SPV can't verify any updates since the last tree interval.  `pool.resolve()` always uses the chain "safe root" which waits up to [12 confirmations](https://github.com/handshake-org/hsd/blob/27745128b3d8f5dd72b4cfbc933e14a1029dbe18/lib/blockchain/chain.js#L3548-L3574) after tree intervals for name resolution, and we could cut that down to 0 at least for these RPC calls, but the data is still potentially outdated. I think the expected UX for these RPCs is to get data directly from the chain anyway before Urkel updates so it might defeat the purpose.

We might be able to pack some more SPV proofs into this, i.e. the full node could use normal bitcoin-like SPV proofs to show us the recent transactions for a name along with the "old" urkel data, but that might be getting to complicated...

PR #647, opened 2021-10-18
rpc/pool: use urkel proofs for namestate in spv mode

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/646
Also see https://github.com/kyokan/bob-wallet/pull/409

Restores SPV node RPC functionality for `getnameinfo` `getnameresource` `verifymessagewithname` and `getnamebyhash`.

Each of these rpcs now accepts an extra boolean parameter `safe`. When `true`, the node gets historical data from the "safe height" which is **the last tree root commitment with more than 12 confirmations**. This is obviously historical data that any full node will have locally, but is also data that can be retrieved from the p2p network by an SPV node. This is the exact same mechanism used by SPV (and hnsd) for light client name resolution.

Example:
## Full Node
```
# get current info
hsd-rpc getnameinfo chaos

# get info committed at last tree interval
hsd-rpc getnameinfo chaos true
```

## SPV Node
```
# ALWAYS NULL
hsd-rpc getnameinfo chaos

# get info committed at last tree interval (gets proof from p2p network)
hsd-rpc getnameinfo chaos true
```

Here's the tradeoffs for using "safe height":
👍 SPV nodes can getnameinfo and getnameresource without relying on a centralized API
👎 This data may be outdated since it can only be verified with the most recent urkel tree root commitment 
👎 Getting Urkel proofs from the p2p network may be slow and the RPC may time out

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Node changes
 
+- RPCs `getnameinfo` `getnameresource` `verifymessagewithname` and `getnamebyhash`
+now accept an additional boolean parameter `safe` which will resolve the name from the Urkel
+tree at the last "safe height" (committed tree root with > 12 confirmations). SPV
+nodes can use this option and retrieve Urkel proofs from the p2p network to respond
+to these calls.
+
 - New RPC methods:
   - `decoderesource` like `decodescript` accepts hex string as input and returns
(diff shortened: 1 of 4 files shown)


pcfreak30:
This is working fine for my needs. I acknowledge the risks of cached data from this.

Issue #642, opened 2021-10-08
Consensus failure due to full disk


pinheadmz:
Syncing Bob Wallet on a Windows machine with limited disk space, I encountered a pretty bad failure state. The error in the log reported a bad block - throwing `bad-register-owner` which means that a REGISTER was being verified but was not spending the `ns.owner` output (like imagine an auction loser, instead of the winner, trying to REGISTER).

I believe this error was thrown because the `ns.owner` data failed to be written in the Urkel Tree. Then when `chain.js` attempted to verify the next block, `verifyCovenants()` failed because the data was missing.

This highlights an atomicity problem between blockchain and urkel tree - critical consensus data is stored in two places and a sync failure could mean peers get banned for serving invalid blocks.

What we should do is, if a disk-write fails, gracefully shut down the process. See https://github.com/bcoin-org/bcoin/issues/933

PR #650, opened 2021-10-21
Gracefully shut down node on critical errors like full disk

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/642
(although there may be more places where this extreme action is warranted)

This PR adds a new test utility: RAMDisk. This is a cute trick where we create a virtual storage device in RAM that is really small (1 MB) then run a full node inside it until we fill up the disk. I tried to ensure that even in test suite failures we don't end up with a bunch of tiny volumes still mounted, but it may take some more work to make sure everything is super clean.

TODO:
- [ ] Apply to MIgrateBlockstore and test
- [x] Either skip test on Windows or implement RAMDisk for Windows (currently CI only runs on Linux)
- [x] Remove logConsole from test nodes (I just wanna see how it runs on CI first)

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2172,6 +2172,10 @@ class Chain extends AsyncEmitter {
       await this.db.compactTree(entry);
       await this.syncTree();
       this.emit('tree compact end', entry.treeRoot, entry);
+    } catch(e) {
+      const error = new CriticalError(e.message);
+      this.emit('abort', error);
+      throw error;
     } finally {
       unlock();
     }
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -26,6 +26,7 @@ const Script = require('../script/script');
 const {VerifyError} = require('../protocol/errors');
 const {OwnershipProof} = require('../covenants/ownership');
 const AirdropProof = require('../primitives/airdropproof');
+const {CriticalError} = require('../errors');
 const thresholdStates = common.thresholdStates;
 const {states} = NameState;
 
@@ -996,8 +997,10 @@ class Chain extends AsyncEmitter {
       const ns = await view.getNameState(this.db, nameHash);
 
       if (ns.isNull()) {
-        if (!covenant.isClaim() && !covenant.isOpen())
-          throw new Error('Database inconsistency.');
+        if (!covenant.isClaim() && !covenant.isOpen()) {
+          this.emit('abort', 'Unexpected null NameState.');
+          throw new CriticalError('Database inconsistency.');
+        }
 
         const name = covenant.get(2);
         ns.set(name, height);
@@ -1893,7 +1896,12 @@ class Chain extends AsyncEmitter {
     }
 
     // Save block and connect inputs.
-    await this.db.save(entry, block, view);
+    try {
+      await this.db.save(entry, block, view);
+    } catch (e) {
+      this.emit('abort', e.message);
+      throw new CriticalError(e);
+    }
 
     // Expose the new state.
     this.tip = entry;
@@ -1952,7 +1960,12 @@ class Chain extends AsyncEmitter {
         entry.height, util.hex32(entry.version));
     }
 
-    await this.db.save(entry, block);
+    try {
+      await this.db.save(entry, block);
+    } catch (e) {
+      this.emit('abort', e.message);
+      throw new CriticalError(e);
+    }
 
     this.logger.warning('Heads up: Competing chain at height %d:'
       + ' tip-height=%d competitor-height=%d'
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -998,8 +998,9 @@ class Chain extends AsyncEmitter {
 
       if (ns.isNull()) {
         if (!covenant.isClaim() && !covenant.isOpen()) {
-          this.emit('abort', 'Unexpected null NameState.');
-          throw new CriticalError('Database inconsistency.');
+          const error = new CriticalError('Database inconsistency.');
+          this.emit('abort', error);
+          throw error;
         }
 
         const name = covenant.get(2);
@@ -1899,8 +1900,9 @@ class Chain extends AsyncEmitter {
     try {
       await this.db.save(entry, block, view);
     } catch (e) {
-      this.emit('abort', e.message);
-      throw new CriticalError(e);
+      const error = new CriticalError(e.message);
+      this.emit('abort', error);
+      throw error;
     }
 
     // Expose the new state.
@@ -1963,8 +1965,9 @@ class Chain extends AsyncEmitter {
     try {
       await this.db.save(entry, block);
     } catch (e) {
-      this.emit('abort', e.message);
-      throw new CriticalError(e);
+      const error = new CriticalError(e.message);
+      this.emit('abort', error);
+      throw error;
     }
 
     this.logger.warning('Heads up: Competing chain at height %d:'
--- a/lib/node/fullnode.js
+++ b/lib/node/fullnode.js
@@ -197,7 +197,7 @@ class FullNode extends Node {
   init() {
     // Bind to errors
     this.chain.on('error', err => this.error(err));
-    this.chain.on('abort', msg => this.abort(msg));
+    this.chain.on('abort', err => this.abort(err));
 
     this.mempool.on('error', err => this.error(err));
     this.pool.on('error', err => this.error(err));
@@ -349,23 +349,6 @@ class FullNode extends Node {
     this.logger.info('Node is closed.');
     this.emit('closed');
   }
-
-  /**
-   * Emergency shutdown.
-   * @param {String} msg - error message
-   * @returns {Promise}
-   */
-
-   async abort(msg) {
-      this.logger.error(`Critical error, shutting down: ${msg}`);
-      try {
-        this.emit('abort', msg);
-        await this.close();
-      } catch (e) {
-        this.logger.error(`Error occurred during shutdown: ${e.message}`);
-        process.exit(-2);
-      }
-   }
 
   /**
    * Rescan for any missed transactions.
--- a/lib/node/fullnode.js
+++ b/lib/node/fullnode.js
@@ -197,6 +197,17 @@ class FullNode extends Node {
   init() {
     // Bind to errors
     this.chain.on('error', err => this.error(err));
+    this.chain.on('abort', async (msg) => {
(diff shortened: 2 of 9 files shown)


pinheadmz:
This error should be emitted on out-of-bounds reorg on a pruning or tree-compacted node as well (see https://github.com/handshake-org/hsd/pull/669)

pinheadmz (commented):
Other questions:
- Should we cover WalletDB with these as well and test?
- Do we actually need to change anyhting in migration code? Since a failed migration will throw in `chain.open()` which will crash `FullNode.open()`

lib/blockchain/chain.js
```diff
@@ -935,8 +935,10 @@ class Chain extends AsyncEmitter {
       const ns = await view.getNameState(this.db, nameHash);
 
       if (ns.isNull()) {
-        if (!covenant.isClaim() && !covenant.isOpen())
+        if (!covenant.isClaim() && !covenant.isOpen()) {
+          this.emit('CRITICAL ERROR', 'Unexpected null NameState.');
           throw new Error('Database inconsistency.');
```
nodech, line 940: Maybe for the critical errors, using different Error class would be more useful in the future if we want to identify those.

e.g.

```js
class CriticalError extends Error {
  constructor(module, msg) {
    super();

    assert(typeof msg === 'string');

    this.type = 'CriticalError';
    this.module = module;
    this.message = msg;

    if (Error.captureStackTrace)
      Error.captureStackTrace(this, CriticalError);
  }
}

throw new CriticalError('chain', 'Database inconsistency');

```

lib/blockchain/chain.js
```diff
@@ -1891,7 +1898,12 @@ class Chain extends AsyncEmitter {
         entry.height, util.hex32(entry.version));
     }
 
-    await this.db.save(entry, block);
+    try {
+      await this.db.save(entry, block);
```
nodech, line 1902: Maybe add modifications to the `chaindb.commit`:
```diff
diff --git a/lib/blockchain/chaindb.js b/lib/blockchain/chaindb.js
index 4b3dfcdc..b09574e0 100644
--- a/lib/blockchain/chaindb.js
+++ b/lib/blockchain/chaindb.js
@@ -295,11 +295,24 @@ class ChainDB {
   }
 
   /**
-   * Commit current batch.
+   * try committing current batch.
    * @returns {Promise}
    */
 
   async commit() {
+    try {
+      return await this._commit();
+    } catch (e) {
+      throw new CriticalError(e.message);
+    }
+  }
+
+  /**
+   * Commit current batch.
+   * @returns {Promise}
+   */
+
+  async _commit() {
     assert(this.current);
     assert(this.pending);
 
```

lib/blockchain/chain.js
```diff
@@ -935,8 +935,10 @@ class Chain extends AsyncEmitter {
       const ns = await view.getNameState(this.db, nameHash);
 
       if (ns.isNull()) {
-        if (!covenant.isClaim() && !covenant.isOpen())
+        if (!covenant.isClaim() && !covenant.isOpen()) {
+          this.emit('CRITICAL ERROR', 'Unexpected null NameState.');
```
nodech, line 939: I would probably use `abort` for the event name. smaller and more precise.

lib/node/fullnode.js
```diff
@@ -194,6 +194,11 @@ class FullNode extends Node {
   init() {
     // Bind to errors
     this.chain.on('error', err => this.error(err));
+    this.chain.on('CRITICAL ERROR', (msg) => {
+      this.logger.warning(`Critical error, shutting down: ${msg}`);
```
nodech, line 198: Should not this be `logger.error` ?

lib/node/fullnode.js
```diff
@@ -194,6 +194,11 @@ class FullNode extends Node {
...
+      this.close();
```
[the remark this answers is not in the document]
nodech: We want the caller to continue throwing error, not wait for the close. So normal emitter is okay. What we need to change is how we handle `this.close` here.  This needs to handle the failure of the `close` so we don't get `unhandledRejection` which will abruptly shutdown the node. I believe we should do the same as what unhandledRejection would, but with better message.

So instead we can use normal promise methods:
```js
  this.close().catch((e) => {
    this.logger.error(`Failed closing nodes: ${e.message}`);
    process.exit(-2); // some specific code somewhere in the constants
  });
```

test/node-disk-fill-test.js
```diff
@@ -0,0 +1,125 @@
+/* eslint-env mocha */
+
+'use strict';
+
+const os = require('os');
+const assert = require('bsert');
+const FullNode = require('../lib/node/fullnode');
+const RAMDisk = require('./util/ramdisk');
```
nodech, line 8: Instead of using RAMDisk and making this test harder to run, we could just hijack each and every method that can fail for the expected errors. Thats more work, but we can start with simple chaindb.current.write monkeypatch.

test/node-disk-fill-test.js
```diff
@@ -0,0 +1,125 @@
...
+
+  afterEach(async () => {
+    if (node && node.opened)
+      await node.close();
+
+    if (disk && disk.path)
+      await disk.close();
+  });
+
+  process.on('SIGINT', async () => {
```
nodech, line 39: I am not sure why this is here. Also if we 100% need this, we need to clean up after this test is done, but I don't see reason for this.

lib/blockchain/chain.js
pinheadmz: Yeah this makes sense. I didn't add anything to chaindb because it doesnt extend EventEmitter. But using a new error class is the right way.

test/node-disk-fill-test.js
pinheadmz: This was a failsafe for cleaning up the ramdisk when writing the tests locally. We can remove it and the whole ramdisk thing

lib/blockchain/chain.js
pinheadmz: Where do you think `CriticalError` class should live? We don't have a global module for these, but this type of error could come from anywhere

lib/node/fullnode.js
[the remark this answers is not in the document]
pinheadmz: Great thanks will add.

lib/blockchain/chain.js
pinheadmz: i like it thanks

lib/node/fullnode.js
pinheadmz: yes thanks

lib/node/fullnode.js
```diff
@@ -194,6 +194,16 @@ class FullNode extends Node {
   init() {
     // Bind to errors
     this.chain.on('error', err => this.error(err));
+    this.chain.on('abort', (msg) => {
```
nodech, line 197: This needs to be `async` otherwise `this.close` would not work.
We make sure we don't get `unhandledRejection` by using `try/catch`.

also `await` before `this.close`.

lib/node/spvnode.js
```diff
@@ -129,6 +129,16 @@ class SPVNode extends Node {
   init() {
     // Bind to errors
     this.chain.on('error', err => this.error(err));
+    this.chain.on('abort', (msg) => {
```
nodech, line 132: same as fullnode

package.json
```diff
@@ -69,7 +69,7 @@
     "test-browser": "NODE_BACKEND=js bmocha --reporter spec test/*.js",
     "test-file": "bmocha --reporter spec",
     "test-file-browser": "NODE_BACKEND=js bmocha --reporter spec",
-    "test-ci": "nyc --reporter lcov bmocha -- --reporter spec test/*.js"
+    "test-ci": "HSD_TEST_SUDO=true nyc --reporter lcov bmocha -- --reporter spec test/*.js"
```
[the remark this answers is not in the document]
nodech: No longer necessary if we get rid of Ramdisk.

nodech (changes requested):

lib/node/fullnode.js
pinheadmz: I'm confused https://github.com/handshake-org/hsd/pull/650#discussion_r863638293

Can we add await to the listener if its not asycemitter?

nodech: yeah, we can. It's just if that promise failed (the async method itself) we would get unhandledRejection because there's no one there to catch those errors, that's bad behaviour. But if we make sure promise is handled nicely. E.g. if I rewrote this in this form:

```javascript
this.chain.on('abort', (msg) => {
  this.close.catch((e) => {
    // ...
    // process.exit(INTERNAL_ERROR);
  });
});
```

I think it becomes apparent why we can do this and why we absolutely need try/catch in there. `async/try/catch` is just nicer way to write same as above.

pinheadmz:
rebased on master

pinheadmz:
Added a new commit to throw `CriticalError` and abort the node if compactTree throws. I can reproduce the behavior:

- created a fixed 4MB volume on OSX
- start hsd --prefix=<that volume>
- run script that adds 180 name updates to Urkel Tree and 723 blocks, filling the volume
- execute `rpc compacttree`

The error is a bit unexpected but makes sense:
`ENOENT: no such file or directory, mkdir '/Volumes/test_4MB/regtest/tree~'`

On OSX at least, the `~` directory can not even be created. It looks like a failed `mkdir` doesn't throw but when we try to open that directory we throw, sure ok.

In my test setup, adding only 713 blocks instead of 723 leaves enough room on the volume for `compactTree` to succeed

pinheadmz:
The error makes more sense on Ubuntu, otherwise the same test (using a 4MB tmpfs volume)

```
[info] (chain) Compacting Urkel Tree...
[error] (node) Critical Error: ENOSPC: no space left on device, write
    at Chain.compactTree (/home/pinheadmz/Desktop/work/hsd/lib/blockchain/chain.js:2176:21)
    at async RPC.compactTree (/home/pinheadmz/Desktop/work/hsd/lib/node/rpc.js:1117:7)
```

pinheadmz:
Looking good on windows as well inside MSYS2:

```
[debug] (chain) Retargetting to: 000000000000091b8e0000000000000000000000000000000000000000000000 (0x1a091b8e).
[error] (node) Critical Error: IO error: E:\hsd\chain/000003.log: There is not enough space on the disk.


    at Chain.setBestChain (C:\msys64\home\pinheadmz\work\hsd\lib\blockchain\chain.js:1903:21)
```

nodech (approved):

PR #655, opened 2021-11-04
Add net tests and fix unknown packets

nodech:
- Move framer/parser tests from `protocol-test` to `net-test` and get rid of `protocol-test`.
 - Fix Unknown packet not having consistent internal type and add rawType alias for all other packets.
 - Add packets.decode tests to `net-test`.

Closes https://github.com/handshake-org/hsd/pull/250 and is continuation of the https://github.com/handshake-org/hsd/pull/277


pinheadmz:
Confirmed this patch fixes a critical network issue where unknown packets will cause a connection to be dropped:

### Host peer (running master, then PR branch):
```
hsd --no-dns --no-wallet --network=regtest --listen=true --log-level=spam --memory=true
```

### Attack peer
With this patch applied, sending unknown packet type right after p2p handshake is complete:
```diff
diff --git a/lib/net/peer.js b/lib/net/peer.js
index c63eed5ee..bfbd4fccb 100644
--- a/lib/net/peer.js
+++ b/lib/net/peer.js
@@ -499,6 +499,11 @@ class Peer extends EventEmitter {
         await this.wait(packetTypes.VERSION, Peer.HANDSHAKE_TIMEOUT);
 
       assert(this.version !== -1);
+
+      this.logger.warning('sending unknown packet...');
+      this.send(new packets.UnknownPacket(202, Buffer.alloc(32, 0x01)));
+      return;
+
     } else {
       // We're shy. Wait for an introduction.
       if (this.version === -1)
```

### Run:
```
hsd --no-dns --no-wallet --network=regtest --port=10000 --brontide-port=20000 --http-port=30000 --memory=true --only=127.0.0.1 --log-level=spam
```

### Results:

On master branch, the host peer drops the p2p connection with this error in the log:
```
[debug] (net) Error: Socket Error: ERR_ASSERTION (127.0.0.1:58592)
```
With this patch applied:
```
[warning] (net) Unknown packet: 202 (127.0.0.1:58597).
```
... and the connection remains intact!

PR #657, opened 2021-11-06
Wallet record tests

nodech:
This is a port of a single line fix from #250 and expands on that and adds tests.

Tasks:
 - [x] BlockMeta fix
 - [x] ChainState tests
 - [x] BlockMeta tests
 - [x] MapRecord tests
 - [x] TXRecord tests

### DIFF
--- a/lib/wallet/records.js
+++ b/lib/wallet/records.js
@@ -140,19 +140,6 @@ class BlockMeta extends bio.Struct {
   }
 
   /**
-   * Instantiate block meta from json object.
-   * @private
-   * @param {Object} json
-   */
-
-  fromJSON(json) {
-    this.hash = json.hash;
-    this.height = json.height;
-    this.time = json.time;
-    return this;
-  }
-
-  /**
    * Instantiate block meta from serialized tip data.
    * @private
    * @param {Buffer} data
@@ -181,7 +168,7 @@ class BlockMeta extends bio.Struct {
    */
 
   getSize() {
-    return 42;
+    return 40;
   }
 
   /**
@@ -197,13 +184,25 @@ class BlockMeta extends bio.Struct {
   }
 
   /**
+   * Instantiate block meta from json object.
+   * @param {Object} json
+   */
+
+  fromJSON(json) {
+    this.hash = util.parseHex(json.hash, 32);
+    this.height = json.height;
+    this.time = json.time;
+    return this;
+  }
+
+  /**
    * Convert the block meta to a more json-friendly object.
    * @returns {Object}
    */
 
   getJSON() {
     return {
-      hash: this.hash,
+      hash: this.hash.toString('hex'),
       height: this.height,
       time: this.time
     };
--- a/test/wallet-records-test.js
+++ b/test/wallet-records-test.js
@@ -1,0 +1,167 @@
+'use strict';
+
+const assert = require('bsert');
+const random = require('bcrypto/lib/random');
+const consensus = require('../lib/protocol/consensus');
+const Records = require('../lib/wallet/records');
+const ChainEntry = require('../lib/blockchain/chainentry');
+const {
+  ChainState,
+  BlockMeta
+} = Records;
+
+describe('Wallet Records', function() {
+  describe('ChainState', function() {
+    function check(state, expected) {
+      assert.strictEqual(state.startHeight, expected.startHeight);
+      assert.bufferEqual(state.startHash, expected.startHash);
+      assert.strictEqual(state.height, expected.height);
+      assert.strictEqual(state.marked, expected.marked);
+    };
+
+    function getRandomChainState(marked = false) {
+      return {
+        startHeight: random.randomInt(),
+        startHash: random.randomBytes(32),
+        height: random.randomInt(),
+        marked: marked
+      };
+    }
+    it('should encode/decode with defaults', () => {
+      const state = new ChainState();
+      const encoded = state.encode();
+      const decoded = ChainState.decode(encoded);
+
+      assert.deepStrictEqual(state, decoded);
+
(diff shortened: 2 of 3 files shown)


lib/wallet/records.js
```diff
@@ -181,7 +168,7 @@ class BlockMeta extends bio.Struct {
    */
 
   getSize() {
-    return 42;
+    return 40;
```
pinheadmz, line 171: Digging in to this a bit to make sure that we don't break any serialized data already in users' DB. Since we use `StaticWriter` for `encode()` probably we were just allocating 2 more bytes than we need every time we write one of these, but those two extra bytes are just in memory and nothing will be changed how we read/write to disk... is that right?

test/wallet-records-test.js
```diff
@@ -0,0 +1,425 @@
...
+      const wtx = TXRecord.fromTX(tx, block);
+      compareTXRecord(wtx, data);
+    });
+
+    it('should set and unset block', () => {
+      const {data, tx, block} = getRandomTXRecordData(true, true);
+      const wtx = TXRecord.fromTX(tx);
+
+      assert.strictEqual(wtx.getBlock(), null);
+      assert.strictEqual(wtx.getDepth(random.randomInt()), 0);
```
pinheadmz, line 335: Why'd you test with random here? It'll always be `0` right? Just a bit unusual for our test style.

test/wallet-records-test.js
```diff
@@ -0,0 +1,425 @@
...
+      const encoded = map.encode();
+      const decoded = MapRecord.decode(encoded);
+
+      assert.bufferEqual(encoded, Buffer.from('00'.repeat(4), 'hex'));
+      assert.strictEqual(map.wids.size, 0);
+      assert.strictEqual(decoded.wids.size, 0);
+    });
+
+    it('should encode/decode map', () => {
+      const rand = random.randomRange(1, 100);
```
pinheadmz, line 376: Same here, why not `const` an array of 10 hard coded ints? I reccomend [The Numbers](https://lostpedia.fandom.com/wiki/The_Numbers) :-P

lib/wallet/records.js
nodech: Actually, no. Serialization on this record is never called, all the usage and db entries related to this are partial. That's why it was there for so long. Otherwise any attempt to serialize BlockMeta would throw assertion error because of the offset/size not matching in the final `.write()` method for the serializer.

I am not 100% sure how I found it in the bcoin, most likely when working on bmultisig.

You can reproduce by setting getSize back to `42`, you will see `should encode/decode` of the `BlockMeta` test will fail with `Out of bounds write`.

test/wallet-records-test.js
nodech: The main purpose of these tests is to make sure serializations are correct, they don't correspond to the logic they are being used. So I think, random data better tests serialization than just using fixed value.

test/wallet-records-test.js
nodech: Here it does not really matter whether it's fixed or not lol. Still went with random number of entries in the map.
Next test is using fixed value of `20`.

lib/wallet/records.js
pinheadmz: oh I see, we only ever write to DB `BlockMeta.toHash()` or height property, never serialize the whole object. OK!

test/wallet-records-test.js
pinheadmz: OK

test/wallet-records-test.js
pinheadmz: okie doke.

pinheadmz (approved):
ACK 5f115324128df48d68dbde696664f91494656c07

```
Hash: SHA256

ACK 5f115324128df48d68dbde696664f91494656c07
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmGmK/EACgkQ5+KYS2KJ
yTq2wBAAyi6Y/j+rPs60gnIOkA+x9O/lnsMScWDb+nGlfyYNVvukikZTe7ymoWN7
hipNJSfvR9k0E6+PilLJh2rXxd/OrHi5QnmXgPjcDTIpSeLC6Y54H3Jqx9HtHfmn
4SaILB2dQGl5Rcj57EWPpAks/VdgqI+WGky9U6tkqAq/nOeH2RG+R0WBvoEnEvz9
jWOho78QsYa45eX2I+Yb5DPiYCt6AgaiA+qtOhokDw3Q8L3jD0BJGHhMwLPHbOI/
v0VfqUNCDIO55esUgje7kEPhAROaZHvGOuaeAzeTLlsebrvrubfNBaV/m+O6kib5
v8hfbhCmF87yMiDqGV65ot8AyVHP3UhMwncZh3QF/42NQgCbAXAJbCOdqBawuI4/
p4cie8s882dFEHkct2dBIyYyMMJ3GdAy+rPd0l1HAWzev0+S3det5DkmcfxoFgwF
scTU8bZfedHpFxFPRaHBunuxnrpYgs/IscO3cKZJ+ynwsPOO/E+1X66n0EFKYPm0
WBtYpSwY1tzc8w+GjBDSQsWzePT+9H5HQ0gVEqtmVMT40x0whRDPgIpRolH1PgMr
hKfVy7oa1CqpPfr3qlMWLxK8yCBHQVKraaJtYO5RkWmymlkpfDmNFO0OfPIsJUkm
pv80v+j/I2X2IhQAAyEBRz8i+iNDGOqA/9BmdnAHt5cA8iw9Ikc=
=V3h1
-----END PGP SIGNATURE-----
```

</details>

PR #658, opened 2021-11-11
Add unicode lint - allow only ASCII symbols

nodech:
### Motivation
Motivation for this PR is recent attack using unicode:
  - https://www.trojansource.codes/
  - https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574

Approach to address the issue varies, recent example:
 - https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html - where they go with deny-by-default for the specific unicode symbols.

Another unicode based issue: https://certitude.consulting/blog/en/invisible-backdoor/

We can be even more strict with our lints and this is example of only allowing ASCII symbols and have separate allowed lists that appear out of necessity. This is in case something similar is discovered with different set of codes.

### Details
Simplest form of the allowed list check:
```sh
grep $'[^\u0020-\u007e\t]' ./**/*.js | grep -v '^\./lib/hd/words/\(french\|italian\|japanese\|spanish\|chinese-traditional\|chinese-simplified\)\.js'
```

Simple form of the disallowed list check
```sh
grep $'[\u202A-\u202E\u2066-\u2069]' ./**/*.js
```

### Future considerations
There are still files that I had to ignore for this:
 - Mnemonic words - any future PRs that affect those files (there should not be any though) or adds new language support needs to go through different checks, maybe we can have `disallowed_list` checks for those.

Even though `script/unicode-lint.sh` is not ignored, if changes are made to this file it needs more careful review what is added to the allowed list or use older unicode-lint to lint new one before executing it.

### Other projects
This currently does not include other projects, only thing it checks is node_modules as of now. But we could potentially have this linter in separate repo and add to all other projects as well.

### Note
This is little bit hacky version that is pretty slow and we can probably do better!
**Any contributions or even different implementation all together are welcome.**

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -23,7 +23,7 @@ jobs:
 
     - name: Install
       run: sudo apt-get install -y libunbound-dev |
-           npm install nyc coveralls bslint @hns-dev/bsdoc
+           npm install nyc@15.1.0 coveralls@3.1.1 bslint@5.15.3 @hns-dev/bsdoc@1.0.0
 
     - name: Lint
       run: npm run lint-ci
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -26,7 +26,7 @@ jobs:
            npm install nyc coveralls bslint @hns-dev/bsdoc
 
     - name: Lint
-      run: npm run lint
+      run: npm run lint-ci
 
     - name: Build Docs
       run: npm run build-docs
--- a/scripts/unicode-lint.sh
+++ b/scripts/unicode-lint.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+filter_ignore() {
+  # Filter Mnemonic keywords
+  filtered=`echo "$1" | grep -v '^\./lib/hd/words/\(french\|italian\|japanese\|spanish\|chinese-traditional\|chinese-simplified\)\.js$'`
+
+  # filter cached files
+  filtered=`echo "$filtered" | grep -v '^./node_modules/.cache'`
+
+  # filter ci dependencies
+  filtered=`echo "$filtered" | grep -v '^./node_modules/\(bslint\|@hns-dev/bsdoc\)'`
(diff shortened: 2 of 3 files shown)


pinheadmz (commented):
fun fact: BSD `grep` on OSX doesn't work with this script, gotta install GNU `grep` with homebrew...

scripts/unicode-lint.sh
```diff
@@ -0,0 +1,59 @@
...
+  echo "Found blacklisted symbols"
+  echo "More info at https://github.com/handshake-org/hsd/pull/658"
+  echo "Files:"
+  for file in $blacklist_matches; do
+    echo "  $file"
+  done
+  exit 1
+fi
+
+# Check files only have whitelisted characters
```
pinheadmz, line 43: If the dangerous characters can be isolated easily like you have already, whats the benefit of the allowed characters check?

pinheadmz: Nevermind, just saw the second article about the javascript attack

nodech:
> fun fact: BSD `grep` on OSX doesn't work with this script, gotta install GNU `grep` with homebrew...

Oh, I will revert this back to `find + grep` then, or will check what needs to change. This is just much quicker.

.github/workflows/node.js.yml
```diff
@@ -26,7 +26,7 @@ jobs:
            npm install nyc coveralls bslint @hns-dev/bsdoc
```
nodech, line 26: We may want to use specific versions here, otherwise in the future CI lint for these may break. E.g. nyc or coveralls added new dep that has disallowed characters.

pinheadmz:
On both Ubuntu 20 and OSX the script is finding "not allowed symbols" in a ton of files... I'm trying to figure out why

pinheadmz:
> On both Ubuntu 20 and OSX the script is finding "not allowed symbols" in a ton of files... I'm trying to figure out why

Still getting these errors on both machines... I'm not sure how to proceed with this. Could be differences in shell versions or something? I'm starting to wonder if we should write a short program in C to scan all text files for unicode bytes...

pinheadmz:
OK I've got a great little C program for this:

https://github.com/pinheadmz/unicode-comb

It scans everything, including node modules! I've tested it by pasting one of the bidirectional characters into random files.

PR #664, opened 2021-11-26
Propagate the user agent from node to pool

Falci:


nodech:
I think it's okay for users to rewrite Agent, but I was wondering if for general cases we should add special option, so they can just annotate hsd version. E.g.
 `./hsd --agent-comment other-variation` - giving us version - `/hsd:3.0.1(other-variation)/`.

This way we don't lose information about original version. Ofc someone really wants to, they could use `agent` to totally rewrite and that's okay as well, but in general should give better alternative ?

nodech:
@Falci could you also use that as an alternative in `bob-wallet` ?

It could be: `/hsd:3.0.1 (bob:version)/`. I see you are also preserving version there, so I guess it will be ok.

Falci:
@nodech it sounds good. I'll work on that.

pinheadmz:
I agree with @nodech I use this commit for easyhandshake: https://github.com/pinheadmz/hsd/commit/8c4f397ee7baff4ba9d7568657f9308c667c32f6

lib/node/fullnode.js
```diff
@@ -94,6 +94,8 @@ class FullNode extends Node {
 
     // Pool needs access to the chain and mempool.
     this.pool = new Pool({
+      agent: this.config.str('agent'),
+      agentComment: this.config.str('agentComment'),
```
pinheadmz, line 98: Can we just add one?

lib/net/pool.js
```diff
@@ -4564,6 +4564,20 @@ class PoolOptions {
...
+      if (this.agent.endsWith('/')) {
+        agentComment = `${this.agent.slice(0, -1)}(${options.agentComment})/`;
+      } else {
+        agentComment = `${this.agent}(${options.agentComment})`;
+      }
+;
+      assert(agentComment.length <= 255);
+      this.agent = agentComment;
+    }
+
```
pinheadmz, line 4580: Many thoughts:
- assertion errors should report why they throw ("...must be string")
- comment should be appended to current (default) agent: `/hsd:v3.0.1/sinpapeles`
- total string should be checked for length and asserted with explanation if error
- why do we care if the ua comment ends with a slash>

pinheadmz (changes requested):

lib/node/fullnode.js
Falci: The PR was about the first. You guys asked for this extra.
Why both?

> Ofc someone really wants to, they could use agent to totally rewrite and that's okay as well, but in general this should give better alternative ?

lib/net/pool.js
Falci: This implementation follows @nodech's comment:

> `./hsd --agent-comment other-variation` - giving us version - `/hsd:3.0.1(other-variation)/`.

pinheadmz:
@nodech maybe we need to agree on how this should work. Here's what I'm thinking:

- only one new option, `--agent-comment` or `--agent` I don't care
- whatever that string is, gets appended to default hsd version: `/hsd:v3.0.1/<<<WHATEVER>>>`
- if `null` or empty sting, default agent continues to be: `/hsd:v3.0.1`

Users that want to replace the entire agent string can fork hsd, but IMO this is the best way to add extra strings like bob wallet version and still get stats about hsd version deployment

nodech:
> @nodech maybe we need to agree on how this should work. Here's what I'm thinking:
> 
>     * only one new option, `--agent-comment` or `--agent` I don't care
> 
>     * whatever that string is, gets appended to default hsd version: `/hsd:v3.0.1/<<<WHATEVER>>>`
> 
>     * if `null` or empty sting, default agent continues to be: `/hsd:v3.0.1/`
> 
> 
> Users that want to replace the entire agent string can fork hsd, but IMO this is the best way to add extra strings like bob wallet version and still get stats about hsd version deployment

That sounds good with one modification: `/hsd:v3.0.1/something-else:version/` and this will be in line with [BIP 14](https://github.com/bitcoin/bips/blob/master/bip-0014.mediawiki)

As for user agent comments, they are meant for something else.

pinheadmz:
Ok SGTM. So you want to remove any trailing slashes from the input string and always add one back in the final output? What about the other special characters mentioned in the BIP?

nodech:
Because we are accepting whole version in the latter part of `/hsd../passed-agent/`, we can't actually disallow `():` - they can be passed. Only thing we should disallow there is `/`, because user may want to pass something like: `mymodifications:v0.1.1 (Windows)`

rithvikvibhu:
Hey, just checking in if it's ready to merge. I think it's fine to have both `agent` and `agentComment` - most would prefer `agentComment` and when integrating with other apps (like fingertip) they can change `agent` to that app's name.

pinheadmz:
Closing this and moving discussion to updated PR https://github.com/handshake-org/hsd/pull/710
@nodech @rithvikvibhu @Falci would appreciate your review there.

PR #666, opened 2021-12-03
MTX.fund - Fixes

nodech:
This is alternative implementation of #639 that solves issue in the `CoinSelector`.

Separate `existing` and `preferred` inputs:
  - existing are the inputs that are present in the mtx.
  - preferred now refers to inputs passed by `options.inputs`.
  - Use `CoinView` to resolve `coins`.
 
Closes #639
Closes #901

### DIFF
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1415,6 +1415,7 @@ class CoinSelector {
 
   constructor(tx, options) {
     this.tx = tx.clone();
+    this.view = tx.view;
     this.coins = [];
     this.outputValue = 0;
     this.index = 0;
@@ -1682,14 +1683,29 @@ class CoinSelector {
       for (let i = 0; i < this.inputs.size; i++)
         coins.push(null);
 
-      for (const coin of this.coins) {
-        const {hash, index} = coin;
-        const key = Outpoint.toKey(hash, index);
-        const i = this.inputs.get(key);
+      // check coinview first
+      for (const [key, index] of this.inputs.entries()) {
+        const prevout = Outpoint.fromKey(key);
 
-        if (i != null) {
-          coins[i] = coin;
+        if (this.view.hasEntry(prevout)) {
+          const coinEntry = this.view.getEntry(prevout);
+
+          coins[index] = coinEntry.toCoin(prevout);
           this.inputs.delete(key);
+        }
+      }
+
+      // skip this if we resolved all inputs.
+      if (this.inputs.size > 0) {
+        for (const coin of this.coins) {
+          const {hash, index} = coin;
+          const key = Outpoint.toKey(hash, index);
+          const i = this.inputs.get(key);
+
+          if (i != null) {
+            coins[i] = coin;
+            this.inputs.delete(key);
+          }
         }
       }
 
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1115,8 +1115,8 @@ class MTX extends TX {
     // Select necessary coins.
     const select = await this.selectCoins(coins, options);
 
-    // Make sure we empty the input array.
-    this.inputs.length = 0;
+    for (const coin of select.viewCoins)
+      this.view.addCoin(coin);
 
     // Add coins to transaction.
     for (const coin of select.chosen)
@@ -1403,6 +1403,27 @@ class MTX extends TX {
 /**
  * Coin Selector
  * @alias module:primitives.CoinSelector
+ * @property {TX} tx - clone of the original mtx.
+ * @property {CoinView} view - reference to the original view.
+ * @property {Coin[]} coins - List of available coins.
+ * @property {Amount} outputValue - Total output value.
+ * @property {Coin[]} chosen - List of chosen coins to add.
+ * @property {Coin[]} viewCoins - List of coins to only apply to coinview.
+ * @property {Address} changeAddress - change address.
+ * @property {Amount} change - change value.
+ * @property {Amount} fee - Value of fee.
+ * @property {String} selection - selection type.
+ * @property {Boolean} subtractFee - whether to subtract fee from the output.
+ * @property {Number} subtractIndex - index of the output to subtract fee from.
+ * @property {Number} height - height of the chain. (to check spendability)
+ * @property {Number} depth - depth or confirmations for the coin.
+ * @property {Number} hardFee - fixed fee.
+ * @property {Number} rate - Rate of dollarydoo per kB.
+ * @property {Number} maxFee - maximum fee we are willing to pay.
+ * @property {Boolean} round - round to the nearest kilobyte.
+ * @property {Number} coinbaseMaturity - when do CBs become spendable.
+ * @property {Function} estimate - Input script size estimator.
+ * @property {Object[]} inputs - preferred inputs.
  */
 
 class CoinSelector {
@@ -1420,6 +1441,7 @@ class CoinSelector {
     this.outputValue = 0;
     this.index = 0;
     this.chosen = [];
+    this.viewCoins = [];
     this.change = 0;
     this.fee = CoinSelector.MIN_FEE;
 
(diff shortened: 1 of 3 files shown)


nodech:
Test is just repeat of the matrix, it can be rewritten and shortened as a function..

lib/primitives/mtx.js
```diff
@@ -1413,6 +1413,27 @@ class MTX extends TX {
 /**
  * Coin Selector
  * @alias module:primitives.CoinSelector
```
pinheadmz, line 1415: Nice job on the jsdocs properties -- do we actually need this alias?

http://hsd-dev.org/hsd/module-primitives.CoinSelector.html

I'm not sure why it needs to be listed as a primitive since the code is not in the `primitves/` directory...

http://hsd-dev.org/hsd/module-primitives.html

pinheadmz: Also we're not describing the `options` here but I see that is missing from many classes in hsd, so that should be a future project...

lib/primitives/mtx.js
```diff
@@ -1413,6 +1413,27 @@ class MTX extends TX {
...
+ * @property {Coin[]} chosen - List of chosen coins to add.
+ * @property {Coin[]} viewCoins - List of coins to only apply to coinview.
+ * @property {Address} changeAddress - change address.
+ * @property {Amount} change - change value.
+ * @property {Amount} fee - Value of fee.
+ * @property {String} selection - selection type.
+ * @property {Boolean} subtractFee - whether to subtract fee from the output.
+ * @property {Number} subtractIndex - index of the output to subtract fee from.
+ * @property {Number} height - height of the chain. (to check spendability)
+ * @property {Number} depth - depth or confirmations for the coin.
```
pinheadmz, line 1429:
```suggestion
 * @property {Number} depth - minimum confirmation depth of coins to spend.
```

? nit

test/mtx-test.js
```diff
@@ -62,4 +70,600 @@ describe('MTX', function() {
...
+        index: coin.index
+      }));
+    });
+
+    it('should fund with preferred inputs - view', async () => {
+      const mtx = new MTX();
+      const coin = dummyCoin(wallet1.getAddress(), 1000000);
+
+      mtx.addOutput(wallet2.getAddress(), 1500000);
+      mtx.view.addCoin(coin);
```
pinheadmz, line 154: Could also be `mtx.addCoin(coin)` right? I tried this and test passed -- is there any significant difference?

lib/primitives/mtx.js
```diff
@@ -1688,13 +1711,58 @@ class CoinSelector {
...
+  /**
+   * Resolve coins for existing inputs.
+   */
+
+  resolveInputs() {
+    if (this.inputs.size === 0)
+      return;
 
+    // Ensure we have coins for the existing inputs.
+    const view = this.tx.view;
```
pinheadmz, line 1746: This is confusing me a little bit because the constructor already assigns `this.tx = tx.clone();` and `this.view = tx.view;` Except for the last `for` loop (line 1782) you use `this.view` which is what I'd expect ...?

pinheadmz (commented):
Questions and comments below. One other thing I think we should test is some signatures / witness stacks. There's no `mtx.check()` in these tests and one of the important reasons for this patch is to enable the wallet to fund a MTX that already has a signed input. Use cases are coinjoins, atomic swaps, wallet plugins that handle fancy non standard scripts, etc. We already kinda have that in the atomic-swap test but I think it'd be prudent to include at least one external signature check in `mtx-test.js` as well.

lib/primitives/mtx.js
```diff
@@ -1413,6 +1413,27 @@ class MTX extends TX {
...
+ * @property {Number} subtractIndex - index of the output to subtract fee from.
+ * @property {Number} height - height of the chain. (to check spendability)
+ * @property {Number} depth - depth or confirmations for the coin.
+ * @property {Number} hardFee - fixed fee.
+ * @property {Number} rate - Rate of dollarydoo per kB.
+ * @property {Number} maxFee - maximum fee we are willing to pay.
+ * @property {Boolean} round - round to the nearest kilobyte.
+ * @property {Number} coinbaseMaturity - when do CBs become spendable.
+ * @property {Function} estimate - Input script size estimator.
+ * @property {Object[]} inputs - preferred inputs.
```
pinheadmz, line 1436: There's more confusion here because we have `inputs` which you describe as "preferred inputs".... but we also have object property `preferredInputs` 😬

pinheadmz: IIUC, `inputs` is actually the original tx existing inputs and `preferredInputs` are passed in as part of the options object

lib/primitives/mtx.js
```diff
@@ -1688,13 +1711,58 @@ class CoinSelector {
    */
 
   fund() {
-    // Ensure all preferred inputs first.
-    if (this.inputs.size > 0) {
-      const coins = [];
+    this.resolveInputs();
+    this.resolvePreferred();
```
pinheadmz, line 1715: Should we just call these `resolveExisting` and `resolvePreferred`? Try to be consistent with the two groups of inputs

pinheadmz:
Another thing I noticed is that in the worst case scenario, we will loop through `this.coins` twice and then iterate through a third time to fill the remaining inputs -- do you think it makes any sense to merge the existing & preferred loops? This will also get a lot better when we refactor txdb and not pull ALL COINS out for every tx...

lib/primitives/mtx.js
```diff
@@ -1705,33 +1773,62 @@ class CoinSelector {
...
+    if (this.inputs.size > 0)
+      throw new Error('Could not resolve existing inputs.');
 
-      for (const coin of coins) {
-        this.tx.addCoin(coin);
-        this.chosen.push(coin);
-      }
+    for (const coin of coins) {
+      view.addCoin(coin);
+      this.viewCoins.push(coin);
```
pinheadmz, line 1783: Why is this ending different than `resolvePreferred()`? Do we already have these coins in `chosen`?

```js
    for (const coin of coins) {
      this.tx.addCoin(coin);
      this.chosen.push(coin);
    }
```

lib/primitives/mtx.js
```diff
@@ -1688,13 +1711,58 @@ class CoinSelector {
...
-    if (this.inputs.size > 0) {
-      const coins = [];
+    this.resolveInputs();
+    this.resolvePreferred();
+
+    if (this.isFull())
+      return;
+
+    while (this.index < this.coins.length) {
+      const coin = this.coins[this.index++];
```
pinheadmz, line 1721: Is it possible that a coin would be added twice? If, for example, `this.coins[1]` was actually spent by one of the preferred or existing inputs, could that coin be mistakenly added a second time to fund the rest of the transaction? I actually am not even sure `mtx.check()` would catch a double-spend like this, we might need an explicit test

pinheadmz:
The anyone can renew test can also be updated:

https://github.com/handshake-org/hsd/blob/06aabfffe6a26b9c5c1417d80691a8e943a48f1e/test/anyone-can-renew-test.js#L309-L312

Gksvxc7up:
> This is alternative implementation of #639 that solves issue in the `CoinSelector`.
> 
> Separate `existing` and `preferred` inputs:
> 
> * existing are the inputs that are present in the mtx.
> * preferred now refers to inputs passed by `options.inputs`.
> * Use `CoinView` to resolve `coins`.
> 
> Closes #639

test/http-test.js

pinheadmz:
@nodech checking in on this?

lib/primitives/mtx.js
nodech: this.tx.addCoin will also create `inputs` in the tx. We don't want to do that here, because those inputs already exist in the `this.tx`. We are just updating view (Comment above)

test/mtx-test.js
nodech: Same as above. We want to make sure view is processed properly.

nodech:
One thing that users will have to consider with these changes is that, coinview should always be supplied with coin information when using `preferred inputs` and `existing inputs`. If you don't add coins to the coinview, the code will try to fill them in from the coin list it gets, but if the coin list is big that will hit the performance. - That feature is there to help users just in case.

As for double spend question: It's not job of the CoinSelector to resolve double spents that happen because of the `existing` and `preferred` inputs. The coinselector itself wont double spend the coins if they are correctly applied to the coinview.

Issue #644, opened 2021-10-09
Pruning: `keepblocks` value should cover a 15-day auction period


pinheadmz:
Pruning mode in hsd means only the last 288 blocks are kept on disk:

https://github.com/handshake-org/hsd/blob/d393c5a4ac4f65fc6230a37ba24ed1ee9dd26a02/lib/protocol/networks.js#L394-L398

We should consider bumping this up to one complete auction cycle so that "rescan auction" is still available in pruned mode.

nodech:
Does that mean it would be: `bidding period` + `reveal period` ?
```
That would mean:
  blocks per day - 144
  bidding period: 5 days - 144 * 5 - 720 blocks
  reveal period: 10 days - 144 * 10 - 1440 blocks
  total: 2160 blocks?
```

pinheadmz:
@nodech yup, seems like a ton huh. I guess pruning node operators should already know they can't rescan their wallet, and the `rescan auction` function is the only other thing that will break on pruned nodes. However, I *do* want Bob Wallet to compete with Namebase's UI which can show a user all auction activity at any arbitrary time, and I *also* want a pruning mode available in Bob to reduce the burden on full node users... I'm not sure what the compromise should be. Maybe we can reduce it to 5 days, just the bidding period. Rescanning an auction once REVEAL starts doesn't really do you any good if you're not already "watching" the auction.

PR #669, opened 2021-12-09
[consensus] Enable Urkel Tree compaction

### LINKED ISSUE #660: syncTree() could use one more log message
This function can take several seconds and should log something before it starts so impatient users don't think their process is borked:

https://github.com/handshake-org/hsd/blob/e8c9632c3c821831599a53167c5c2378fbbfcb32/lib/blockchain/chain.js#L124-L154

pinheadmz:
Closes #660

Adds new rpc method `compacttree` which deletes from disk an enormous amount of historical data in the Urkel Tree that is only necessary in the case that a chain reorganization occurs deeper than ~288 blocks. Note that this condition also applies to blockchain-pruning nodes and in the event of such a deep reorg, a pruned node will fail into an unrecoverable state. Pruning is available as a configuration setting or can be executed while running with rpc `pruneblockchain`.

### What does compacting the Urkel Tree mean?

Since the tree is append-only, when an existing name is updated (anything from a REVEAL to a REGISTER to an UPDATE) the old data still remains on disk. A new tree node is written and some file pointers are shuffled around to determine the new root hash of the tree (which is committed to in a block header).

If a chain reorg crosses back over a tree interval (36 blocks on mainnet) then an old tree root is pulled from the unzipped chain and used to revert the entire Urkel Tree to its old state, which is possible since all the historical data and pointers are still there on disk.

To compact the Urkel Tree means to take a given tree root and write a fresh Urkel Tree (in flat-file format on disk) containing ONLY the data connected directly to the root, then rename that directory to `~/.hsd/tree`, wiping out all the historical data that is no longer part of the current state.

### Strategy in this PR

Since compacting means the tree can no longer be reverted, it means that a chain reorg only one block deep that crosses a tree interval will permanently annihilate the full node. Therefore, what we must do is:

1. Use the historical data we still have on disk to rewind the Urkel Tree to an old state (~288 blocks ago / about 7-8 tree intervals ago)
2. Run the compacting process on that historical tree root, creating an unrecoverable reorg boundary two days in the past
3. Replay the blockchain back in to the Urkel Tree, all ~288 blocks, including all namestate updates crossing back over those 7-8 tree commitments until the Urkel Tree is back in sync with the chain

### End result

Whether or not a full node is pruning the blockchain, executing this RPC will apply the same restriction at the moment it is completed: a 288-block reorg is the end of the game. Of course UNLIKE a pruning node, this compacting process is NOT on-going, so over time the Urkel Tree will bloat back up again (for better or for worse) until the RPC is called again.

### Mainnet testing

I created a second branch of this PR with [WIP rpc dumpzone](https://github.com/handshake-org/hsd/pull/534) applied. I have hsd synced to height 97555 already and started it in an isolated mode:

`hsd --no-dns --no-wallet --only=127.0.0.1`

Then executed `hsd-rpc dumpzone hns-97555-full.zone` to get a snapshot of the Urkel Tree state.

Then executed `hsd-rpc compacttree` to run the pruning process.

The size of the directory `~/.hsd/tree` dropped from 6.8 GB down to ~500 MB 🥳  and the process took about 2.5 minutes total 🎉 

```
[I:2021-12-09T00:46:06Z] (net) Removed loader peer (127.0.0.1:12038).
[D:2021-12-09T00:46:06Z] (node-rpc) Handling RPC call: compacttree.
[D:2021-12-09T00:46:06Z] (node-rpc) []
[I:2021-12-09T00:46:06Z] (chain) Compacting Urkel Tree...
...
[I:2021-12-09T00:48:13Z] (chain) Synchronizing Tree with block history...
...
[I:2021-12-09T00:48:32Z] (chain) Synchronized Tree Root: 9d8eefb2af6c7ab640826f245040c693738ad7cd629b86f32ba040b3ad7a5f8c.
```

Finally, I ran `hsd-rpc dumpzone hns-97555-compacted.zone` and compared the two output zone files which matched exactly 🚀 

TODO:
- [x] add tests
- [x] consider edge cases like executing too frequently (?)
- [x] determine best deployment (should it be automatic?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,16 @@
 ## unreleased
 
 ### Node changes
-  - `FullNode` and `SPVNode` now accept the option `--agent` which adds a string
+ - `FullNode` and `SPVNode` now accept the option `--agent` which adds a string
   to the user-agent of the node (which will already contain hsd version) and is
   sent to peers in the version packet. Strings must not contain slashes and total
-  user-agent string must be 255 characters or less.
+  user-agent string must be 255 charact
 
-- `FullNode` parses new configuration option `--compact-tree` which will compact
-the Urkel Tree only when the node first opens. This is the preferred method
-because it will run before the node connects to the network and the processing
-time will not affect peers.
+  - `FullNode` parses new configuration option `--compact-tree` which will compact
+  the Urkel Tree when the node first opens, by deleting historical data. It will
+  keep up to the last 288 blocks worth of tree data on disk (7-8 tree intervals)
+  exposing the node to a similar deep reorganization vulnerability as a
+  chain-pruning node.
 
 ## v3.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,15 @@
 ### Node changes
  - `FullNode` and `SPVNode` now accept the option `--agent` which adds a string
   to the user-agent of the node (which will already contain hsd version) and is
-  sent to peers in the version packet. Strings must not contain slashes and total
-  user-agent string must be 255 charact
+  sent to peers in the version packet. Strings must not contain slashes and
+  total user-agent string must be less than 255 characters.
 
-  - `FullNode` parses new configuration option `--compact-tree` which will compact
-  the Urkel Tree when the node first opens, by deleting historical data. It will
-  keep up to the last 288 blocks worth of tree data on disk (7-8 tree intervals)
-  exposing the node to a similar deep reorganization vulnerability as a
-  chain-pruning node.
+  - `FullNode` parses new configuration option `--compact-tree-on-init` and
+  `--compact-tree-init-interval` which will compact the Urkel Tree when the node
+  first opens, by deleting historical data. It will try to compact it again
+  after `tree-init-interval` has passed. Compaction will keep up to the last 288
+  blocks worth of tree data on disk (7-8 tree intervals) exposing the node to a
+  similar deep reorganization vulnerability as a chain-pruning node.
 
 ## v3.0.0
 
@@ -32,8 +33,10 @@
 ### Wallet API changes
 
 - New RPC methods:
-  - `signmessagewithname`: Like `signmessage` but uses a name instead of an address. The owner's address will be used to sign the message.
-  - `verifymessagewithname`: Like `verifymessage` but uses a name instead of an address. The owner's address will be used to verify the message.
+  - `signmessagewithname`: Like `signmessage` but uses a name instead of an
+    address. The owner's address will be used to sign the message.
+  - `verifymessagewithname`: Like `verifymessage` but uses a name instead of an
+    address. The owner's address will be used to verify the message.
 
 - New wallet creation accepts parameter `language` to generate the mnemonic phrase.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## unreleased
 
+**When upgrading to this version of hsd you must pass
+`--chain-migrate=3` when you run it for the first time.**
+
 ### Node changes
  - `FullNode` and `SPVNode` now accept the option `--agent` which adds a string
   to the user-agent of the node (which will already contain hsd version) and is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   sent to peers in the version packet. Strings must not contain slashes and total
   user-agent string must be 255 characters or less.
 
+- `FullNode` parses new configuration option `--compact-tree` which will compact
+the Urkel Tree only when the node first opens. This is the preferred method
+because it will run before the node connects to the network and the processing
+time will not affect peers.
+
 ## v3.0.0
 
 **When upgrading to this version of hsd you must pass
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -146,7 +146,7 @@ class Chain extends AsyncEmitter {
 
     const {compactionHeight} = await this.db.getTreeState();
     const {compactTreeInitInterval} = this.options;
-    const compactFrom = compactionHeight + compactTreeInitInterval;
+    const compactFrom = compactionHeight + keepBlocks + compactTreeInitInterval;
 
     if (compactFrom > this.height) {
       this.logger.debug(
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -101,12 +101,26 @@ class Chain extends AsyncEmitter {
     this.setDeploymentState(state);
 
     if (!this.options.spv) {
-      if (this.options.compactTree)
-        await this.compactTree();
+      const {txStart} = this.network;
+      const startFrom = txStart + this.network.block.keepBlocks;
 
-    // Replay the blockchain from last committed tree root
-    // back up to the tip to rebuild tree and txn.
-      await this.syncTree();
+      if (this.options.compactTreeOnInit && this.height > startFrom) {
+        const {compactionHeight} = await this.db.getTreeState();
+        const {compactTreeInitInterval} = this.options;
+        const compactFrom = compactionHeight + compactTreeInitInterval;
+
+        if (compactFrom <= this.height) {
+          await this.compactTree();
+        } else {
+          this.logger.debug(
+            `Tree will compact at ${compactFrom} height.`);
+          await this.syncTree();
+        }
+      } else {
+        // Replay the blockchain from last committed tree root
+        // back up to the tip to rebuild tree and txn.
+        await this.syncTree();
+      }
     }
 
     this.logger.memory();
@@ -2118,17 +2132,28 @@ class Chain extends AsyncEmitter {
     const oldestBlock = this.height - this.network.block.keepBlocks;
 
     // Distance from that block to the start of the oldest tree interval.
-    const toNextInterval =
-      this.network.names.treeInterval -
-      (oldestBlock % this.network.names.treeInterval);
+    const toNextInterval = (this.network.names.treeInterval -
+      (oldestBlock % this.network.names.treeInterval))
+      % this.network.names.treeInterval;
 
     // Get the oldest Urkel Tree root state a pruning node can recover from.
     const oldestTreeIntervalStart = oldestBlock + toNextInterval + 1;
     const entry = await this.db.getEntryByHeight(oldestTreeIntervalStart);
 
     try {
+      // TODO: For RPC calls, If compaction fails while compacting
+      // and we never hit syncTree, we need to shut down the node
+      // so on restart chain can recover.
+      // Error can also happen in syncTree, but that means the DB
+      // is done for. (because restart would just retry syncTree.)
+      // It's fine on open, open throwing would just stop the node.
+
       // Rewind Urkel Tree and delete all historical state.
+      this.emit('tree compact start', entry.treeRoot, entry);
       await this.db.compactTree(entry);
+      this.emit('tree compact end', entry.treeRoot, entry);
+
+      await this.syncTree();
     } finally {
       unlock();
     }
@@ -3701,7 +3726,8 @@ class ChainOptions {
     this.maxOrphans = 20;
     this.checkpoints = true;
     this.chainMigrate = -1;
-    this.compactTree = false;
+    this.compactTreeOnInit = false;
+    this.compactTreeInitInterval = 10000;
 
     if (options)
       this.fromOptions(options);
@@ -3814,9 +3840,17 @@ class ChainOptions {
       this.chainMigrate = options.chainMigrate;
     }
 
-    if (options.compactTree != null) {
-      assert(typeof options.compactTree === 'boolean');
-      this.compactTree = options.compactTree;
+    if (options.compactTreeOnInit != null) {
+      assert(typeof options.compactTreeOnInit === 'boolean');
+      this.compactTreeOnInit = options.compactTreeOnInit;
+    }
+
+    if (options.compactTreeInitInterval != null) {
+      const {keepBlocks} = this.network.block;
+      assert(typeof options.compactTreeInitInterval === 'number');
+      assert(options.compactTreeInitInterval >= keepBlocks,
+        `compaction interval must not be smaller than ${keepBlocks}.`);
+      this.compactTreeInitInterval = options.compactTreeInitInterval;
     }
 
     if (this.spv || this.memory)
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -126,11 +126,35 @@ class Chain extends AsyncEmitter {
    */
 
   async syncTree() {
-    const {treeInterval} = this.network.names;
-    const last = this.height - (this.height % treeInterval);
-
     this.logger.info('Synchronizing Tree with block history...');
 
+    const {treeInterval} = this.network.names;
+
+    // Current state of the tree, freshly loaded from disk.
+    // It will be in the most recently-committed state,
+    // which was at the last tree interval. There might have been
+    // new blocks added to the chain since then.
+    const currentRoot = this.db.treeRoot();
+
+    // Get the block that corresponds to beginning of the last tree interval.
+    // The tree root in this block header will be the first block that
+    // commits to the new tree root hash, but the transactions it contains
+    // have not yet been added to the tree.
+    const last = this.height - (this.height % treeInterval);
+    const entry = await this.db.getEntryByHeight(last + 1);
+
+    // Require that tree is less than one tree interval behind the chain.
+    // We can skip this check only if the blockchain is less than
+    // one tree interval old, OR the chain is already at the tree interval,
+    // meaning the tree is already in sync with the chain and there is no delta.
+    if (entry) {
+      assert(entry.treeRoot.equals(currentRoot));
+    } else {
+      assert(last === 0 || last === this.height);
+    }
+
+    // Replay all blocks since the last tree interval to rebuild
+    // the `txn` which is the in-memory delta between tree interval commitments.
     for (let height = last + 1; height <= this.height; height++) {
       const entry = await this.db.getEntryByHeight(height);
       assert(entry);
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -130,9 +130,11 @@ class Chain extends AsyncEmitter {
 
     const {treeInterval} = this.network.names;
 
-    // Current state of the tree, freshly loaded from disk.
-    // It will be in the most recently-committed state,
-    // which was at the last tree interval. There might have been
+    // Current state of the tree, loaded from chain database and
+    // injected in chainDB.open(). It should be in the most
+    // recently-committed state, which should have been at the last
+    // tree interval. We might also need to recover from a
+    // failed compactTree() operation. Either way, there might have been
     // new blocks added to the chain since then.
     const currentRoot = this.db.treeRoot();
 
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -2128,7 +2128,7 @@ class Chain extends AsyncEmitter {
 
     try {
       // Rewind Urkel Tree and delete all historical state.
-      await this.db.compactTree(entry.treeRoot);
+      await this.db.compactTree(entry);
     } finally {
       unlock();
     }
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -148,8 +148,6 @@ class Chain extends AsyncEmitter {
   async syncTree() {
     this.logger.info('Synchronizing Tree with block history...');
 
-    const {treeInterval} = this.network.names;
-
     // Current state of the tree, loaded from chain database and
     // injected in chainDB.open(). It should be in the most
     // recently-committed state, which should have been at the last
@@ -158,38 +156,22 @@ class Chain extends AsyncEmitter {
     // new blocks added to the chain since then.
     const currentRoot = this.db.treeRoot();
 
-    // Get the block that corresponds to beginning of the last tree interval.
-    // The tree root in this block header will be the first block that
-    // commits to the new tree root hash, but the transactions it contains
-    // have not yet been added to the tree.
-    let last = this.height - (this.height % treeInterval);
-    let entry = await this.db.getEntryByHeight(last + 1);
+    // We store commit height for the tree in the tree state.
+    // commitHeight is the height of the block that committed tree root.
+    // Note that the block at commitHeight has different tree root.
+    const treeState = await this.db.getTreeState();
+    const {commitHeight} = treeState;
 
-    // Using the current tree root hash, rewind the blockchain to the beginning
-    // of the corresponding tree interval. If the tree has been compacted,
-    // this may be SEVERAL tree intervals behind the current chain tip
-    // (this is required to support chain reorgs to a limited depth,
-    // similar to pruning nodes keeping the last 288 blocks on disk).
-    // We can skip this only if the blockchain is less than
-    // one tree interval old, OR the chain is already at the tree interval,
-    // meaning the tree is already in sync with the chain and there is no delta.
-    if (!entry) {
-      assert(last === 0 || last === this.height);
-    } else {
-      for (;;) {
-        if (entry.treeRoot.equals(currentRoot))
-          break;
-
-        last -= treeInterval;
-        assert(last > this.height - this.network.block.keepBlocks);
-        entry = await this.db.getEntryByHeight(last + 1);
-        assert(entry);
-      }
+    // sanity check
+    if (commitHeight < this.height) {
+      const entry = await this.db.getEntryByHeight(commitHeight + 1);
+      assert(entry.treeRoot.equals(treeState.treeRoot));
+      assert(entry.treeRoot.equals(currentRoot));
     }
 
     // Replay all blocks since the last tree interval to rebuild
     // the `txn` which is the in-memory delta between tree interval commitments.
-    for (let height = last + 1; height <= this.height; height++) {
+    for (let height = commitHeight + 1; height <= this.height; height++) {
       const entry = await this.db.getEntryByHeight(height);
       assert(entry);
 
@@ -2106,11 +2088,13 @@ class Chain extends AsyncEmitter {
    * Compact the Urkel Tree.
    * Removes all historical state and all data not
    * linked directly to the provided root node hash.
-   * @param {Hash} root
    * @returns {Promise}
    */
 
   async compactTree() {
+    if (this.options.spv)
+      return;
+
     if (this.height < this.network.block.keepBlocks)
       throw new Error('Chain is too short to compact tree.');
 
@@ -2131,10 +2115,11 @@ class Chain extends AsyncEmitter {
     // Oldest block available to a pruning node.
     const oldestBlock = this.height - this.network.block.keepBlocks;
 
+    const {treeInterval} = this.network.names;
+
     // Distance from that block to the start of the oldest tree interval.
-    const toNextInterval = (this.network.names.treeInterval -
-      (oldestBlock % this.network.names.treeInterval))
-      % this.network.names.treeInterval;
+    const toNextInterval = (treeInterval - (oldestBlock % treeInterval))
+      % treeInterval;
 
     // Get the oldest Urkel Tree root state a pruning node can recover from.
     const oldestTreeIntervalStart = oldestBlock + toNextInterval + 1;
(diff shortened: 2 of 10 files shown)


pinheadmz:
Maybe include #675

pinheadmz:
I'm marking this as ready for review now by solving the last problem:

### What is the best method for deployment?

There are now two ways to do this:

1. `hsd-rpc compacttree` can be run at any time while the node is alive but is not recommended because the node is locked for several minutes and this is not great for your peers (who may even timeout if you take too long) 

2. `hsd --compact-tree` this argument can be added at launch on command line or left in the config file and the node will run the compact process when it launches, before the node connects to the network.

These options give node runners and integrated apps like Bob Wallet the flexibility to compact the tree either automatically or manually, either way with the node operator knowing when it will happen. I think this is a better method then setting some disk size threshold, etc which will run the compactor at unexpected times.

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # HSD Release Notes & Changelog
 
+## unreleased
+
+### Node changes
+
+- `FullNode` has new RPC command `compacttree` which reduces the Urkel Tree disk
+usage by deleting historical data. It will keep up to the last 288 blocks worth
+of tree data on disk (7-8 tree intervals) exposing the node to a similar deep
+reorganization vulnerability as a chain-pruning node.
```
brandondees, line 10: Great explanation here. 

I wonder if this suggests a want for a mechanism to detect a large re-org and warn the operator / bail out. I understand the likelihood of this occurring on mainnet is very small but if the worst case scenario is feasible maybe it's a nice emergency brake to have. Unless there is already something for that.

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+
+- `FullNode` parses new configuration option `--compact-tree` which will compact
+the Urkel Tree only when the node first opens. This is the preferred method
+because it will run before the node connects to the network and the processing
+time will not affect peers.
```
brandondees, line 15: Smart. Prune mode is for the block data and this is the urkel tree, so they're orthogonal to each other, right?

pinheadmz: Right, but obviously closely related, chain and tree rely on each other especially during reorgs. By keeping the tree-compactor to < prune amount we can always be able to prune the tree no matter what configuration the chain is in (confirmed by tests) and vice versa.

If you are NOT pruning (full archival chain) but have compacted the tree, you could theoretically recover from a >288-block reorg but I did not implement that, and I think it will require some kind of rebuild from genesis (using all data on disk)

CHANGELOG.md
pinheadmz: hsd will warn in the log when a block competitor has been discovered and again when a reorg actually happens. There are also events emitted so a serious node operator or complex application can react immediately. In the event that a reorg goes deeper than the prune depth (2 days / 288 blocks) there's not much warning to be had: the node just crashes, there's no other choice.

brandondees: thanks for clarifying, that seems adequate

lib/node/rpc.js
```diff
@@ -1104,6 +1105,27 @@ class RPC extends RPCBase {
     }
   }
 
+  async compactTree(args, help) {
+    if (help || args.length !== 0)
+      throw new RPCError(errs.MISC_ERROR, 'compacttree');
```
brandondees, line 1110: I'm not getting the meaning of this bit.

pinheadmz: Take a look at the other rpc functions. They all could use work but what this tells the user is the usage of the command very simply (accepts no arguments)

CHANGELOG.md
brandondees: I see. Is that 288 blocks configurable anywhere or is it hard-coded? Seems like it might come in handy to have the option, e.g. within tests.

brandondees (approved):
LGTM

Didn't read through all the tests extremely carefully but they do seem like good starting coverage and documentation overall.

Might be nice to add a little more log reporting about how well it's working when applied. Might help node operators with health monitoring or to detect efficiency regressions etc.

lib/node/rpc.js
brandondees: Ah ok, just seemed weird for it to be an exception with no explanation info in the message but knowing the convention it makes sense.

CHANGELOG.md
pinheadmz: In bcoin and hsd it's hard coded. Bitcoin core is more flexible but 288 is the minimum according to spec.

brandondees: minimum for pruning? there's no urkel tree equivalent in btc is there?

brandondees:
This looks like a great improvement, and I don't see any issues with the diff as written (take that with a grain of salt due to my unfamiliarity with the overall codebase and code conventions).

Another tangential question came to mind while reviewing: Are tree updates computed incrementally per block or is it an all-at-once tree update process once every interval? Is that something to consider further optimizations on in either direction?

CHANGELOG.md
pinheadmz: https://github.com/bitcoin/bips/blob/master/bip-0159.mediawiki

pinheadmz:
> Another tangential question came to mind while reviewing: Are tree updates computed incrementally per block or is it an all-at-once tree update process once every interval? Is that something to consider further optimizations on in either direction?

Changes to the Urkel Tree are stored in memory between tree intervals and written to disk every 36 blocks. This is an optimization and described in the white paper. If you shut down your node between tree intervals the data on disk will actually be behind. That's why we call syncTree() on startup to refill the tree data in memory from the block data on disk.

brandondees:
Gotcha, so there's already not much else to be done there unless you wanted to have hyper-fine configuration of what goes into memory vs disk at what times.

pinheadmz:
Worth pointing out that https://github.com/handshake-org/hsd/pull/669/commits/0ddd228d8c1a09ce2b9356757198fd87d1060013 makes this PR NOT backwards compatible. After running on my own local hsd node and compacting the tree, I then switched back to master branch to work on something else and got `Database Inconsistency` error.

pinheadmz:
TODO: remove RPC command

pinheadmz:
@nodech I removed the RPC `compacttree` and cleaned up the tests.

nodech:
## General
  The first thing I was thinking about is, whether we are still Full Node after
this. Do we want to only support this with Pruned nodes? We could of course
have two options for this pruned blocks and pruned chain. Even if for full node
it is possible to unprune the node by reindexing the chian.  

  So I would say, initially support this for prunes only and if someone
knows what they are doing they could use `--force-tree-compact` that can
be used for full nodes as well?  

  I think, it's better to support `--force-tree-compact` for full node
compaction, only after when we have ability to reindex the tree.

## Atomicity and `layout.s`
  The chaindb and tree are separate trees and we need to have them synchronized
logically. Only way to control where we are at in the Tree Hash history is
the `inject`. Last tree root hash is not useful information by itself. It needs
to come from the chaindb.  

  Issues `layout.s` solves comes when we crash during `connect`/`disonnect`.
`layout.s` is the only way we synchronize right now. Note, it's not useful
to use `tip` block hash, because tree is only committed once in every 36
blocks, unless you go to the last commit height and grab that block.
Let's go through issues that can occur and how layout.s solves.  

  Let's consider `connect` failing. What connect does is: first it commits
data to the Tree and then we atomically add all changes to the `chaindb`
and also write `layout.s` as new tree root. If for some reason node
crashes (or gets shut down) between `Tree.commit` and `chaindb.save`
we just end up with the "extra" data in the tree, but there will be
no inconsistency when node is rerun. Because `layout.s` will properly
inject previous `Tree.hash` instead of the `last` one (that one was
not saved in chaindb because of failure) and try to connect the same
block again. Because writes to the `chaindb.leveldb` are atomic,
once we have written succesfully, we are assured there's no inconsistency
between tree and the chain. It's also good to note, that this would
not work in `chaindb.save` -> `tree.commit` order, because in case of
failure we would not have tree data, even though chaindb would think
it's time to move on (on next rerun).  

  This problem becomes more exaggerated when you consider `reset` or `reorg`.
When reorg happens, chaindb will revert its state to the past and then move
forward to the new chain. If node crashes/gets shut down after it's reverted to
the past, `layout.s` is the only way to tell `tree` that we are actualy in the
older tree and latest `tree.rootHash` is not relevant.  

  If you consider someone restarting node with `--compact-tree` after reverting
to the past (on reorg) you can easily see, why should compact tree also depend
ony `laoyut.s` instead of `tree.rootHash`.  

  So to sum up, tree.rootHash is never relevant for atomicity/consistency
between chaindb and tree. We always trust chaindb to provide the tree.rootHash.
**chaindb root hash can be behind tree root hash, but NEVER ahead**.  

  Another note: The issue with blockstore was similar in a way, but it would
the node not starting and would fail to sync. This issue on the other hand,
wont get noticed because it wont result to crash. It will just have wrong
information in the tree.

## Atomicity with the compaction.
  The issue is related to above mentioned chaindb rootHash(`layout.s`).
We need to only rely on chaindb to give us information about tree situation.  

  In order to keep compacting code working with `layout.s`, we need to
move that `layout.s` pointer to the past FIRST and then compact. In this
case we make sure, if compaction fails, syncTree will just redo the same
thing and in the worst case end up with wasted space. We can't compact first,
because if it succeeds and then chaindb write fails, chaindb will be AHEAD.  

  syncTree needs to use `layout.s` instead as well. As mentioned above,
tree hash root should never be used for anything related to the chaindb.
Because of the `syncTree` nature, it's okay even if compaction failed
and we restart node without compact flag nd chaindb is behind. syncTree
will update chaindb from `saveNames` until it's caught up.  

  We can continue above mentioned example to show the issue of not using
`layout.s`. If chaindb reorg/reset went in the past, but crashed/closed
syncTree will actually be AHEAD and will continue appending to the wrong
tree hash.

## To summurize
  - chaindb root hash can be behind tree root hash, but NEVER ahead.
  - within compaction we need to move chain root hash(layout.s) behind even
    if we fail to compact and never run it again.
  - syncTree should use chain root hash (layout.s)

## Nits
  You can move out `syncTree` from `compactTree` and have it like this.
```javascript
  if (!this.options.spv) {
    if (this.options.compactTree)
      await this.compactTree();
      
    await this.syncTree();
  }
```
a

pinheadmz:
> You can move out `syncTree` from `compactTree` and have it like this.

This will call `syncTree()` an unnecessary second time if the option is set. (`compactTree()` calls `syncTree()` automatically already)

But I fixed that with an `else` in d6563351f0e37a1ca35f5164663e0222ce484346

pinheadmz:
@nodech thank you so much for your brilliant review and great catch regarding `layout.s` I rebased the branch to revert the layout.s deprecation and added in an extra commit to that value in `compactTree()` that is in commit 6cfd96b87133422f5ac6bed976ce7bd19177dc84

I also added a test in commit 80baf3441f4e66ec53b3062049d2272d61e5a9d0 to cover the "failed connect()" case as you describe. You were right, the code totally failed that test until `layout.s` was restored.

pinheadmz:
I think next steps with this branch conceptually (and maybe more tests) are around reorgs, and network services. As discussed, we might want to limit Urkel root search and send a polite reject message to light clients if we can't find their tree root.

nodech (changes requested):
This got me wondering if we should change pruned mode. instead of just storing
288(keepBlocks) we store `keepBlocks + treeInterval` which for main network will be `288 + 36`.

This will allow tree/chain to always have at least 288 blocks/tree roots.


Right now, we can only say that we have 252 blocks for sure (keepblocks - treeInterval) that can be reverted w/o issues.

pinheadmz:
> This got me wondering if we should change pruned mode. instead of just storing 288(keepBlocks) we store `keepBlocks + treeInterval` which for main network will be `288 + 36`.
> 
> This will allow tree/chain to always have at least 288 blocks/tree roots.
> 
> Right now, we can only say that we have 252 blocks for sure (keepblocks - treeInterval) that can be reverted w/o issues.

see https://github.com/handshake-org/hsd/issues/644 -- it might make more sense to keep even more than that.

As it relates to compact tree though I don't think we need to change it. It means that compacted tree will EITHER be 7 or 8 tree intervals behind. So worst case scenario is a 252-block reorg that bricks your node. A regular pruned node fails at 288-block reorg... Either case is so bad theres almost no difference.

test/chain-tree-compaction-test.js
```diff
@@ -0,0 +1,581 @@
...
+          if (prune && i <= chain.height - network.block.keepBlocks)
+            assert.strictEqual(block, null);
+          else
+            assert(block);
+        }
+      });
+    });
+  }
+
+  describe('SPV', function() {
```
nodech, line 480: You can move out `prefix` for easy cleanup:
```javascript
    let prefix;
    beforeEach(() => {
      prefix = path.join(
        os.tmpdir(),
        `hsd-tree-compacting-test-${Date.now()}`
      );
    });

    afterEach(async () => {
      await fs.rimraf(prefix);
    });
```

test/chain-tree-compaction-test.js
```diff
@@ -0,0 +1,581 @@
...
+        prune: true // also ignored
+      });
+
+      await node.ensure();
+      await node.open();
+      await node.close();
+    });
+  });
+
+  describe('Full Node', function() {
```
nodech, line 501: Same as in SPV.

lib/blockchain/chain.js
```diff
@@ -126,11 +130,48 @@ class Chain extends AsyncEmitter {
...
+    // one tree interval old, OR the chain is already at the tree interval,
+    // meaning the tree is already in sync with the chain and there is no delta.
+    if (!entry) {
+      assert(last === 0 || last === this.height);
+    } else {
+      for (;;) {
+        if (entry.treeRoot.equals(currentRoot))
+          break;
+
+        last -= treeInterval;
```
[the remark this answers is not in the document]
pinheadmz: Is that not covered already?

```js

        entry = await this.db.getEntryByHeight(last + 1);
        assert(entry);
```

test/chain-tree-compaction-test.js
pinheadmz: yep i like it thanks, 5420112e6c7cd7faf14f3f9e446f8360ef83cc6b

test/chain-tree-compaction-test.js
pinheadmz: 5420112e6c7cd7faf14f3f9e446f8360ef83cc6b

lib/blockchain/chain.js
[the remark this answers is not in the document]
nodech: yeah, I am just talking about paranoid checks )) to triple check

lib/blockchain/chain.js
```diff
@@ -100,8 +100,12 @@ class Chain extends AsyncEmitter {
 
     this.setDeploymentState(state);
 
-    if (!this.options.spv)
-      await this.syncTree();
+    if (!this.options.spv) {
+      if (this.options.compactTree)
+        await this.compactTree();
```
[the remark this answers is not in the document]
pinheadmz: OK, I like this too. Had to modify tests to manually call `syncTree()` in some places but the separation makes sense. And since we removed the rpc call especially since this only happens in `open()` now.

lib/blockchain/chain.js
```diff
@@ -126,11 +130,48 @@ class Chain extends AsyncEmitter {
...
-    const last = this.height - (this.height % treeInterval);
 
-    this.logger.info('Synchronizing Tree with block history...');
+    // Current state of the tree, loaded from chain database and
+    // injected in chainDB.open(). It should be in the most
+    // recently-committed state, which should have been at the last
+    // tree interval. We might also need to recover from a
+    // failed compactTree() operation. Either way, there might have been
+    // new blocks added to the chain since then.
+    const currentRoot = this.db.treeRoot();
```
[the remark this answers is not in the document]
pinheadmz: Is this still a concern? `syncTree()` is only called in `chain.open()` now, and `layout.s` will be injected during `chain.db.open()`

lib/blockchain/chain.js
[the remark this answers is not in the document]
pinheadmz: Hm, seems covered to me, but what dya have in mind?

lib/blockchain/chain.js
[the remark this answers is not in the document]
nodech: yep!

lib/blockchain/chain.js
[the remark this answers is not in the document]
nodech: Something simple, like
```js
        assert(last > this.height - this.network.block.keepBlocks);
```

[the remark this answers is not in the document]
pinheadmz: Ok, added. I still think its redundant but you are so handsome!

pinheadmz:
This feature has crept in size and now I'm afraid of how much integration modes we need to test.

For each Full, Prune, SPV:
- test migration which turns treeRoot into treeState in DB
- compact on init and after interval
- compact with RPC
- reconstruct tree

Maybe even on all 3 platforms too - when we did the blockstore migration I ran a lot of test on mainnet nodes and caught a few unexpected behaviors.

lib/blockchain/chain.js
```diff
@@ -158,38 +156,22 @@ class Chain extends AsyncEmitter {
...
-        last -= treeInterval;
-        assert(last > this.height - this.network.block.keepBlocks);
-        entry = await this.db.getEntryByHeight(last + 1);
-        assert(entry);
-      }
+    // sanity check
+    if (commitHeight < this.height) {
+      const entry = await this.db.getEntryByHeight(commitHeight + 1);
+      assert(entry.treeRoot.equals(treeState.treeRoot));
+      assert(entry.treeRoot.equals(currentRoot));
```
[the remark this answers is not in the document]
nodech: This only checks `if (commitHeight < this.height)`. `currentRoot` is actually set from the `this.height` entry, this just makes sure that entry to interval is the same as currentRoot.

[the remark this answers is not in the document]
nodech: okay, scratch that.
Only reason this sanity check is here, because it cross-checks with the block information.

lib/node/rpc.js
```diff
@@ -1119,6 +1120,27 @@ class RPC extends RPCBase {
     }
   }
 
+  async reconstructTree(args, help) {
+    if (help || args.length !== 0)
+      throw new RPCError(errs.MISC_ERROR, 'reconstructtree');
```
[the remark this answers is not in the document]
nodech: This follows the similar process to `compactTree` and `pruneBlockchain`..

This is slower indeed, but still hopefully you never need this feature.

test/chain-migration-test.js
```diff
@@ -808,6 +821,161 @@ describe('Chain Migrations', function() {
...
+      let error;
+      try {
+        await chain.open();
+      } catch (e) {
+        error = e;
+      }
+
+      // Now our error should be incorrect tree (after migration)
+      assert(error, 'Chain must throw an error.');
+      assert.strictEqual(error.message, `Missing node: ${root.toString('hex')}.`);
```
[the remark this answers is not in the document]
nodech: No it is not, not sure why I wrote this way. But this one makes sure the "fake root" filled with `0x01` that does not exists is properly migrated. So chain should fail to inject incorrect root even after migration.

lib/blockchain/chain.js
```diff
@@ -2118,17 +2132,28 @@ class Chain extends AsyncEmitter {
     const oldestBlock = this.height - this.network.block.keepBlocks;
 
     // Distance from that block to the start of the oldest tree interval.
-    const toNextInterval =
-      this.network.names.treeInterval -
-      (oldestBlock % this.network.names.treeInterval);
+    const toNextInterval = (this.network.names.treeInterval -
+      (oldestBlock % this.network.names.treeInterval))
+      % this.network.names.treeInterval;
```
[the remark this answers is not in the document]
nodech: yep.

test/chain-tree-compaction-test.js
```diff
@@ -517,28 +540,150 @@ describe('Tree Compacting', function() {
...
+
+      assert.strictEqual(startEvent.length, 1);
+      assert.strictEqual(endEvent.length, 1);
+
+      const [rootHash, entry] = endEvent[0].values;
+      // We don't have anything in the tree.
+      assert.bufferEqual(rootHash, Buffer.alloc(32, 0x00));
+      // 100 - 40 = 60
+      // 60 % 5 = 0
+      // So nearest one is 61.
```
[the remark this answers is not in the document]
nodech: It was `keepBlocks`

test/chain-tree-compaction-test.js
```diff
@@ -113,15 +114,43 @@ describe('Tree Compacting', function() {
...
       }
 
-      let name, nameHash;
-      const treeRoots = [];
+      let name, nameHash, listener;
+      let treeRoots = [];
+
+      const checkTree = async (tree, compacted = false) => {
+        for (const [index, hash] of treeRoots.entries()) {
+          if (compacted && index < (treeRoots.length - 8)) {
```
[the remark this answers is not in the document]
nodech: We expect to always have last 8 tree roots because in the beginning of the test we make sure to have `1:8` ratio with `treeInterval` and `keepBlocks`. In the beginning of the test we mine `20 blocks` (maybe that should be dynamic based on those)

smcki012:
Fantastic work on this. Benchmarks are very promising. HSD is getting really lean for node operators!

lib/node/rpc.js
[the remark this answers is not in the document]
pinheadmz: ok

test/chain-migration-test.js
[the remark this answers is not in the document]
pinheadmz: ah ok, comments confused me. so the "revert" stuff is just to reset the test environment so we can migrate again. and what were testing is that old layout.s is now `treeState.treeRoot`

lib/blockchain/chaindb.js
```diff
@@ -1732,6 +1818,9 @@ class ChainDB {
     if (this.options.prune)
       throw new Error('Cannot reset when pruned.');
 
+    if (this.treeState.compactionHeight !== 0)
+      throw new Error('Cannot reset when tree is compacted.');
```
pinheadmz, line 1822: I suppose we *could* actually allow this as long as `entry.height >= compactionHeight` ?

nodech: This is mostly meant for SPV, so decided not to disable it on tree compaction. But yeah, if node is not pruned (which is also prohibited) we could technically allow reset up to `entry.height >= compactionHeight`. Problem could occur if reset block == compactionHeight and at that point reorg is detected in that place - that could lead to issue where user can't reorg properly in the past. Ofc this is highly unlikely, but given nature of compaction I believe we should leave how it is.

lib/blockchain/chain.js
```diff
@@ -2044,6 +2102,98 @@ class Chain extends AsyncEmitter {
...
+    // To support chain reorgs of limited depth we compact the tree
+    // to some commitment point in recent history, then rebuild it from there
+    // back up to the current chain tip. In order to support pruning nodes,
+    // all blocks above this depth must be available on disk.
+    // This actually further reduces the ability for a pruning node to recover
+    // from a deep reorg. On mainnet, `keepBlocks` is 288. A normal pruning
+    // node can recover from a reorg up to that depth. Compacting the tree
+    // potentially reduces that depth to 288 - 36 = 252. A reorg deeper than
+    // that will result in a `MissingNodeError` thrown by Urkel inside
+    // chain.saveNames() as it tries to restore a deleted state.
```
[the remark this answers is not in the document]
nodech: I believe It still reduces to 252 (worst case).

test/chain-tree-compaction-test.js
```diff
@@ -761,6 +761,88 @@ describe('Tree Compacting', function() {
...
+      node = new FullNode(nodeOptions);
+      compactWrapper(node);
+
+      await node.open();
+      assert.strictEqual(compacted, false);
+
+      waiter = forEvent(node, 'connect');
+      // commit 1 + treeInterval to be sure we need to recompact.
+      await node.rpc.generateToAddress(
+        [network.names.treeInterval + 1, new Address().toString('regtest')]
```
pinheadmz, line 831: Why not run the tests on the actual boundary?

```diff
diff --git a/test/chain-tree-compaction-test.js b/test/chain-tree-compaction-test.js
index cc5d59859..f6d79abc8 100644
--- a/test/chain-tree-compaction-test.js
+++ b/test/chain-tree-compaction-test.js
@@ -808,8 +808,8 @@ describe('Tree Compacting', function() {
       await node.open();
       assert.strictEqual(compacted, true);
 
-      // setup interval - 1 blocks for next test.
-      blocks = interval - 1;
+      // setup keepBlocks + interval - 1 blocks for next test.
+      blocks = interval + network.names.treeInterval - 1;
       waiter = forEvent(node, 'connect', blocks);
       await node.rpc.generateToAddress(
         [blocks, new Address().toString('regtest')]
@@ -826,9 +826,9 @@ describe('Tree Compacting', function() {
       assert.strictEqual(compacted, false);
 
       waiter = forEvent(node, 'connect');
-      // commit 1 + treeInterval to be sure we need to recompact.
+      // commit 1 more block to be sure we need to recompact.
       await node.rpc.generateToAddress(
-        [network.names.treeInterval + 1, new Address().toString('regtest')]
+        [1, new Address().toString('regtest')]
       );
       await waiter;
       await node.close();

```

lib/blockchain/chaindb.js
pinheadmz: cool

lib/blockchain/chain.js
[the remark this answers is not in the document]
pinheadmz: yep, as discussed on tg.

nodech (approved):

pinheadmz:
ACK 6e3a113b8cf038a0554aa3cca5e485d196f4e0be

```
Hash: SHA256

ACK 6e3a113b8cf038a0554aa3cca5e485d196f4e0be
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKaKMQACgkQ5+KYS2KJ
yTryyg//ax9DOb5unr1CNhwCNLaUDA0hS01tvzMRnrEYEmcWtyOC6eSFlnLnuO/V
/9wUYaoGyeJBBZx6QKW//zjUv8D0Xyc/46wtDdqmG9Oo0N29CUMTSjJKkKsjRLWG
mQNWCG0ai/wilp/x+p32usQYQLeD6p58DkkUhbXugKeJgcSKC98VOXj3o1Z4/r6K
z3glTLRoltihWFsVTVp648EC3+QxjiAxn4pOXJxyIPwOm6MyZSubmF3pVVw4Cybm
G5U2+rtTgXyPY0BFC5t74SYSRLccmur40F/f+fH5xSw8TYYd24dkwLuwMup01z3r
OkHV78XMaQ8IfuIlTjImWQmEHPUCsnJ10nWaATa6HN1kOqp8TzV05S8CNVKOJOW9
7lf9D0MTw3NHSiFDKoilmWm+V4gFWuEx6qHiVL8VOI6QOsDnYYbPBVwPMxyT5GZJ
blTw8CyGWRROyEJ1L69k3UUMZ32u7lLt+G3guc0Vj87xqmc0XLhYG/LI+GxOsath
0fh4EQZxdMaWAVukZgdUaHAea0zRPWVG2Kj1OwDxXgFSVIzcLhMKJEYYsd1zLm88
IJ2DdIoaF3yEFdZedrfSVE1wDsso+QUFITu1KQD6iCiW/Gj533p5vFUFI55xO66z
+BMgLu1wjf6h8Yj1oUX86hNI8lKlfVrIcdlgQ92Y7ZCYlCpp8/E=
=qIZu
-----END PGP SIGNATURE-----
```

</details>

pinheadmz:
ACK 74cc006ef089208a8a715746c9bed470a24a0b70

```
Hash: SHA256

ACK 74cc006ef089208a8a715746c9bed470a24a0b70
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKaLQIACgkQ5+KYS2KJ
yToirxAAypbhBjbCYVgyqCg4y80e46mC/hTueIUOxZHqMYctKRxFhFiXmBn3dk9X
0X0RyzwWQ1IOv1L+CrrPWoqITVBlRaq/pEhvKyzkIvHlbul69ekZsWpVH6iPjl4c
joSK0MeNXjKJJKO+I7JH6fShR/QLY8RNGfcO1jo2Ce70rg7+sw3mbf8ikp4sBfgr
YeATGgLuSysGeluSfrYS+bfn+KwdwCxp58b2/toDZkkvLjtHqd4t/zu9HG+0rSg3
pYM3HFbEb8+O7H8HhHjJEFraCduRFar6SgDRH2UP/dcWMxS739zFOBbK673Cw/6e
vyIUm0lNXYEIERxAhJqwtDn9885rnQwMJTlxem7odCDQYizGtQb/0Ym6Sq2RKpfD
+qAWRAhxh1TcJln/K6J2T87e7u1PZp+dHM0GP3+4Bbld5xQU/DU/tllYWtDM7btK
nNegA4f0NfDVmamL3QdG35cWm9z9vfA3K7c0AeS5aenpV84ezCpBpa7aBlQ1M1V/
y6WO0qa0DyHdF4oau75WeYFB9wpTF7RLpgwGPQVdhWePsicMX8W9AoIGOzvLbnk0
nhBxGR+7a/hayD0ovGZKmM4VXeBFqDGnc+hmgFj/quEAeX+5MoHDSjgGFxeKOsz0
xRzcg0ROxXpfwubkk3jrQW4bm4ZugQ0zUPBuSwvIH98+lp5/Qmw=
=dtxM
-----END PGP SIGNATURE-----
```

</details>

satoshi-MAM:
Exposing bad individuals

PR #671, opened 2021-12-09
mempool: reflect spent coins in mempool coinview

nodech:
This is port of https://github.com/bcoin-org/bcoin/pull/641 by tynes in #153.

Reviewed for `hsd` usage of `getSpentView`, its same as `bcoin`.

Closes #153.

From commit message 8cecb6b6af9d1c39c87016b092251e57c3f2e4e1

```
Currently coinview does not account for spent coins in the mempool,
This does not create problems because we have additional checks in
right places which detect double spends, but technically
coinview should help you detect double spent in the mempool as well.
This way it will be compatible with chain.getCoinView.

getSpentView will still return all outputs that are available
in the mempool. (This could also return spentView from indexers if
available, this method is used by `signrawtransaction`.)

Port of bcoin commit: 32cba1bf4aa293bcf49c1849587dc1482ce98050
https://github.com/bcoin-org/bcoin/commit/32cba1bf4aa293bcf49c1849587dc1482ce98050

Co-authored-by: Nodar Chkuaselidze <nodar.chkuaselidze@gmail.com>
Co-authored-by: Mark Tyneway <mark@purse.io>```

### DIFF
--- a/test/mempool-test.js
+++ b/test/mempool-test.js
@@ -233,6 +233,63 @@ describe('Mempool', function() {
     }));
   });
 
+  it('should get spent coins and reflect in coinview', async () => {
+    const wallet = new MemWallet();
+    const addr = wallet.getAddress();
+
+    const dummyCoin = dummyInput(addr, random.randomBytes(32));
+
+    const mtx1 = new MTX();
+    mtx1.addOutput(wallet.getAddress(), 50000);
+    mtx1.addCoin(dummyCoin);
+
+    wallet.sign(mtx1);
+
+    const tx1 = mtx1.toTX();
+    const coin1 = Coin.fromTX(tx1, 0, -1);
+
+    const mtx2 = new MTX();
+    mtx2.addOutput(wallet.getAddress(), 10000);
+    mtx2.addOutput(wallet.getAddress(), 30000); // 10k fee
+    mtx2.addCoin(coin1);
+
+    wallet.sign(mtx2);
+
+    const tx2 = mtx2.toTX();
+
+    await mempool.addTX(tx1);
+
+    {
+      const view = await mempool.getCoinView(tx2);
+      const sview = await mempool.getCoinView(tx2);
+      assert(view.hasEntry(coin1));
+      assert(sview.hasEntry(coin1));
+      assert.strictEqual(mempool.hasCoin(coin1.hash, coin1.index), true);
+      assert.strictEqual(mempool.isSpent(coin1.hash, coin1.index), false);
+    }
+
+    await mempool.addTX(tx2);
+
+    {
+      const view = await mempool.getCoinView(tx1);
+      const sview = await mempool.getSpentView(tx1);
+
+      assert(!view.hasEntry(dummyCoin));
+      assert(sview.hasEntry(dummyCoin));
+      assert.strictEqual(mempool.hasCoin(coin1.hash, coin1.index), false);
+      assert.strictEqual(mempool.isSpent(coin1.hash, coin1.index), true);
+    }
+
+    {
+      const view = await mempool.getCoinView(tx2);
+      const sview = await mempool.getSpentView(tx2);
+      assert(!view.hasEntry(coin1));
+      assert(sview.hasEntry(coin1));
+      assert.strictEqual(mempool.hasCoin(coin1.hash, coin1.index), false);
+      assert.strictEqual(mempool.isSpent(coin1.hash, coin1.index), true);
+    }
+  });
+
   it('should handle locktime', async () => {
     const key = KeyRing.generate();
     const addr = key.getAddress();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/mempool-test.js [hunk 1]
```diff
@@ -233,6 +233,63 @@ describe('Mempool', function() {
...
+
+    wallet.sign(mtx2);
+
+    const tx2 = mtx2.toTX();
+
+    await mempool.addTX(tx1);
+
+    {
+      const view = await mempool.getCoinView(tx2);
+      const sview = await mempool.getCoinView(tx2);
```
pinheadmz, line 264:
```suggestion
      const sview = await mempool.getSpentView(tx2);
```

Right ??

test/mempool-test.js
```diff
@@ -233,6 +233,63 @@ describe('Mempool', function() {
...
+    }
+
+    await mempool.addTX(tx2);
+
+    {
+      const view = await mempool.getCoinView(tx1);
+      const sview = await mempool.getSpentView(tx1);
+
+      assert(!view.hasEntry(dummyCoin));
+      assert(sview.hasEntry(dummyCoin));
```
pinheadmz, line 278: Not sure if you meant `coin1` here? I think these two lines can go in the previous block after `tx1` is added, and ALSO right here on L278 you can add similar checks for `coin1` now that tx2 is added.

test/mempool-test.js
```diff
@@ -233,6 +233,63 @@ describe('Mempool', function() {
...
+      assert.strictEqual(mempool.hasCoin(coin1.hash, coin1.index), true);
+      assert.strictEqual(mempool.isSpent(coin1.hash, coin1.index), false);
+    }
+
+    await mempool.addTX(tx2);
+
+    {
+      const view = await mempool.getCoinView(tx1);
+      const sview = await mempool.getSpentView(tx1);
+
```
pinheadmz, line 276: nit: other blocks dont have newline here, but i dont care

pinheadmz: why aren't you checking the tx1 view right after adding tx1 to mempool? You add tx2 first then check for `dummyCoin`?

test/mempool-test.js
```diff
@@ -440,6 +497,8 @@ describe('Mempool', function() {
       memory: true
     });
 
+    const wallet = new MemWallet();
```
pinheadmz, line 500: Required? There is a global `wallet` at the top of the test. Also, we don't need to set network?

nodech: It's "cleaner" with new instance.

test/mempool-test.js
nodech: sview/view checks for dummyCoin can be done before tx2 is added, but that block mainly checks hasCoin and isSpend for the coin. The checks you are talking is the next block.

nodech:
We could refactor that to get rid of `isDoubleSpend` check there, true. Then it will have to move into `maybeOrphan` (or even rename the method). `maybeOrphan` checks against `coinView` and does not take into account `spentView`, so it will detect mempool spent coins as orphans as of now. `isDoubleSpend` and `maybeOrphan` separation simplifies the logic though.

pinheadmz (approved):
ACK 0a694309195c1236134f0a297c96f187551daa0f

```
Hash: SHA256

ACK 0a694309195c1236134f0a297c96f187551daa0f
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmHUmjkACgkQ5+KYS2KJ
yTqrwRAAm/uxjc4wduHSAux0Rj4jAMOY89Jq/90h3QArYU+bLy4R23EnYPH6vJML
F1XiHwKnkbV0nQTljYrcaTZCGZtNfBUbKvV6yzMlEee7IiTO6SsTMA21GGEObeZl
Ew6dwEq6w14XfWFpgWiBO6wiNzffx1HI7WxaMtBVf2KYCEHQY4/W4kJjkBOpJxe7
9SbcDjvScQBWdMVBouVrXyz0cBf0OMROJP0XB5kAz/CPcEeduHo6hos3apCX0s2l
sc6izi8QWORLPfh1ZEawZWXeJNctZzSQGZcJNXsgqPqFjks12nh00PCSGJBtOhYi
8nmwmM6e1lPHBo/csqW5V2bwjdgnqvi6qm7sq70NR79q9ZnZgkGrA4FWoJyGFPEE
+TgwZ5g2nwa5tXBXYVg6QPrOotNn7yrTAb7/E30PE14tsUt9ls2yxnOdX/zPvP/v
FwM+tA5BGGAEjix78keU9LVGEPTHapjWS8CNCNK6bW2Ya0krRO/ZykAekSHzSYYr
+ErNH89CiAXlGoocE2wRtahoz0502tCQQg5FFWCP4IHatJSYByYA//O6nkybGRVh
OSxhqtmWt1QjkcIArJY4o6vnhBm91XjlGloB2IfAKCRGIcMnfGXJzwAck2135ilW
GLatPvQNh/z31bmNfJsopCyBZE0Nk1qU3JR2fUEgXfwKebseu7c=
=OQ8G
-----END PGP SIGNATURE-----
```

</details>

PR #679, opened 2022-01-01
fix(rpc): Can't sign/verify with name before auction is closed.

### LINKED ISSUE #626: signmessage during REVEAL
Currently it's possible to signmessage during REVEAL while my bid is in the lead but higher lockups are still unrevealed.

like this (at least up to block 79215):
Name:      cmax
String:    Highest Lockup still not revealed
Signature: bmhWbed/6745rkwzJufRYhFvEi0iSLvbZAHT8zHr1dV+j7XLVa57soU2Po8Vtj/hHBGQA+CDf3WzbEneAAQfHA==

Falci:
Fix #626

### DIFF
--- a/lib/node/rpc.js
+++ b/lib/node/rpc.js
@@ -2156,6 +2156,8 @@ class RPC extends RPCBase {
     const name = valid.str(0, '');
     const sig = valid.buf(1, null, 'base64');
     const str = valid.str(2);
+    const network = this.network;
+    const height = this.chain.height;
 
     if (!name || !rules.verifyName(name))
       throw new RPCError(errs.TYPE_ERROR, 'Invalid name.');
@@ -2165,6 +2167,9 @@ class RPC extends RPCBase {
 
     if (!ns || !ns.owner)
       throw new RPCError(errs.MISC_ERROR, 'Cannot find the name owner.');
+
+    if (!ns.isClosed(height, network))
+      throw new Error('Invalid name state.');
 
     const coin = await this.chain.getCoin(ns.owner.hash, ns.owner.index);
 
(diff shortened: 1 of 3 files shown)


lib/node/rpc.js
```diff
@@ -2166,6 +2169,11 @@ class RPC extends RPCBase {
     if (!ns || !ns.owner)
       throw new RPCError(errs.MISC_ERROR, 'Cannot find the name owner.');
 
+    const state = ns.state(height, network);
+
+    if (state < states.CLOSED)
```
pinheadmz, line 2174: Technically `REVOKED` is > `CLOSED`, any reason why this isn't `===` ?

Falci: You're. It should throw error if `!== CLOSED`. Added an extra test and fixed.

realrasengan:
This is great!

I had one question, since namestate exposes an `isClosed(height, network)`:
```
    return this.state(height, network) === states.CLOSED;
```
would it make sense to use that?  Sorry to be nitpicky -- but whatever we decide will probably also help to define the coding style go forward!

test/wallet-rpc-test.js
```diff
@@ -466,6 +466,138 @@ describe('Wallet RPC Methods', function() {
     });
   });
 
+  describe('signmessagewithname & verifymessagewithname (reveal in progress)', () => {
```
pinheadmz, line 469: So I think these tests can the REVOKE tests can be combined into the existing `describe` block for simplicity. Change line 401 to mine 1 block only, then the first few `it` blocks can be the not-closed-yet tests. Then after line 466 you can mine more blocks and squeeze in the REVOKE test, so that `describe` block is all one long timeline.

pinheadmz (changes requested):
In addition to merging the tests, I think all this PR can be squashed into one commit. We don't need extra "Lint" commits in the history ;-)

PR #681, opened 2022-01-01
Update coveralls action

Falci:
I'm trying to fix the [Source not available](https://coveralls.io/builds/45317392/source?filename=lib%2Fprotocol%2Fconsensus.js#L4720) from Coveralls.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -35,6 +35,6 @@ jobs:
       run: npm run test-ci
 
     - name: Coverage
-      uses: coverallsapp/github-action@v1.1.1
+      uses: coverallsapp/github-action@master
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}


pinheadmz:
I think you just have to be logged in to GitHub ? And/or logged in to coveralls. It's possible you might also need to be a repo owner ?

pinheadmz:
0AAB1586-C3A7-4B87-9458-E9DD8B67B45A

this is what I see when I click that link

Falci:
Hmm.. weird.. 

image

Falci:
Yep, just need to login there as well 🤦‍♂️

pinheadmz:
OK I think this is fine for now but I have issues with coveralls anyway, for example this pull request coveralls reports

'Overall coverage decreased (-4.1%) to 62.689%'

We may have some non-deterministic tests, but this kind of result is what makes coveralls useless ;-/

PR #688, opened 2022-02-08
blockstore - separate write and prune batches.

nodech:
ChainDB and BlockStore are separate and committing to them is not really atomic,
so we need to make sure the order operations can handle failure at any step.
Previously blockstore writes and prunes would happen after the chaindb was
written, but that could lead to the situation where the blockstore did not
actually have a block but chaindb had the information.
  This PR separates write and prune batches for the blockstore, so we can try
writing to blockstore first. In case writing to blockstore fails whole operation
will get aborted. Nothing gets written into the chaindb, so the information is
consistent. In case blockstore writes a block and then chaindb write fails,
we just get extra data in the blockstore. But most of the time, the same block
write will happen again (either main or alt chains).
  Pruning always happens after the chaindb was updated to avoid situation, where
chaindb thinks the data is stored, but blockstore has removed it. If blockstore
fails to prune after chaindb removed the information, worst case scenario
(prune mode) we get a maxFileLength(128 MB by default) space wasted.

Issue found here: https://github.com/kyokan/bob-wallet/issues/453

### DIFF
--- a/test/chain-blockstore-test.js
+++ b/test/chain-blockstore-test.js
@@ -1,0 +1,279 @@
+'use strict';
+
+const assert = require('bsert');
+const {BloomFilter} = require('bfilter');
+const Network = require('../lib/protocol/network');
+const {FileBlockStore, LevelBlockStore} = require('../lib/blockstore');
+const Chain = require('../lib/blockchain/chain');
+const WorkerPool = require('../lib/workers/workerpool');
+const Miner = require('../lib/mining/miner');
+const {rimraf, testdir} = require('./util/common');
+const MemWallet = require('./util/memwallet');
+
+const network = Network.get('regtest');
+
+const blockStoreTypes = {
+  'File': async (location) => {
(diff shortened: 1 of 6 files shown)


test/chain-blockstore-test.js
```diff
@@ -0,0 +1,279 @@
...
+        await chain.reset(1);
+      } catch (e) {
+        err = e;
+      }
+
+      assert(err);
+      assert(err.message, 'Failed to prune.');
+
+      // recover start method.
+      db.start = dbStart;
```
pinheadmz, line 268: should this move up before the assertions? if this test fails and this restore isnt executed, wont it have downstream effects on all following tests?

pinheadmz:
Thanks for the fast work on this, LGTM now after a few nits. Makes sense to split up write/prune and execute them before/after the levelDB index as you've done. I wonder if this is the kind of thing we should add to a patch version and release, maybe Bob Wallet should do the same.

PR #691, opened 2022-02-23
test: Added more tests to NOINPUT implementation

Anunayj:
Due to how NOINPUT is implemented, nSequence for the given input
is rendered ineffective, this deviates from the original implementation
in the eltoo paper, and developers need to be aware about this little quirk.

### DIFF
--- a/test/sighash-test.js
+++ b/test/sighash-test.js
@@ -21,6 +21,7 @@ const network = Network.get('regtest');
 
 // sighash types
 const {
+  ALL,
   SINGLEREVERSE,
   NOINPUT,
   ANYONECANPAY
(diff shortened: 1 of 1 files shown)


test/sighash-test.js
```diff
@@ -249,6 +250,85 @@ describe('Signature Hashes', function () {
...
+      )
+       
+      assert(mtx.verify());
+      // A malacious user can simply change the sequence of the input and spend
+      // it rendering relative timelocks uneffective, use nLockTime to enforce
+      // timelocks.
+      mtx.input(0).sequence = 0xffffffff;
+      assert(mtx.verify());
+      
+    });
```
Anunayj, line 329: I think this is a relatively important issue, cause a developer unaware about this might easily end up implementing a scheme where presigned relative timelocks are in use with conjunction with NOINPUT (Example: force closing a payment channel). But might miss the detail that due to this bug the other party can simply change the nSequence to anything they want rendering it ineffective.

PR #694, opened 2022-02-28
Fix CPFP when tx fee is less than MIN_RELAY

Anunayj:
Mempool inclusion policy does not consider descendant fee when
checking transaction for minimum fee, this would allow you to
incentivize the miner to include your transaction without having
to wait for a block with low weight or your transaction to get
evicted.

Network policy would still disallow relaying transactions < minFeeRate, but this would allow CPFP to be used as a fee bumping mechanism in case a low fee transaction does get into mempool.

### DIFF
--- a/test/miner-test.js
+++ b/test/miner-test.js
@@ -9,6 +9,7 @@ const Mempool = require('../lib/mempool/mempool');
 const Miner = require('../lib/mining/miner');
 const Address = require('../lib/primitives/address');
 const MTX = require('../lib/primitives/mtx');
+const Coin = require('../lib/primitives/coin');
 const MemWallet = require('./util/memwallet');
 const {BufferSet} = require('buffer-map');
 
@@ -77,6 +78,7 @@ describe('Miner', function() {
 
   let walletAddr;
   const txids = new BufferSet();
+  let coin, parentTX;
 
   it('should generate 20 blocks to wallet address', async () => {
     walletAddr = wallet.createReceive().getAddress();
@@ -110,7 +112,6 @@ describe('Miner', function() {
     }
 
     assert.strictEqual(mempool.map.size, 10);
-
     const block = await miner.mineBlock(chain.tip, addr);
     await chain.add(block);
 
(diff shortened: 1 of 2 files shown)


test/miner-test.js
```diff
@@ -180,4 +181,81 @@ describe('Miner', function() {
...
+    assert.rejects(
+      async () => await mempool.addTX(tx, -1),
+      {
+        code: 'duplicate',
+        reason: 'bad-txns-inputs-spent'
+      }
+    );
+    // orignal tx is still in mempool
+    assert.strictEqual(mempool.map.size, 1);
+  });
```
pinheadmz, line 234: This looks more like a mempool test than a miner test. Do we need to add it? Is this error already covered in the mempool tests?

test/miner-test.js
```diff
@@ -180,4 +181,81 @@ describe('Miner', function() {
...
+
+  it('should include child transaction if child pays enough fee (CPFP)', async () => {
+    const fee = 1000;
+
+    // Fee should be enough for both the first transation and second transaction
+    assert(fee > 140 + 108);
```
pinheadmz, line 240: Where did you get these values from (140 and 108)? It might be more clear to see the values pulled from the MempoolEntry itself is possible. In addition, asserting that the `descSize` and `descFee` of the parent TX is being updated when child is added to mempool and that the value is correct.

test/miner-test.js
```diff
@@ -180,4 +181,81 @@ describe('Miner', function() {
...
+    parentTX = mtx.toTX();
+
+    await mempool.addTX(parentTX, -1);
+    assert.strictEqual(mempool.map.size, 1);
+
+    const block = await miner.mineBlock(chain.tip, addr);
+    await chain.add(block);
+
+    // TX is still in mempool, nothing in block except coinbase
+    assert.strictEqual(mempool.map.size, 1);
```
pinheadmz, line 209: I think this test block should begin either with a `mempool.reset()` or assertion that mempool is empty. What might be even better is starting a new `describe` block called `CPFP` or something. Especially since the "should not include free transaction in a block" test is basically a duplicate. We know we can't include free TXs in blocks from L127-L165.

pinheadmz:
I'm still thinking this through... would be nice to see another test or two to cover edge cases, for example:

A CPFP chain of 3, 4, 5 txs long? Maybe even a `MEMPOOL_MAX_ANCESTORS` or `MEMPOOL_MAX_ANCESTORS + 1` length CPFP chain? This would probably be a mempool test, not a miner test. I think this limit is covered in wallet test but not yet in mempool.

What happens if a tx in a CPFP chain is confirmed but the rest is not? What if its the first or even middle TX in a CPFP chain? We should make sure that a TX's descendant is removed from the mempool for any reason that we do indeed re-adjust that TX's `descSize` and `descFee`. We might even need to evict a TX if it *becomes* free.

Same applies to "orphan" TXs. We may get parent / child TXs in reverse order -- what happens then?

test/miner-test.js
Anunayj: Oh yea, I'll fix that, sorry mb there

test/miner-test.js
Anunayj: yea I think a separate describe block is better, since this itself got way too long for just miner test. or maybe a complete different test file for CPFP and in future RBF (and other fee bumping mechanisms) might be better.

Anunayj:
I think at this point, we should just make a separate test for and call it `fee-bumping.js` or something. This itself is getting too long for being in `miner-test.js`

Issue #689, opened 2022-02-11
HTTP api `info` returns incorrect ports


rithvikvibhu:
Start hsd with (different p2p and brontide ports):
```sh
hsd --prefix /tmp/hsd_1 --network regtest --no-dns true --no-wallet --index-tx true --index-address true --port 14138 --brontide-port 46816 --http-port 14137 --listen true
```

And the port `14138` is picked up correcty. Log says:
```
[info] (net) Pool server listening on 0.0.0.0 (port=14138).
```

But with `hsd-cli info`:
```json
{
   "pool": {
    "host": "106.x.x.x",
    "port": 14038,
    "brontideHost": "0.0.0.0",
    "brontidePort": 46806,
    "identitykey": "aorsxa...",
    "agent": "/hsd:3.0.1/",
    "services": "11",
    "outbound": 0,
    "inbound": 1
  },
}
```
The `host` IP address is correctly set to public IP, but `port` and `brontidePort` are wrong and show the default ports (14038, 46806).

Can connect to port 14138, so log is correct and `info` is wrong.

PR #696, opened 2022-03-05
Clarify ports / public ports in node HTTP getinfo

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/689

The issue is that the ports displayed in `info` are the *public* ports set by `--public-port` and `--public-brontide-port` not `--port` and `--brontide-port`. Public hosts and ports are what we advertise to peers when we are listening for inbound connections, this is the hostname that gets gossiped around the network. The other options set the *actual* host and port the server actually listens to 🤷‍♂️ 

## Current and unchanged behavior of hsd

- We will not advertise our own address unless `--listen=true`
- On mainnet (only) we will not advertise our own address unless an address is specifically passed with `--public-host`
  - This behavior was added to hsd to prevent unreachable nodes from accidentally (user error) advertising

This PR adds some new fields to the `pool` object in the "get info" JSON. The existing data will now indicate the actual listening host/port and the `public` object will indicate the advertised data as set by the user. (See examples below).

This PR also refactors the `node-http-test` module to make more room for this coverage, and introduces a new `p2p-test` module which connects mainnet nodes together with different settings and monitors the `ADDR` packets transmitted.

### ex. 1

` hsd --memory=true --network=main`

```
  "pool": {
    "host": "0.0.0.0",
    "port": 12038,
    "brontidePort": 44806,
    "identitykey": "ang7l54xdrg47duk3jax3f5blmgblfkkwq6t6fs7luyoawehh5jk4",
    "agent": "/hsd:3.0.1/",
    "services": "1",
    "outbound": 8,
    "inbound": 0,
    "public": {
      "listen": false,
      "host": null,
      "port": null,
      "brontidePort": null
    }
  },
```
### ex. 2

` hsd --memory=true --network=main --listen=true`

```
  "pool": {
    "host": "0.0.0.0",
    "port": 12038,
    "brontidePort": 44806,
    "identitykey": "al47esnqoajntzsekilqjoq6fcyeq7uuzxr4d2j62ug34kydhbqbg",
    "agent": "/hsd:3.0.1/",
    "services": "1",
    "outbound": 8,
    "inbound": 0,
    "public": {
      "listen": true,
      "host": null,
      "port": null,
      "brontidePort": null
    }
  },
```

### ex. 3

` hsd --memory=true --network=main --listen=true --public-host=200.99.98.97 --public-port=12345 --public-brontide-port=54321`

```
  "pool": {
    "host": "0.0.0.0",
    "port": 12038,
    "brontidePort": 44806,
    "identitykey": "anv6dprpxnyqjxgmljw2ys7p5yi7s4syg6vc5x2eptym3x7jaluha",
    "agent": "/hsd:3.0.1/",
    "services": "1",
    "outbound": 8,
    "inbound": 0,
    "public": {
      "listen": true,
      "host": "200.99.98.97",
      "port": 12345,
      "brontidePort": 54321
    }
  },
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## unreleased
 
-### Node changes
+### Node API changes
+
+- HTTP API endpoint `/` (`hsd-cli getinfo`) now includes "public" networking settings.
 
 - RPCs `getnameinfo` `getnameresource` `verifymessagewithname` and `getnamebyhash`
 now accept an additional boolean parameter `safe` which will resolve the name from the Urkel
@@ -14,11 +16,12 @@ to these calls.
   - `decoderesource` like `decodescript` accepts hex string as input and returns
   JSON formatted DNS records resource.
 
-### Wallet changes
+### Wallet API changes
 
 - New RPC methods:
   - `createbatch` and `sendbatch` create batch transactions with any number
   of outputs with any combination of covenants.
+
 ## v4.0.0
 
 **When upgrading to this version of hsd you must pass
--- a/lib/net/hostlist.js
+++ b/lib/net/hostlist.js
@@ -1033,6 +1033,12 @@ class HostList {
 
     if (!src) {
       for (const dest of this.local.values()) {
+        if (this.network.type === 'main') {
+          // Disable everything else for now.
+          if (dest.type < HostList.scores.UPNP)
+            continue;
+        }
+
         if (dest.addr.hasKey())
           continue;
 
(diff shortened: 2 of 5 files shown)


rithvikvibhu (approved):
LGTM. pool host and ports and the public host and ports are shown correctly.

Isn't there a `public-brontide-port` like how `port` has `public-port`?
https://hsd-dev.org/guides/config.html doesn't have it so I guess not, just checking.

Edit: just re-read the desc above and there is. So the config page needs to be edited to include `--public-brontide-port`

rithvikvibhu:
Also, how do you feel about defaulting `public-port` to `port`? Same for `brontide-port`.

pinheadmz:
> Also, how do you feel about defaulting `public-port` to `port`? Same for `brontide-port`.

I think no, because they are used for separate things. `port` and `publicPort` are both set to the same network default. If a user wants to set both they should set both manually.

rithvikvibhu (approved):

PR #697, opened 2022-03-09
feat(wallet): sendbid accepts addr to bid to

Falci:
This change adds the option to (optionally) pick which address will be used to submit a bid.

```sh
hsw-rpc sendbid $name $amount $lockup [$address]
```

### DIFF
--- a/lib/wallet/rpc.js
+++ b/lib/wallet/rpc.js
@@ -2098,7 +2098,8 @@ class RPC extends RPCBase {
     const opts = this._validateBid(args, help, 'sendbid');
     const wallet = this.wallet;
     const tx = await wallet.sendBid(opts.name, opts.bid, opts.value, {
-      account: opts.account
+      account: opts.account,
+      address: opts.address
     });
 
     return tx.getJSON(this.network);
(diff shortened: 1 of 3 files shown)


pinheadmz:
Can you explain your use case a bit more? This action is potentially risky since a BID without REVEAL can destroy money, and the wallet may not have the private key for the arbitrary address you are passing to the API call?

Falci:
> the wallet may not have the private key for the arbitrary address you are passing to the API call

If the wallet don't know the address used for BID, it will fail to load its account and won't be able to generate the nonce.
https://github.com/handshake-org/hsd/blob/611421d3513e6514e8e6754c7bf21bd9932dcb20/lib/wallet/wallet.js#L1203-L1221

pinheadmz:
but whats wrong with just using the wallet's next receive address?

https://github.com/handshake-org/hsd/blob/107ed2be9280bb4292d21330e7d1ace62da9a002/lib/wallet/wallet.js#L1694

I'm just wondering if this is so useful we need to change the API and continue to maintain it. (e.g. do we need to add this option to every wallet call? or maybe a bigger custom create-tx endpoint is better suited to cover all similar use-cases?)

I'm not sure what exactly you're up to, but I can also see maybe using the API to create an unsigned BID, then changing the receive address and then signing it -- using hsd as a library or using api calls from the application layer.

Anunayj:
> If the wallet don't know the address used for BID, it will fail to load its account and won't be able to generate the nonce.

Also I feel like depending on a undefined behavior like `generateNonce()` to fail when a bid is made to a non wallet address doesn't feel like that great of a idea, and very confusing for someone who might expect a arbitrary address bid to work.

PR #707, opened 2022-03-26
wallet: emit 'block connect' events

### LINKED ISSUE #621: WalletDB should re-emit block events
This is a pretty simple feature that would make GUI applications easier, for example right now Bob has to unhook rescan events from the wallet client and MITM them in order to notify the user: https://github.com/kyokan/bob-wallet/blob/master/app/background/wallet/service.js#L888-L890

WalletDB (well, wallet and txdb actually) already emit events for `tx`, `address`, and a few others but not for blocks.

Possibly the easiest approach would just be to re-emit events from the wallet's node client right after its done processing them: https://github.com/handshake-org/hsd/blob/master/lib/wallet/walletdb.js#L132-L154

rithvikvibhu:
Closes #621.

Changes:
- Added an emit for `block connect` with `entry, txs`. The first arguemnt is not wallet becaue the block doesn't belong to a specific wallet (unlike other WalletDB events)
- Added emits for `block disconnect` with `entry`.
- Test for WalletDB emitting `tx`, `address`, `block connect`, and `block disconnect`

Used in Bob Wallet https://github.com/kyokan/bob-wallet/pull/469.

### DIFF
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -2255,7 +2255,7 @@ class WalletDB extends EventEmitter {
       return 0;
     }
 
-    let total = 0;
+    const walletTxs = [];
 
     try {
       // We set the state as confirming so that
@@ -2264,8 +2264,9 @@ class WalletDB extends EventEmitter {
       // added and the height is updated.
       this.confirming = true;
       for (const tx of txs) {
-        if (await this._addTX(tx, tip))
-          total += 1;
+        if (await this._addTX(tx, tip)) {
+          walletTxs.push(tx);
+        }
       }
 
       // Sync the state to the new tip.
@@ -2274,12 +2275,14 @@ class WalletDB extends EventEmitter {
       this.confirming = false;
     }
 
-    if (total > 0) {
+    if (walletTxs.length > 0) {
       this.logger.info('Connected WalletDB block %x (tx=%d).',
-        tip.hash, total);
+        tip.hash, walletTxs.length);
     }
 
-    return total;
+    this.emit('block connect', entry, walletTxs);
+
+    return walletTxs.length;
   }
 
   /**
@@ -2329,6 +2332,7 @@ class WalletDB extends EventEmitter {
 
     if (!map) {
       await this.setTip(prev);
+      this.emit('block disconnect', entry);
       return 0;
     }
 
@@ -2345,6 +2349,8 @@ class WalletDB extends EventEmitter {
 
     this.logger.warning('Disconnected wallet block %x (tx=%d).',
       tip.hash, total);
+
+    this.emit('block disconnect', entry);
 
     return total;
   }
--- a/test/wallet-events-test.js
+++ b/test/wallet-events-test.js
@@ -1,0 +1,190 @@
+'use strict';
+
+const assert = require('bsert');
+const Address = require('../lib/primitives/address');
+const FullNode = require('../lib/node/fullnode');
+const Wallet = require('../lib/wallet/wallet');
+const WalletKey = require('../lib/wallet/walletkey');
+const ChainEntry = require('../lib/blockchain/chainentry');
+const MemWallet = require('./util/memwallet');
+
+describe('WalletDB Events', function () {
+  const node = new FullNode({
+    memory: true,
+    network: 'regtest',
+    plugins: [require('../lib/wallet/plugin')]
+  });
+
+  const { wdb } = node.require('walletdb');
+  let wallet;
+
+  async function mineBlocks(n, addr) {
+    addr = addr ? addr : new Address().toString('regtest');
+    for (let i = 0; i < n; i++) {
+      const block = await node.miner.mineBlock(null, addr);
+      await node.chain.add(block);
+    }
+  }
+
+  before(async () => {
+    await node.open();
+    wallet = await wdb.get('primary');
+  });
+
+  after(async () => {
+    await node.close();
+  });
+
+  it('should emit `tx` events', async () => {
+    const waiter = new Promise((resolve) => {
+      wdb.once('tx', (w, tx) => resolve([w, tx]));
+    });
+
+    const walletReceive = await wallet.receiveAddress();
+    await mineBlocks(1, walletReceive);
+
+    const [w, tx] = await waiter;
+
+    assert(w);
+    assert(w instanceof Wallet);
+    assert(w.wid === 0);
+
+    assert(tx);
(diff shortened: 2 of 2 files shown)


lib/wallet/walletdb.js
```diff
@@ -2279,6 +2279,8 @@ class WalletDB extends EventEmitter {
         tip.hash, total);
     }
 
+    this.emit('block connect', entry, txs);
```
pinheadmz, line 2282: Question for you and other reviewers: Should we filter the `txs` array emitted here so only TXs added to the wallet are emitted? We could use the logic on line 2267 above and return a new array of only TXs that belong to us... unless there's a reason why we would need the entire block?

test/wallet-events-test.js
```diff
@@ -0,0 +1,72 @@
...
+  it('should emit `tx` events', async () => {
+    const waiter = new Promise((resolve) => {
+      wdb.once('tx', (w, tx) => resolve([w, tx]));
+    });
+
+    const walletReceive = await wallet.receiveAddress();
+    await mineBlocks(1, walletReceive);
+
+    const [w, tx] = await waiter;
+    assert(w);
```
pinheadmz, line 43: These tests are basically good but don't we care what the emitted values actually are? Or only that they are non-null? I think you can improve these tests by asserting that the actual values are expected (wallet id? txid of tx? at least assert that it is an array?)

test/wallet-events-test.js
```diff
@@ -0,0 +1,72 @@
...
+    const waiter = new Promise((resolve) => {
+      wdb.once('block connect', (entry, txs) => resolve([entry, txs]));
+    });
+
+    const walletReceive = await wallet.receiveAddress();
+    await mineBlocks(1, walletReceive);
+
+    const [entry, txs] = await waiter;
+    assert(entry);
+    assert(txs);
```
pinheadmz, line 70: Same comment here about testing the actual emitted values, but also I'd like to see a block with more than just a coinbase TX. Maybe even include a TX that doesnt belong to the wallet and see if it gets emitted as well (pending discussion on my top comment about filtering the TXs by wallet)

pinheadmz (changes requested):

lib/wallet/walletdb.js
rithvikvibhu: Updated to filter only wallet tx. Note that it's not for a specific wallet; the tx will count if it belongs to at least 1 wallet.

test/wallet-events-test.js
rithvikvibhu: Added more asserts with `instanceof` and some output checks.

test/wallet-events-test.js
rithvikvibhu: Split the test with 2 blocks: One with a wallet tx and another without (and both are non-coinbase tx)

test/wallet-events-test.js
```diff
@@ -0,0 +1,148 @@
...
+
+const assert = require('bsert');
+const Address = require('../lib/primitives/address');
+const FullNode = require('../lib/node/fullnode');
+const Wallet = require('../lib/wallet/wallet');
+const WalletKey = require('../lib/wallet/walletkey');
+const ChainEntry = require('../lib/blockchain/chainentry');
+const MemWallet = require('./util/memwallet');
+
+describe('WalletDB', function () {
```
pinheadmz, line 11: minor nit:
```suggestion
describe('WalletDB Events', function () {
```

pinheadmz (approved):
Besides renaming the test block I think this is RFM

test/wallet-events-test.js
rithvikvibhu: Added.

rithvikvibhu:
Added `block disconnect` event. The test works by calling chain.disconnect on the latest block; once when block has a wallet tx, again when block doesn't have wallet tx.

So it looks like:
```
❯ npm run test-file test/wallet-events-test.js

> hsd@3.0.1 test-file
> bmocha --reporter spec "test/wallet-events-test.js"

  WalletDB Events
    ✓ should emit `tx` events
    ✓ should emit `address` events
    should emit `block connect` events
      ✓ with a block that includes a wallet tx
      ✓ with a block that does not include a wallet tx
    should emit `block disconnect` events
      ✓ with a block that includes a wallet tx
      ✓ with a block that does not include a wallet tx

  6 passing (115ms)
```

pinheadmz (approved):
ACK 450b0fc507a536ac425f2888a173b993e9f0babb

```
Hash: SHA256

ACK 450b0fc507a536ac425f2888a173b993e9f0babb
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKhAW0ACgkQ5+KYS2KJ
yTpZ+A/+M4EXTyzf19o8GthW3tE4HrsFjp8XFJStkbhDNssOBkE9bBqvCf3EdlYM
a3Q8y0M3mJZaIZsjVPEChpn3ZLK3m50m05a+TqICqBaNf7O4JpaKOFYlue/0QnOy
5rPnSpGKcymphM3Ghzu0L4Uc1CuJBrZ/m20wXEE44ilS1ydNXKyg4ZbjnVXf8GK2
H8zn5aBMzWCO0/ifaSPQw2q64XqKowqroRLWA072O/DN6xF4wptfaRhVZvB9bf01
h3l25Xak/fcbpfOPJPJW2WxRAAQAD62bcDS0sBjG0jnn0Dc4JoojLWhAFIc42bzK
DkIjyymzkTYGhjhAyLvpm9oPta76dKf/9CDSTJL4uL66I75nyWjCGgDjz1gOD//W
LhyzuJ4JvwwLPzEsrhusr9U1f4kFrF12/Bp/AdfKawv3r6Rwpz4nsS2DUJwgUFgs
YXUBPOL5FhsgvVTGzlzVahLGmN50nbuxYAZytW1mnYGuqCqHAjOFppK1oDdjpvDB
TN7eaOF5Y1e91gmuZBFdYVo+xY8n3FrFNeuHHxgtwAnQsiZoXMuaOGYXVj/NX2Ki
7Vp3xzUBt11HVsye/p83716dIbJY5xoXGj7h8yMVtvps1JBwDMtahYBGMohq4Awb
jLkPRMUtEErTUPAc4XT2sRss05Qmk492Jx5T75obgVIoO4HlbrQ=
=i6zC
-----END PGP SIGNATURE-----
```

</details>

PR #708, opened 2022-04-07
Limit "send all" batches to policy weight limit and fix estimated weight limit calculation

pinheadmz:
This PR fixes a few related wallet bugs including a new one I hadn't noticed before.

1. If a user tries to `revealAll` or `redeemAll` with too many BIDs (like around 600) the resulting transaction will be too big to send, violating the relay policy weight limit. This is a big issue because loss of funds can occur if reveals are not sent in time. The solution is to recalculate the estimated weight of the batch as we are building it, and break before the result is too heavy.

TODO:
- [ ] test for multisig
- [ ] should this be applied to regular `sendReveal` and `sendRedeem` as well? (They also produce batches)

2. Once the batch has hit the limit and stops adding outputs, there may be a number of BIDs still unrevealed or unredeemed in the wallet. We want the user to be able to send those in another batch right away. However, they would encounter the `"Could not resolve preferred inputs"` error because the second batch would try to double-spend outputs in the first batch. This error is common when users reveal a single name and then call `revealAll` before that first single reveal has confirmed. The solution is pretty simple: when we add reveals to a batch, we skip over outputs that are marked as spent already in txdb. This required the addition of a new wallet function `getUnspentCoin()`

3. While working with the TX size estimator to detect policy weight limit breaches, I noticed the MTX size estimator had unfinished bugs left over from its port from Bitcoin. In particular, we were not applying the witness discount to the scripts and signatures in multisig transactions. If you try to send a multisig transaction right now with master branch, you'll pay almost 100% extra fee because the estimator incorrectly computes a way bigger virtual size. The solution was to tweak the size estimator functions both in `wallet.js` and `mtx.js`. Also, since HNS signatures are fixed size (always 65 bytes as opposed to Bitcoin's 70-73 bytes) the estimator functions can be dialed in to almost always guaranteed accuracy. Tested.


pinheadmz:
Related upstream work: https://github.com/bcoin-org/bcoin/pull/858

lib/wallet/wallet.js
```diff
@@ -3267,7 +3298,15 @@ class Wallet extends EventEmitter {
   }
 
   /**
-   * Input size estimator for max possible tx size.
+   * Estimate witness size given output address.
+   * Unlike Bitcoin, our signatures are always 65 bytes.
+   * However, we still assume that the witness varInt size
+   * is only one byte. In short, this estimate may be off
+   * by 1 (at most) but only if a witness has > 253 items.
```
pinheadmz, line 3305: ...actually if the witness stack is too big, this varInt will be 3 bytes so off by 2 at most.

lib/wallet/wallet.js
```diff
@@ -3267,7 +3298,15 @@ class Wallet extends EventEmitter {
...
+   * Also note we are only processing the witness data here,
+   * which will be scaled down by WITNESS_SCALE_FACTOR to compute
+   * vsize. Input data like prevout and sequence count as base data
+   * and must be added in outside this function.
```
pinheadmz, line 3309: nit: `in`

PR #709, opened 2022-04-10
Removed browser folder and minor JSDoc changes

Anunayj:
Found a minor mistake in JSDoc.
Removed browser folder left over from bcoin.

### DIFF
--- a/lib/primitives/tx.js
+++ b/lib/primitives/tx.js
@@ -1893,7 +1893,7 @@ class TX extends bio.Struct {
    * size as it is framing (exposed on return value as `witness`).
    * @private
    * @param {BufferWriter} bw
-   * @returns {Sizes}
+   * @returns {BufferWriter}
    */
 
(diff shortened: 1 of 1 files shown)


lib/primitives/tx.js
```diff
@@ -1893,7 +1893,6 @@ class TX extends bio.Struct {
    * size as it is framing (exposed on return value as `witness`).
    * @private
    * @param {BufferWriter} bw
-   * @returns {Sizes}
```
pinheadmz, line 1896: Well this function still returns something doesn't it?
Probably `@returns {BufferWriter}` or `@returns {Buffer}`

Anunayj: Yup, it returns a `{BufferWriter}`, I was gonna add a return statement but then it looked like the BufferReader didn't have a `@returns` either so I left it like that, though on second thoughts, I should fix that.

Issue #714, opened 2022-04-14
Can't Install Goosig : error: ‘bcrypto_init’ undeclared here (not in a function)


alifianadexe:
Hey mate, im trying to install hsd and hsd-airdrop but why this get error on goosig. Here the full log : 

```
In file included from ../src/goosig.c:12:0:
../src/goosig.c:234:35: error: ‘bcrypto_init’ undeclared here (not in a function)
 NAPI_MODULE(NODE_GYP_MODULE_NAME, bcrypto_init)                  \
                                   ^
/usr/include/nodejs/src/node_api.h:92:7: note: in definition of macro ‘NAPI_MODULE_X’
       regfunc,                                                        \
       ^~~~~~~
../src/goosig.c:234:1: note: in expansion of macro ‘NAPI_MODULE’
 NAPI_MODULE(NODE_GYP_MODULE_NAME, bcrypto_init)                  \
 ^~~~~~~~~~~
../src/goosig.c:238:1: note: in expansion of macro ‘NAPI_MODULE_INIT’
 NAPI_MODULE_INIT() {
 ^~~~~~~~~~~~~~~~
../src/goosig.c:238:1: warning: missing initializer for field ‘reserved’ of ‘napi_module {aka struct <anonymous>}’ [-Wmissing-field-initializers]
In file included from ../src/goosig.c:12:0:
/usr/include/nodejs/src/node_api.h:59:9: note: ‘reserved’ declared here
   void* reserved[4];
         ^~~~~~~~
../src/goosig.c:235:19: warning: ‘goosig_init’ defined but not used [-Wunused-function]
 static napi_value goosig_init(napi_env env, napi_value exports)
                   ^
../src/goosig.c:238:1: note: in expansion of macro ‘NAPI_MODULE_INIT’
 NAPI_MODULE_INIT() {
 ^~~~~~~~~~~~~~~~
goosig.target.mk:98: recipe for target 'Release/obj.target/goosig/src/goosig.o' failed
make: *** [Release/obj.target/goosig/src/goosig.o] Error 1
make: Leaving directory '/home/adexe/airdrop/hs-airdrop/node_modules/goosig/build'
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/share/node-gyp/lib/build.js:258:23)
gyp ERR! stack     at emitTwo (events.js:126:13)
[...]

PR #715, opened 2022-04-19
ci(tests): Update nodejs versions

Falci:
[Node v18.0.0 (Current)](https://nodejs.org/en/blog/release/v18.0.0/)

Update LTS versions: https://nodejs.org/en/about/releases/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-          node-version: 12.x
+          node-version: 16.x
     - run: |
         npm install -g @hns-dev/bsdoc
         jsdoc -c jsdoc.json
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [12.x, 14.x, 16.x, 18.x]
 
     steps:
 
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "wallet"
   ],
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=12.0.0"
   },
   "dependencies": {
     "bcfg": "~0.1.7",


handshake-enthusiast:
In the meantime https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/docs/install.md:
>`hsd` requires Node.js v12 or higher.

1. Should we update it as well then?  
2. I'm not from the Node world. Does EOL mean that there is a strong reason to stop testing against it?  
3. Should https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/.github/workflows/docs.yml#L14 be updated as well?

handshake-enthusiast:
I see that it's `ci(tests)`, but still:  

4. Should [package.json](https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/package.json#L20) and [package-lock.json](https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/package-lock.json) be updated as well?
```json
  "engines": {
    "node": ">=8.0.0"
  },
```

For me it's quite confusing that something is being tested against newest versions while still allowing old versions where the installation likely fails.

handshake-enthusiast:
A recent issue https://github.com/handshake-org/hsd/issues/714#issuecomment-1099156888 which is related to my comment above.

handshake-enthusiast:
https://handshake.org/download/ says:
>Node.js version 10+ is required to run: Execute `hsd-3.0.1/hsd/bin/hsd`  

([source code](https://github.com/handshake-org/handshake-web/blob/5e84d2ba67f2a4f02bdc22a9e8778b2a08fae260/download/index.html#L101))

pinheadmz:
All good points @handshake-enthusiast I see from nodejs website that v12 is EOL in just a few days! (April 30)

So it does seem prudent to update all the places you found, 14 is the new minimum and that should be reflected in package.json, the docs, and all the CI scripts.

nodech:
You could just add new version on the matrix:
```
        node-version: [12.x, 14.x, 16.x, 18.x]
```

> In the meantime https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/docs/install.md:
> 
> > `hsd` requires Node.js v12 or higher.
> 
>   1. Should we update it as well then?
> 
>   2. I'm not from the Node world. Does EOL mean that there is a strong reason to stop testing against it?
> 
>   3. Should https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/.github/workflows/docs.yml#L14 be updated as well?
>  4. Should [package.json](https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/package.json#L20) and [package-lock.json](https://github.com/Falci/hsd/blob/1fc2bc1f3ba3fab4dc81e8b93a7a049d0d409b40/package-lock.json) be updated as well?

1. I believe we don't. It just needs to state what's the minimum version supported. (If someone wants to use older machine/node after EOL it's their choice)
2. I believe we should test it against minim required version as well, because if it ever breaks we will see it in the tests and maybe fix it or maybe drop the support altogether.
3. That is only generating documentation, so it does not matter.
4. Yes, we need to update package.json to show that hsd needs at least nodejs v12.

Falci:
Ok, here's the recent changes:

* added 12x back to test table;
* updated engine to >= 12.x
* build docs using 16.x

PR #731, opened 2022-06-07
mnemonic: standardize phrase

rithvikvibhu:
`Mnemonic` verifies words by stripping each word, but stores phrase user input as-is.

image

And when `.toSeed()` is used by HD key: https://github.com/handshake-org/hsd/blob/ba949f348c58ec6a6a6231f411fe3caa7c5e7b80/lib/hd/private.js#L460
the new lines are included and these two phrases are considered different and create different wallets.

For context, a bob user faced this and got two different wallets.

This commit saves the validated words -> phrase instead of the phrase passed as argument.
Added a test that covers both single space and `\u3000` japanese space.

### DIFF
--- a/lib/hd/mnemonic.js
+++ b/lib/hd/mnemonic.js
@@ -34,7 +34,7 @@ class Mnemonic extends bio.Struct {
    * @constructor
    * @param {Object} options
    * @param {Number?} options.bit - Bits of entropy (Must
-   * be a multiple of 8) (default=128).
+   * be a multiple of 8) (default=256).
    * @param {Buffer?} options.entropy - Entropy bytes. Will
    * be generated with `options.bits` bits of entropy
    * if not present.
@@ -259,7 +259,12 @@ class Mnemonic extends bio.Struct {
     this.bits = bits;
     this.language = lang;
     this.entropy = entropy;
-    this.phrase = phrase;
+
+    // Japanese likes double-width spaces.
+    if (this.language === 'japanese')
+      this.phrase = words.join('\u3000');
+    else
+      this.phrase = words.join(' ');
 
     return this;
   }
--- a/test/mnemonic-test.js
+++ b/test/mnemonic-test.js
@@ -61,4 +61,25 @@ describe('Mnemonic', function() {
     assert.strictEqual(m2.language, m1.language);
     assert.bufferEqual(m2.toSeed(), m1.toSeed());
   });
+
+  it('should standardize phrase', () => {
+    for (const language of Object.keys(tests)) {
+      const test = tests[language];
+
+      // modified phrase contains new lines before every space
+      const phrase = test[0][1];
+      const phraseWithNl = phrase.replace(/(\s|\u3000)/g, '\n$&');
+
+      // ensure new lines are added
+      assert(!phrase.includes('\n'));
+      assert(phraseWithNl.includes('\n'));
+
+      const m1 = Mnemonic.fromPhrase(phrase);
+      const m2 = Mnemonic.fromPhrase(phraseWithNl);
+
+      // Both mnemonics should have same phrase without new lines
+      assert(m1.getPhrase() === m2.getPhrase());
+      assert(!m1.getPhrase().includes('\n'));
+    }
+  });
 });


rithvikvibhu:
TIL `replaceAll` was introduced in node v15 so was failing on v14 test. Replaced with regex replace().

Anunayj (approved):
ACK

Anunayj:
Also the Test Vector can use some additions, like phrases with newlines, ASCII whitespaces in Japanese phrases.

pinheadmz:
Haven't reviewed your changes yet, but this patch ensures that a "dirty" phrase will not result in an incorrect seed:

```diff
diff --git a/lib/hd/mnemonic.js b/lib/hd/mnemonic.js
index 5a95f2fd8..e9cfbdfa8 100644
--- a/lib/hd/mnemonic.js
+++ b/lib/hd/mnemonic.js
@@ -146,9 +146,6 @@ class Mnemonic extends bio.Struct {
    */
 
   getPhrase() {
-    if (this.phrase)
-      return this.phrase;
-
     // Include the first `ENT / 32` bits
     // of the hash (the checksum).
     const wbits = this.bits + (this.bits / 32);

```

This basically forces the module to always use the `entropy` value, which is (I think) always CORRECT since it is derived from the words in the phrase more carefully. In fact, if we go this route there will be NO NEED for `mnemonic.phrase` to exist at all, because we can always just re-derive it from the entropy. It can be cached maybe but the cached value should only be generated internally and NOT directly from user input :-P

lib/hd/mnemonic.js
```diff
@@ -259,7 +259,12 @@ class Mnemonic extends bio.Struct {
     this.bits = bits;
     this.language = lang;
     this.entropy = entropy;
-    this.phrase = phrase;
+
+    // Japanese likes double-width spaces.
+    if (this.language === 'japanese')
+      this.phrase = words.join('\u3000');
+    else
+      this.phrase = words.join(' ');
```
pinheadmz, line 267: I'm afraid this is going to cause the same problem over again (where different valid strings produce different seeds)

BIP39 states:
> To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt. 

So, I need you to check this: does this NFKD encoding specify anything about spaces?

pinheadmz: Since this \u3000 thing is also present in `getPhrase()` we should make sure its really the standard, otherwise Japanese phrases from hsd will be incompatible with other software

pinheadmz: Doing some research on this:
[Electrum](https://github.com/spesmilo/electrum/blob/0fca35fa4068ea7c9da60cd5330b85c2ca1d0398/electrum/mnemonic.py#L173) only uses single space for this. Although Electrum uses an alteration of bip39 anyway...

rithvikvibhu: Trezor (the reference implementation in BIP32, does the same thing that hsd does (japanese double space, single space for others)
https://github.com/trezor/python-mnemonic/blob/master/src/mnemonic/mnemonic.py#L70

pinheadmz: hm ok

test/mnemonic-test.js
```diff
@@ -61,4 +61,21 @@ describe('Mnemonic', function() {
...
+      // modified phrase contains new lines before every space
+      const phrase = test[0][1];
+      const phraseWithNl = phrase.replace(/(\s|\u3000)/g, '\n$&');
+
+      const m1 = Mnemonic.fromPhrase(phrase);
+      const m2 = Mnemonic.fromPhrase(phraseWithNl);
+
+      // Both mnemonics should have same phrase without new lines
+      assert(m1.getPhrase() === m2.getPhrase());
+      assert(!m1.getPhrase().includes('\n'));
```
pinheadmz, line 78: wanna add sanity checks here too?

```js

      assert(phraseWithNl.includes('\n'));
      assert(!phrase.includes('\n'));
```

rithvikvibhu: Added.

lib/hd/mnemonic.js
Anunayj: > Since this \u3000 thing is also present in `getPhrase()` we should make sure its really the standard, otherwise Japanese phrases from hsd will be incompatible with other software

I checked this a few days ago and even made a comment (now deleted), when deriving the seed phrase there is a step where it does string.normalize('nfkd'), which converts `\u3000` to ASCII. I checked with other wallet implementations too, and got the correct xpriv.

PR #732, opened 2022-06-08
CI - add macos and windows to the test matrix.

nodech:
Separate linter, doc build (do we need it here?) and coverage.

Add windows and macos to the test matrix. With this it should test all versions against all OSes listed.


.github/workflows/node.js.yml
```diff
@@ -38,3 +36,55 @@ jobs:
...
+        node-version: ${{ matrix.node }}
+
+    - name: Update npm
+      run: npm i -g npm
+
+    - name: Install
+      run: npm install
+
+    - name: Test
+      run: npm run test
```
Falci:
```suggestion

```

.github/workflows/node.js.yml
```diff
@@ -38,3 +36,55 @@ jobs:
       uses: coverallsapp/github-action@master
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  build:
+    name: Build & Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
```
Falci, line 45:
```suggestion
        os: [ubuntu-latest, macOS-latest, windows-latest]
```

Falci (commented):
Can we have windows in the same list as ubuntu and macos?

pinheadmz:
The doc build is actually a test of sorts to ensure that jsdocs comments are all properly formatted. And then when we tag a release, it automatically builds and deploys to hsd-dev.org/hsd/

nodech:
> Can we have windows in the same list as ubuntu and macos?

It was in the same list, I was trying to debug why windows failed, so moved out. I will close this PR for now and test on my fork before reopening.

PR #738, opened 2022-06-10
Release v4.0.0-rc.1

nodech:
Check [Release Notes Draft](https://github.com/handshake-org/hsd/blob/8047df7130af7c00ba49d4fe7aac9ec797b23616/docs/release-notes/release-notes-v4.x-draft.md) (Draft because it's release candidate)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hsd",
-  "version": "4.99.0",
+  "version": "4.0.0-rc.1",
   "description": "Cryptocurrency bike-shed",
   "license": "MIT",
   "repository": "git://github.com/handshake-org/hsd.git",
@@ -22,7 +22,7 @@
   "dependencies": {
     "bcfg": "~0.1.7",
     "bcrypto": "~5.4.0",
-    "bdb": "~1.3.0",
+    "bdb": "~1.4.0",
     "bdns": "~0.1.5",
     "bevent": "~0.1.5",
     "bfile": "~0.2.2",
@@ -49,7 +49,7 @@
     "urkel": "~1.0.1"
   },
   "devDependencies": {
-    "bmocha": "^2.1.5"
+    "bmocha": "^2.1.6"
   },
   "main": "./lib/hsd.js",
   "bin": {
--- a/docs/release-notes/release-notes-v4.x-draft.md
+++ b/docs/release-notes/release-notes-v4.x-draft.md
@@ -1,0 +1,95 @@
+v4.x Release notes
+==================
+
+<!-- toc -->
+
+- [v4.0](#v40)
+  * [How to Upgrade](#how-to-upgrade)
+  * [Notable Changes](#notable-changes)
+    + [Node and Chain](#node-and-chain)
+    + [Wallet changes](#wallet-changes)
+    + [DNS](#dns)
+  * [Changelog](#changelog)
+
+<!-- tocstop -->
+
+# v4.0
+## How to Upgrade
+  When upgrading to this version from **hsd v3.x** you must pass `--chain-migrate=3` when you run it for the first time.
+  If you are upgrading from **hsd v2.x**, then you also need to pass `--wallet-migrate=1`.
+
+## Notable Changes
+### Urkel Tree Compaction
+  `FullNode` parses new configuration option `--compact-tree-on-init` and `--compact-tree-init-interval` which will compact the Urkel Tree when the node first opens, by deleting historical data. It will try to compact it again after `tree-init-interval` has passed. Compaction will keep up to the last 288 blocks worth of tree data on disk (7-8 tree intervals) exposing the node to a similar deep reorganization vulnerability as a chain-pruning node. [#669](https://github.com/handshake-org/hsd/pull/669)
+  
+  This feature is disabled by default. Similar to `prune` option of the FullNode, **this will limit the reorganization ability of the node** but will save disk space. Different from `prune`, this does not happen retroactively given the nature of the Urkel Tree, that's why it is called `compaction` which is better term for it. If enabled, compaction will only happen when node starts/restarts. To enable use `--compact-tree-on-init`.
+  
+  After compaction next compaction will run after `--compact-tree-init-interval` blocks have passed and it will also only trigger after start/restart. By default it is set to `10 000` blocks.  
+  Example: `hsd --compact-tree-on-init --compact-tree-init-interval=10000` - this is tree compaction enabled handshake node.
+  
+| Height  | Full Tree Size(no reorg) | Compacted Size |
+| ------  | ------------------------ | -------------- |
+| 100 000 | 6.73 GB                  | 511.37 MB      |
+| 105 000 | 8.04 GB                  | 597.22 MB      |
+| 110 000 | 9.89 GB                  | 727.02 MB      |
+| 115 000 | 11.78 GB                 | 859.22 MB      |
+| 120 000 | 13.40 GB                 | 976.03 MB      |
+
+*Note: Compaction time depends on the compacted size, not the full tree size.*
+  
+
+### Node and Chain
+  - `FullNode` and `SPVNode` now accept the option `--agent` which **adds** a string to the user-agent of the node (which will already contain hsd version) and is sent to peers in the version packet. Strings must not contain slashes and total user-agent string must be less than 255 characters. [#710](https://github.com/handshake-org/hsd/pull/710)
+    - Example: `hsd --agent=example` will give you node with agent: `/hsd:4.0.0-rc.1/example/`
+  - Fix SPV node reorg sync issue. SPV will also check `MTP` and `time-too-new`. [#721](https://github.com/handshake-org/hsd/pull/721)
+  - `hsd` will critically shutdown on Disk write failure or corruption. `FullNode` and `SPVNode` will emit `abort` event on these failures so embedders can use these events to clean up before shutdown. [#650](https://github.com/handshake-org/hsd/pull/650)
+  - `hsd` will recover tree root cache to maintain consistent look up times later for the proof requests. (Making startup time slightly slower) [#726](https://github.com/handshake-org/hsd/pull/726)
+  - Checkpoints up to `100000` block. [#734](https://github.com/handshake-org/hsd/pull/734)
+  - Fix blockstore atomicity issue, resolves rare **Block not found** issue when process is killed during write. [#688](https://github.com/handshake-org/hsd/pull/688)
+  - remove inflation bug softfork covert signaling. [#654](https://github.com/handshake-org/hsd/pull/654)
+
+### Wallet changes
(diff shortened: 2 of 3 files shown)


befranz:
Thinking about to adopt my install/update script for it. All the full nodes installed with the script are hsd v3.01.
Do I get it right that I pass all options ```--compact-tree-on-init```, ```—compact-tree-init-interval``` and probably ```--agent``` as command parameters and not as an option in the config file hsd.conf?
Do I have to remove ```—chain-migrate=3```and ```--compact-tree-on-init``` after the first initial start of hsd v4?
And can I leave ```—compact-tree-init-interval``` as default for every start?

nodech:
> Thinking about to adopt my install/update script for it. All the full nodes installed with the script are hsd v3.01. Do I get it right that I pass all options `--compact-tree-on-init`, `--compact-tree-init-interval` and probably `--agent` as command parameters and not as an option in the config file hsd.conf? Do I have to remove `--chain-migrate=3`and `--compact-tree-on-init` after the first initial start of hsd v4? And can I leave `--compact-tree-init-interval` as default for every start?

Thanks for the feedback. I guess these features need more description.
I added more details to these features.

  Let's start with update of the scripts. This is Release Candidate (beta) and we will take some time (around a week maybe?) after we release and get feedback from the users who use beta. If we don't have any issues we will release `v4.0.0`. I believe it would be better to update installation script after `v4.0.0` is out of beta.

  `--agent` is highly unlikely that users will need to use. It is mostly meant to signal custom clients that have either modified, embeded or extended `hsd` in some way but are part of the network and want to signal it. Most of the time even these nodes may not want to signal it. For example, Bob wallet could theoritically have listening nodes (which I believe they don't) - In that case they could signal `/hsd:version/bob:version/` - But in case there's issue with bob side software they can easily targeted and we are talking about "normal users" and not "super users" who can do bare necessary to secure their nodes. So I would say installer that is meant to be for everybody, does not need special agent.
  
  `--compact-tree-on-init` and `--compact-tree-init-interval` are similar to the `prune` option. So I would say, it depends on the configuration you are going for. For example, if you want Full Node in its truest sense - which needs to have fulll history of the tree and be able to provide proofs for the tree root in the history - then you don't use these options. Similar to how you may not use `prune`. If you are going for "disk space efficiency" - and take every optimization you can get from hsd side then you will enable both: `--compact-tree-on-init` and `--prune`.  
  I tried to expand description of these two options in the release notes, you can check updated docs.
  
  For the question of `hsd.conf` - anything that can be passed as a shell argument to the `hsd` can be provided in `hsd.conf`. `compact-tree-on-init`, `compact-tree-init-interval` and `agent` are all the options that will go to the `hsd.conf`.
  
  I believe there was some doc describing how it worked, but I will just try to explain here anyway. When we have database changes in the codebase (as in v4), node will not start if the migration has not happened. In general this flag is not necessary, this is only used to tell the hsd to run the migrations that are pending and you provide the latest migration ID in that. If the ID is not the latest, hsd will complain. For example users who upgraded from `v2` to `v3` may already have`chain-migrate: 2` in their configuration file, they can just bump that version. If you don't want to have it in the `hsd.conf`, you can start node with `--chain-migrate=3` once and let it run or even restart after it's done w/o argument.
  So basically, that flag is only necessary to be passed when node has not done migration. That's the way we tell users to take action and also a way for users to reconsider if you want to upgrade now. After migration is done, you can either remove that flag or leave it. It wont complain if you pass `--chain-migrate=3` and it has already done migratoins.  
  Note fresh nodes don't need to migrate, they are already using latest database scheme.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hsd",
-  "version": "4.99.0",
+  "version": "4.00.0-rc.1",
```
pinheadmz, line 3: does this middle number need to be single `0`? I dont care just wondering

pinheadmz (approved):
ACK ae9edbf6318580a70a775bd250ea1975ca87fe1e

```
Hash: SHA256

ACK ae9edbf6318580a70a775bd250ea1975ca87fe1e
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKkre0ACgkQ5+KYS2KJ
yTpl/Q//TILGE1X88XxFDRfqQsK0uzFzj5HGJXBszrkLwgb8tkol1exkt4iToCA3
RaN1SzWsCPaSwZ60zzvkKOEAPgc0jYko7MZt9pgOGYiF+9eKGEyPitABDWhvF8Hb
cVGAHm8uyyN08khnSDXI2jrsn1Z7JaUP6KNf8Igb647PTqvfVmdEkXSzcBIXnZPd
r2RKuDLeMJKcL3Uw3TLIAEVMjk3CJpaUi6BRwPA25hvxoGvbLoBNf6PrdTfL26Bn
BCFUfw89Icwo50L/OPIix7WkuEzIUy7IokY7kyJh88mn5Wp+rQDEb1f33sDiXkrg
U2xv690DKgGAesIzRboBcpygdGA7tS2yQn0n2fDdwg5EzJjt0EAIAXnoVcJX9HVo
rk69+oHfPDC2KMs63bhm2UPgn9/gtMdtv8MV48ZZZ9w7M4R+q2NYSfu85MgiCKgb
QMmHqjW3tQdGSdeCr3wWvGJmegBgrLuqHQ7DJ8lBGOk9hHQso+C9UC9F3ANsENvA
Z8qw9zXOq9dD2CTgcdKHJcvhGczuIpZAdB2hJB5VhyVMuXjMzlJn82snENxFTAmh
/FyYMJcg4TdJMblHAH3rDd+qxnTKQ9npiU/mIqgHr0TsyY2jxxIG6KEvLAYqQKjS
gYqqWMacGbi3PZFn7TmBE5iXcS773G2EbErp/T5ar6JyiZcPGU4=
=moLh
-----END PGP SIGNATURE-----
```

</details>

package.json
nodech: yep, it should. typo

pinheadmz (approved):
ACK 76a7eea11b2bab3e283d8cf25181076db5376e97

```
Hash: SHA256

ACK 76a7eea11b2bab3e283d8cf25181076db5376e97
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKkr0QACgkQ5+KYS2KJ
yTpboxAAizxA/ATXCkT1ho8X6SV3vrUnFLjn/E1Y6SGVEIcfqFBngih5Z8La/I1R
Q8AoCKLplh2ow1seMy9ekUPkH/wSi2587hK08KB8+HzKvUHuvicwSzd6IVnURKmN
AnpuHAzAMemH+qFBi1oUPi8+Cx0jYVBSC0HNf2ODhJXJOdTOrCL7/sMqggzHc3vo
MroJCa9n2aNsBNVG9nuS8oaV//7V6wAyYk7tQ1D5bush5qqrEEawphKh4pRR+MaP
mdH88hN0AmyqLy8kvgPFnFw4LelusX7FPCV+/EA98r39rvAPd/4/FFaOcj3KCSHI
ZBURI1WnYWxv6xJYSWSQ6ilCZR5xZqWmeDFiPKjZdMqpX7aRQn2pNa+aAVhb3CA9
TTAbrpWCur1DpkCXWRvGgptYaR5C/NoZ+RMigvGpSKsaKuMcWkzIZ4x8K1t4jwhV
ogDtV1Gd/cMpcyCRyUaIOgkxD6PBFcN47oq9rBore4fSsuQLc4dj/7cnEm+KMP0J
LY1V55K5jd4t0/Zg1wpHydCfwiVhYuQE0nG8zUTVzQy1L7G9OvnfOdfoMtvSNCTF
73uVYSXgbGfnbTKohGeTiB3O5QhIOmKg8b39WYNAXHkfrfCn1plJG2Nt3Zj18vyY
CkU0BNDeRs3ajzWc+MRAUvBQMRlPotz3IXyiNM65iI2HW5UBLa0=
=uUi8
-----END PGP SIGNATURE-----
```

</details>

pinheadmz (approved):
ACK 9f97feee959adaf8c9319f85ae963ba4348f578f

```
Hash: SHA256

ACK 9f97feee959adaf8c9319f85ae963ba4348f578f
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmKkweIACgkQ5+KYS2KJ
yTpwPRAAjgEsQJunnUycySVkvGdUoSo0ldekMdrTZPVRnc6M2S9MDKVniORoKQ/T
ojoK8BLjUqZNi+gNFkgVLQhGrqsmSERVbC3uAYXzP6DkWYqiMrONoOnlaGccQ0gc
x0bHxUvd3PT3x0Y2MucVQ/hQ3PkBX7B25PN3PuMb5LOuSc42dYwcO+Ml9HSE/DaB
2ALOpeF+tzCVhMEY2xQHzqzsRNuoKoc1Sm6eYs3azrdGC/VLJVpbNgUiSHcWNYZd
YYdWeLyt4SFfmWzOO2Q58VU4iZdhcPf6fDPHNd1waPunNSkRc5OnHlLdxrc47Acm
LeZdXwBT+xvjeBRmk6vkuEnzhmNvXaAVcaAPqbN3vCrF1/yxSInVxz56LpeUWFLQ
G6Fuee8hEAuG5IZhTQTzo00HgQJ5fky2rpk6WEDG0NtyLEr6RlUIKcBokZRaSAPs
GWY5w7Bkij23b4f3z4s/ApdL2fjzfbRd0LvExjp9RR7mx5wx1+h2YGQhD4eiYnYP
ELOMYJLmL3R4cPYNbDtRvcQtykgOr/VI5ro8zki15P8x4QkpCf/mQsPHM+Vm7ro+
gDeqiDygm+OJJliM4tt/cVeLeT19BtMEvodOS6oAIfEyJU8MZqiB5YOXLVuRjl9b
BxzRTw7bRWA1wyLRorPtcrvCQfzmGBL/bFwimGuUcobecxbJRfA=
=68Dn
-----END PGP SIGNATURE-----
```

</details>

befranz:
> `--compact-tree-on-init` and `--compact-tree-init-interval` are similar to the `prune` option. So I would say, it depends on the configuration you are going for. For example, if you want Full Node in its truest sense - which needs to have fulll history of the tree and be able to provide proofs for the tree root in the history - then you don't use these options.

Thanks for your comprehensive answer. My script is basically for public listening full nodes, so this means those nodes can not benefit from the compaction, and a full tree history is needed to serve as such?

pinheadmz:
IMO nodar is exaggerating with "full node in the truest sense." ;-)

All full nodes can safely compact the tree. The risk is that they won't survive a reorg greater than 250 blocks

befranz:
> The risk is that they won't survive a reorg greater than 250 blocks

I see, we need at least a few nodes not compacted for the worst case. I guess easyhandshake is such a node. But which else, would it make sense to know/identify them?

pinheadmz:
We considered adding a service flag to indicate limited tree history (we need to do this with pruned nodes as well) but a compacted tree is certainly enough to serve the network. 

In the event that a node operator wants to rebuild the tree completely this is actually available with an undocumented rpc reconstructtree and requires a full archival node.

PR #739, opened 2022-06-10
Update release process.

nodech:
Accumulate experience from our first release since release docs and update.


pinheadmz:
I think we should commit the relese-notes file to master along with bumping version on master to `X.99.0` and forking to the release-proposal brancgh

PR #741, opened 2022-06-12
4.x Clean Install in case of need

miguelgargallo:
In case, someone needs to clean install, because some dependencies deprecated or just malfunctioning, and the end user needs to clean install, here are the steps to perform a clean install on the latest technologies, Ubuntu 22.04

### DIFF
--- a/docs/release-notes/release-notes-v4.x-draft.md
+++ b/docs/release-notes/release-notes-v4.x-draft.md
@@ -1,0 +1,95 @@
+v4.x Release notes
+==================
+
+<!-- toc -->
+
+- [v4.0](#v40)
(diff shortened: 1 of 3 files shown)


pinheadmz:
NACK, what is a "clean install" anyway?
`npm install` and `npm rebuild` should be enough. I don't want to advise anyone delete directories. Maybe we could add an "upgrading" section to https://github.com/handshake-org/hsd/blob/master/docs/install.md

Most importantly I think if you have already cloned hsd, the `latest` tag is funky because we force-push to update where `latest` points. So maybe for upgrading users can `git checkout 4.x` for example

PR #746, opened 2022-06-20
bin: replace shell scripts with js require

rithvikvibhu:
see comment below.

~~Main PR in hs-client, go here for more info:~~
- ~~https://github.com/handshake-org/hs-client/pull/51~~

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -4,11 +4,11 @@
 
 console.error('%s%s',
   'Warning: The `hsd cli` interface is deprecated.\n',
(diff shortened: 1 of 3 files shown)


rithvikvibhu:
Now that hs-client is part of hsd, even this is simpler.
- `./cli` referred to hs-client which is no longer a dep of hsd, changed that
- updated the 2 rpc commands to call cli
- closing the hs-client PR since it's covered here.

PR #751, opened 2022-06-26
How to upgrade description added to the README.md

Sh1Tt:


README.md
```diff
@@ -92,6 +92,28 @@ In order to notify peers that you are accepting inbound, you _must_ pass
...
+$ git pull
+```
+Select a branch and upgrade. 
+
+```
+$ git checkout <branch>
+$ npm install
+```
+
+After installation of the new version, activate it by restarting the HSD.
```
pinheadmz, line 110:
```suggestion
After installation of the new version, activate it by restarting `hsd`.
```

pinheadmz (changes requested):
I think this section should be moved to `/docs/install.md` and maybe look into using `git fetch --tags --force` because we noticed on v4 release that moving the `latest` tag from an old commit to a newer one caused some issues.

PR #755, opened 2022-06-30
build(Docker): Update base image and dependency

Falci:


pinheadmz:
confirmed the docker build was broken and is now fixed. also updated part of the directions which otherwise work great for running in docker. Thanks @Falci

PR #767, opened 2022-09-12
wallet: derive all keys in generateNonce for multisig

rithvikvibhu:
Multisig accounts will generate different nonce given the same nameHash, p2sh address and value as only "own" public key is used. This PR derives public keys from all keys and picks the lexicographically smallest one so all parties generate the same nonce. Implements @pinheadmz's suggestion: https://t.me/hns_tech/70247

### Tests

2 tests are added, one for regular pubkeyhash accounts and another for 2-of-2 multisig. Assertions compare to specific hardcoded values to ensure that this PR is not a breaking change (at least for regular wallets).

Running the test on master passes the first (pubkeyhash account), but fails on the second (multisig) because alice and bob generate different nonces.


test/wallet-test.js
```diff
@@ -3273,4 +3273,63 @@ describe('Wallet', function() {
...
+      assert(addr.toString(network) === 'hs1q8nnrenk92qdmm32zfv56shvhj6n6g8kq3ermpz50k5zxh4qjq34qk7v4d2');
+
+      // Generate nonce for same inputs from both wallets
+      const aliceNonce = await alice.generateNonce(nameHash, addr, value);
+      const bobNonce = await bob.generateNonce(nameHash, addr, value);
+      assert(Buffer.isBuffer(aliceNonce));
+      assert(Buffer.isBuffer(bobNonce));
+
+      // Same nonce is generated by alice and bob
+      assert(aliceNonce.equals(bobNonce));
```
pinheadmz, line 3331: can be `assert.bufferEqual(aliceNonce, bobNonce)`

pinheadmz (commented):
Great solution, elegant code patch. I wonder if this should be published as (maybe very short) HIP since it's an application-level standard we'll need for interoperability with new HNS wallets. In addition, such a HIP may also want to begin to describe some kind of PSBT format that includes bid values from the tx author to the tx signers. Hopefully we have a future where big multisig corporations are bidding huge amounts on names and this kind of convinience would be helpful.

rithvikvibhu:
>I wonder if this should be published as (maybe very short) HIP since it's an application-level standard we'll need for interoperability with new HNS wallets.

:+1: Will have a draft ready along with Bob's multisig PR

pinheadmz:
It just occurred to me that this is also a potentially breaking change. A multisig party may have already placed a bid with their own key's nonce and instead of the group key like you've implemented. Attempting to recover that wallet using `importnonce` after this PR is merged would not work.

I'm not sure what the prettiest code would be but to be totally covered I think the logic for `importnonce` may need to derive both old and new nonces if the wallet is multisig. As long as both nonces are stored in the db, the reveal action will find the correct one and finish the TX.

rithvikvibhu:
I looked for all places `generateNonce` (and by extension `generateBlind`) are used:
- **Get Nonce for Bid** HTTP API ([docs](https://hsd-dev.org/api-docs/?shell--curl#get-nonce-for-bid))
- **importnonce** RPC method
- **makeBid** in wallet

So yes, it's a breaking change when:
- it is a multisig account, and
- has placed bids that: (a) aren't revealed yet, and (b) can still be revealed, and
- the wallet is recovered from seed / ledger after the bid was placed, and
- own key is not the first key lexically (but unlikely, so we can ignore this point)

Storing both nonces in db doubles what's stored on disk for all current multisig wallets as well as those created in the future.
Instead, how about one of these (in the order I prefer them, first being the best of them):

Option 1:
- new param to `importnonce` like `multisig-legacy: bool` that uses the old generateNonce (defaults to false)
- same change to the HTTP api
- Maybe: when this is true, we generate both old and new. If false, then only new.

Option 2:
- a script/api-method/function/whatever that reads all existing `BidBlind`s and generates old nonces for them
- useful if one wants to `importnonce` (many bids) without change and in the end call this to generate and store old nonces

Option 3:
- no invisible upgrade path, warn users not to have active and unrevealed bids (in multisig) before update
- I'm considering this as an option (even if it's last) because multisig isn't widely used yet and regular wallets aren't affected
- in case a wallet really needs to be recovered, a script that does option 2 can be run separately

I like the first one. Was thinking of a wallet migration or similar, but it wouldn't work here because it affects _new_ wallets, whereas migrations are for existing ones.

pinheadmz:
I like the direction of option 1, just adding a `--legacy` flag or something. I think what that could do is generate and store nonces using _every public key in the multisig set_. That would ensure that ANY member of the party would be able to repair these bids, assuming they know the secret bid value.

We may even just want to do this without a flag (by default) since it is a bit of an edge case anyway. That would just be for rpc/http`importnonce` and not all `generateNonce`'s.

> Storing both nonces in db doubles what's stored on disk for all current multisig wallets as well as those created in the future.

I'm not too worried about storing a few extra hashes in the DB for now. Truth is, the wallet DB stores a ridiculous amount of extra data (e.g. everyone else's bids on auctions you participated in, future bids on re-opened names, etc) and pruning this garbage will be another big task that will increase in urgency as time goes on...

rithvikvibhu:
Okay so now the way it works is:

>**Note:** Doesn't affect pkh accounts, only multisig.

Normal behavior (makeBid, etc.):
- Smallest public key is used to generate nonce
- Only 1 blind is stored based on that nonce

Importnonce RPC/HTTP behavior (`forAllKeys` is `true`):
- `generateNonce` returns N nonces (one for each key)
- `generateBlind` stores N blinds, but returns the own-pub-key blind only
- This way, the importnonce response doesn't change and is backward compatible

lib/wallet/wallet.js
```diff
@@ -1197,10 +1197,11 @@ class Wallet extends EventEmitter {
    * @param {Buffer} nameHash
    * @param {Address} address
    * @param {Amount} value
-   * @returns {Buffer}
+   * @param {boolean} forAllKeys
+   * @returns {Buffer|Buffer[]}
    */
 
-  async generateNonce(nameHash, address, value) {
+  async generateNonce(nameHash, address, value, forAllKeys) {
```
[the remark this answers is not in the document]
rithvikvibhu: Yeah I didn't like it either. Now it has `generateNonce(s)` and `generateBlind(s)`. But `generateNonce` cannot call its plural method [0] because the order of the nonces/blinds is different:
- hsd behavior has been own public key everywhere
- now we switch to smallest public key

So generateNonce has the new style (smallest) whereas generateNonces returns first based on own key so that there's no breaking change.

>The http call is kinda funny because it doesn't add anything to db ... That would make this PR a breaking change, though.

I don't think Bob uses that remote call since it always has access to the wallet directly. But since it's breaking, can be kept aside for a different PR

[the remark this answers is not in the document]
rithvikvibhu: Also, each commit here undos a lot of code from the previous commit so might be easier to just look at overall PR changes. Will squash if the code makes sense now.

lib/wallet/rpc.js
```diff
@@ -2665,9 +2665,11 @@ class RPC extends RPCBase {
     const nameHash = rules.hashName(name);
     const address = parseAddress(addr, this.network);
 
-    const blind = await wallet.generateBlind(nameHash, address, value);
+    const blinds = await wallet.generateBlinds(nameHash, address, value);
 
-    return blind.toString('hex');
+    // Return only first blind (based on own key)
+    // to stay backward-compatible
+    return blinds[0].toString('hex');
```
pinheadmz, line 2672: eh, backwards-compat isnt so precious here, lets
```js
return blinds.map(b => b.toString('hex'));
```

It'll be a breaking change I guess, if you only have one key youll get an array with a single item but I'm ok with that. Just needs to go in the changelog.

test/wallet-test.js
```diff
@@ -3273,4 +3273,141 @@ describe('Wallet', function() {
...
+      assert(!await bob.txdb.hasBlind(expectedBlinds.bob));
+
+      // 2) Generate all blinds:
+      const bobBlinds = await bob.generateBlinds(nameHash, addr, value);
+      // own public key is used for blind (backwards-compatible)
+      assert.bufferEqual(bobBlinds[0], expectedBlinds.bob);
+      // but blinds for all key are saved in db
+      assert(await bob.txdb.hasBlind(expectedBlinds.alice));
+      assert(await bob.txdb.hasBlind(expectedBlinds.bob));
+    });
```
pinheadmz, line 3411: I think you should also try to verify that both alice and bob can create a valid reveal even if the nonce is not based on their own key. That might even be best in `wallet-rpc-test.js` since thats the command thats been set up to cover the edge case.

nodech (changes requested):
I can we can avoid duplicate blind as well as additional flags in RPC/API by utilizing almost unused flags in account.

Right now flags have 7 bits unused, we can take 2nd bit for this. We can introduce new flag in account - `e.g. this.multisigBlinds = true` 

Existing wallets in the db, will have this flag set to 0 - meaning old behaviour. New wallets will default to new flag. This adds overhead to techdebt, but avoids all API changes. This does not even need migration. Can be released with minor.

pinheadmz:
> Right now flags have 7 bits unused, we can take 2nd bit for this. We can introduce new flag in account - `e.g. this.multisigBlinds = true`

@nodech  as discussed, this doesn't solve anything because after this is merged, we can't guarantee everyone in every multisig party is running updated software. So in a recovery scenario (with `importnonce`) we still need to try all nonces.

lib/wallet/http.js
```diff
@@ -1029,7 +1029,8 @@ class HTTP extends Server {
       }
 
       const nameHash = rules.hashName(name);
-      const nonce = await req.wallet.generateNonce(nameHash, address, bid);
+      const nonces = await req.wallet.generateNonces(nameHash, address, bid);
+      const nonce = nonces[0];
```
pinheadmz, line 1033: I'm not sure but this endpoint might need to return an array as well. Let's take a look at bob's repair bid functionality. IIRC, we use the HTTP endpoint (which does not save anything to DB) to generate blinds until we get the right one, then call importnonce once with the correct values?

rithvikvibhu: Bob always has access to the wallet and calls `generateNonce` (not `generateBlind` that saves blind). It doesn't depend on RPC/HTTP at all:

```js
const wallet = await this.node.wdb.get(this.name);
...
const nonce = await wallet.generateNonce(nameHash, from, value);
const blind = Rules.blind(value, nonce).toString('hex');
```
https://github.com/kyokan/bob-wallet/blob/ff199207817101f51a9d6065f74f65f637f41c2d/app/background/wallet/service.js#L1093

rithvikvibhu: Oh just saw that the actual repair does use the HTTP API:
https://github.com/kyokan/bob-wallet//blob/master/app/pages/Auction/RepairBid.js#L46-L50

I think we can make Bob use the wallet directly (similar to previous comment) to not depend on HTTP at all,
or will need to expose `generateNonces` plural (requires changes to hs-client also).

lib/wallet/rpc.js
rithvikvibhu: Now everything (internal methods, RPC, HTTP) all use smallest key and blinds and nonces are arrays. Added to CHANGELOG.

test/wallet-test.js
rithvikvibhu: Added a multisig auction test in `wallet-rpc-test` from OPEN to CLOSED, including a failed no-blind reveal and importnonce that makes it work.

pinheadmz (commented):
New test is great! just nits left

test/wallet-rpc-test.js
```diff
@@ -952,4 +952,119 @@ describe('Wallet RPC Methods', function() {
...
+
+      // Fund Alice's wallet
+      await wclient.execute('selectwallet', ['msAlice']);
+      const addrAlice = await wclient.execute('getnewaddress', []);
+      await nclient.execute('generatetoaddress', [50, addrAlice]);
+
+      // Fund Bob's wallet
+      await wclient.execute('selectwallet', ['msBob']);
+      const addrBob = await wclient.execute('getnewaddress', []);
+      await nclient.execute('generatetoaddress', [50, addrBob]);
```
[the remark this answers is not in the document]
rithvikvibhu: d'oh!

rithvikvibhu:
Addressed all comments and cleaned up commits.

PR #768, opened 2022-09-23
mempool: reject TXs that exceed consensus covenant block limits

pinheadmz:
Users have been able to broadcast huge batches with 1,000 OPENs in a single TX. That would make an entire block invalid, so there's no reason to relay such a TX. We could also decide to add a POLICY rule that restricts this even further (the recent wallet batching work limits to like 200)


nodech (approved):
Good find.

I agree, would be nice to have additional policy for mempool/mempoolentry to further limit them if need be.

PR #769, opened 2022-09-26
wallet: increase lookahead configuration to 2 ** 32.

nodech:
**NOTE This requires migrations so needs to land in 5.x**
If there are any breaking changes to layout we want in DB, this is probably good time.

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -305,6 +305,7 @@ class HTTP extends Server {
         mnemonic: mnemonic,
         accountKey: accountKey,
         watchOnly: valid.bool('watchOnly'),
+        lookahead: valid.u32('lookahead'),
         language: language
       });
 
--- a/lib/wallet/migrations.js
+++ b/lib/wallet/migrations.js
@@ -164,7 +164,7 @@ class MigrateChangeAddress extends AbstractMigration {
 
 /**
  * Migrate account for new lookahead entry.
- * Applies to WalletDB v0
+ * Applies to WalletDB v1
  */
 
 class MigrateAccountLookahead extends AbstractMigration {
@@ -203,9 +203,10 @@ class MigrateAccountLookahead extends AbstractMigration {
       parse: key => layout.W.decode(key)[0]
     });
 
-    for (const wid of wids) {
+    for (const wid of wids)
       await this.migrateWallet(b, wid);
-    }
+
+    this.db.writeVersion(b, 2);
   }
 
   async migrateWallet(b, wid) {
@@ -224,12 +225,14 @@ class MigrateAccountLookahead extends AbstractMigration {
   }
 
   accountEncode(raw) {
+    // flags, type, m, n, receiveDepth, changeDepth
     const preLen = 1 + 1 + 1 + 1 + 4 + 4;
     const pre = raw.slice(0, preLen);
     const lookahead = raw.slice(preLen, preLen + 1);
     const post = raw.slice(preLen + 1);
     const newLookahead = Buffer.alloc(4, 0x00);
-    newLookahead[0] = lookahead[0];
+
+    newLookahead.writeUInt32LE(lookahead[0], 0);
 
     return Buffer.concat([
       pre,
@@ -241,7 +244,7 @@ class MigrateAccountLookahead extends AbstractMigration {
   static info() {
     return {
       name: 'Account lookahead migration',
-      description: 'Account lookahead now supports up to 2 ** 32'
+      description: 'Account lookahead now supports up to 2^32 - 1'
     };
   }
 }
--- a/lib/wallet/migrations.js
+++ b/lib/wallet/migrations.js
@@ -163,6 +163,90 @@ class MigrateChangeAddress extends AbstractMigration {
 }
 
 /**
+ * Migrate account for new lookahead entry.
+ * Applies to WalletDB v0
+ */
+
+class MigrateAccountLookahead extends AbstractMigration {
+  /**
+   * Create migration object.
+   * @param {WalletMigratorOptions}
+   */
+
+  constructor (options) {
+    super(options);
+
+    this.options = options;
+    this.logger = options.logger.context('account-lookahead-migration');
+    this.db = options.db;
+    this.ldb = options.ldb;
+  }
+
+  /**
+   * We always migrate account.
+   */
+
+  async check() {
+    return types.MIGRATE;
+  }
+
+  /**
+   * Actual migration
+   * @param {Batch} b
+   * @returns {Promise}
+   */
+
+  async migrate(b) {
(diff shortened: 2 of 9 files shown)


pinheadmz:
Concept ACK, didn't test code yet but looks like the right approach with migration. Question: should we add an RPC to bump up lookahead on existing accounts? Looks like the only way to set a non-default lookahead with this branch is when you create something

nodech:
yeah, I believe it only matters when creating account/wallet. Modifying existing accounts lookahead vs adding new account both require rescan to have a proper state, so I would say it will be preferred to just create new wallet/account or recreate account is better in general for that reason. It is easier to understand that adding new account needs rescan.

lib/wallet/migrations.js
```diff
@@ -162,6 +162,90 @@ class MigrateChangeAddress extends AbstractMigration {
...
+  async migrate(b) {
+    const wids = await this.ldb.keys({
+      gte: layout.W.min(),
+      lte: layout.W.max(),
+      parse: key => layout.W.decode(key)[0]
+    });
+
+    for (const wid of wids) {
+      await this.migrateWallet(b, wid);
+    }
```
pinheadmz, line 208: Do we also need to bump the walletDB version? A migrated wallet run in an old version of hsd will crash with a scary error instead of a version mismatch:

```
EncodingError: Out of bounds read (offset=89).
    at BufferReader.readU32BE (/Users/matthewzipkin/Desktop/work/hsd/node_modules/bufio/lib/reader.js:256:10)
    at readKey (/Users/matthewzipkin/Desktop/work/hsd/lib/wallet/account.js:857:30)
    at Account.read (/Users/matthewzipkin/Desktop/work/hsd/lib/wallet/account.js:781:19)
...
```

lib/wallet/migrations.js
```diff
@@ -162,6 +162,90 @@ class MigrateChangeAddress extends AbstractMigration {
   }
 }
 
+/**
+ * Migrate account for new lookahead entry.
+ * Applies to WalletDB v0
```
pinheadmz, line 167: we are already on WalletDB v1, this should migrate us to v2:

https://github.com/handshake-org/hsd/blob/309df94ad5b2874944a32de63b449ac0a542068d/lib/wallet/walletdb.js#L65

test/wallet-migration-test.js
```diff
@@ -540,6 +571,126 @@ describe('Wallet Migrations', function() {
...
+      ldb = walletDB.db;
+    });
+
+    afterEach(async () => {
+      if (ldb.opened)
+        await ldb.close();
+    });
+
+    const newToOld = (raw) => {
+      const preLen = 1 + 1 + 1 + 1 + 4 + 4;
```
pinheadmz, line 608: I'd include a comment here:

```js
// flags, type, m, n, receiveDepth, changeDepth
```

test/wallet-migration-test.js
```diff
@@ -540,6 +571,126 @@ describe('Wallet Migrations', function() {
...
+      };
+
+      walletDB.options.walletMigrate = 0;
+
+      await walletDB.open();
+      const wallet = walletDB.primary;
+      const wallet2 = await walletDB.get(1);
+      await checkLookahead(wallet, TEST_LOOKAHEAD + 0);
+      await checkLookahead(wallet2, TEST_LOOKAHEAD + 10);
+      await walletDB.close();
```
pinheadmz, line 691: now that migration is complete, we could add one more test setting lookahead `> 255`, close and re-open DB and make sure the new value is there and not rolled over a U8

pinheadmz (commented):
Great job, I have a few suggestions some are just optional style nits except I do think walletDB version does need to be incremented and the messages with new max value should probably include the `-1`. I prefer `^` to `**` but thats up to you.

pinheadmz:
I also rebased your branch on master and will f-push

test/wallet-migration-test.js
nodech: This makes sure migration is correct, that test is wallet-unit-test. (Also because of the new format that test has to manually convert new to old)

lib/wallet/http.js
```diff
@@ -305,6 +305,7 @@ class HTTP extends Server {
         mnemonic: mnemonic,
         accountKey: accountKey,
         watchOnly: valid.bool('watchOnly'),
+        lookahead: valid.u32('lookahead'),
```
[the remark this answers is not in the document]
nodech: Yes, I was waiting for this to merge first. But I will create PR rn.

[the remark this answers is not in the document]
nodech: https://github.com/handshake-org/hs-client/pull/52

pinheadmz (approved):
ACK fa9cc462bf9a5cd67fe9de8c1eede78159bd3163

```
Hash: SHA256

ACK fa9cc462bf9a5cd67fe9de8c1eede78159bd3163
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmO4hO0ACgkQ5+KYS2KJ
yTq6chAAuwlJpwk3dKNhm0GiaKqJVuzIu7Sc3ouUgWwMcCB0MByUWCPc7QfYirKT
jbePCeEL4D6HwZ6UHrh+APa0Y5lGda+RsJBgF8QHbst3BpdCk7aufvzD9jQrZ2BV
7dMkCofAm7MC9C2yH/3VQQzENM8iQjxacjmVB+ctJQpapaLVU2YSgEullN2qNjMp
hgWpE14/GS1dqzqeOKmWBzUIo1lioxgPFwbtKk7r087e6CUnINPwNs2LZqhl9NBz
gjYRaR9FjGsxBDbs8UD/Milvi5gEYA8uGlXwQsrITTDWjcgUAt5TeME+Rh7vjkVJ
76vP1UikJYyIho6eLmXf+KWH+Nlf8QoB83EAfHD43feqs1TNDX5lSlzMg/9ng+Hw
O6w8CoirEuZZsWR5VjiAwXkTnkYO7OAOShSQt6XSbeNVB/wO3ADuieQfZmHNo67G
61iQ4d1aucUmMAgFGLp8vqfmPaobLNl3k+Ydt2shZsK1Blt0oHTHHNE3WCkwxbT0
px6Pq1P4xbkjhRlW5ppoU3yvI7XQ7KyVgD0cFBfsmLQYQEzc4pXO696MbX5B3efT
yntkFFIIvtaFd15VhNyd1BOtY7nHholcBOHFF8/2T0Nu73bWdtEamb9MyfO42iiC
08radjyaqyZcKilWHWP5EJnUsylgyM/dGgcYKMt/g11Wk++M6nY=
=lZ1L
-----END PGP SIGNATURE-----
```

</details>

PR #771, opened 2022-10-05
Chain reset/reorg tests for tree/txn state

nodech:
Add chain reset/reorg tests with tree/txn state.


test/util/chain.js
```diff
@@ -0,0 +1,129 @@
...
+const assert = require('bsert');
+const Network = require('../../lib/protocol/network');
+const Chain = require('../../lib/blockchain/chain');
+const Miner = require('../../lib/mining/miner');
+const WorkerPool = require('../../lib/workers/workerpool');
+const rules = require('../../lib/covenants/rules');
+const store = require('../../lib/blockstore');
+const {testdir} = require('./common');
+
+const chainUtils = exports;
```
pinheadmz, line 12: This is cool. We don't have to address in this PR but I can see this module being classified and used throughout the testing suite.

PR #773, opened 2022-10-07
Mempool: handle covenant conflicts during reorg and save TXs for later if possible

pinheadmz:
This PR should improve UX and increase miner fee profits after reorgs and in some edge cases involving covenant conflicts.

There are some covenants that can not coexist in a block, for example `UPDATE` can only happen once per name per block. That also means that at any time in the mempool, there can only be one `UPDATE` per name. After a reorginzation when transactions get removed from blocks and re-enter the mempool, we need to track these conflicts carefully and make sure that "the next valid transaction" in the covenant chain is the only covenant for a given name in the mempool.

By storing the conflicting TXs outside the mempool in a new data structure, a miner can still earn fees LATER when the parent TXs confirm and the child TXs are once again valid.

### DIFF
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -98,6 +98,14 @@ class Mempool extends EventEmitter {
     // Recently rejected txs by hash
     this.rejects = new RollingFilter(120000, 0.000001);
 
+    // TXs removed from the mempool because their
+    // parent TX was removed from a block and their
+    // covenants prevent them from joining in a block or mempool.
+    // Will be valid again once the parent is re-confirmed.
+    this.disconnected =
+                    new BufferMap(); // parent hash -> BufferSet[child hashes]
+    this.children = new BufferMap(); // child tx hash -> TX (not MempoolEntry)
+
     // Extensions of blockchain indexes by tx hash for API
     this.coinIndex = new CoinIndex();
     this.txIndex = new TXIndex();
@@ -211,17 +219,26 @@ class Mempool extends EventEmitter {
       const entry = this.getEntry(hash);
 
       if (!entry) {
+        // Confirmed TX was not in mempool
+        // maybe clear other references and conflicts
         this.removeOrphan(hash);
         this.removeDoubleSpends(tx);
+        this.removeDoubleOpens(tx);
+
+        // Confirmed TX resolves an orphan
         if (this.waiting.has(hash))
           await this.handleOrphans(tx);
-        this.removeDoubleOpens(tx);
+
         continue;
       }
 
       this.removeEntry(entry);
 
       this.emit('confirmed', tx, block);
+
+      // Confirmed TX reconnects a child
+      if (this.disconnected.has(hash))
+        await this.reconnectSpenders(tx);
 
       entries.push(entry);
     }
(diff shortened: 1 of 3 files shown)


lib/mempool/mempool.js
```diff
@@ -78,20 +78,37 @@ class Mempool extends EventEmitter {
...
+
+    // Recently rejected txs by hash
     this.rejects = new RollingFilter(120000, 0.000001);
 
+    // TXs removed from the mempool because their
+    // parent TX was removed from a block and their
+    // covenants prevent them from joining in a block or mempool.
+    // Will be valid again once the parent is re-confirmed.
+    this.disconnected =
+                    new BufferMap(); // parent hash -> BufferSet[child hashes]
```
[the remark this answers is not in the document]
nodech: I feel like children should be called else: `disconnectedTXMap` or something.

lib/mempool/mempool.js
```diff
@@ -1851,6 +1885,75 @@ class Mempool extends EventEmitter {
...
+    for (const hash of hashes) {
+      const tx = this.children.get(hash);
+      assert(tx);
+
+      try {
+        const missing = await this.insertTX(tx, -1);
+        if (missing)
+          throw new Error('spender is orphan.');
+        // Recurse only on success
+        await this.reconnectSpenders(tx);
```
nodech, line 1948: I don't think we should reconnect spenders here again.. For example, if we have `update` -> `update` -> `update` chain in there, we only need to move into a single layer only, not in depth. Otherwise we will encounter unique name issues?

lib/mempool/mempool.js
```diff
@@ -1851,6 +1885,75 @@ class Mempool extends EventEmitter {
...
+      } catch (err) {
+        this.logger.debug(
+          'Could not reconnect spender %x: %s.',
+          tx.hash(), err.message);
+      }
+    }
```
nodech, line 1954: I think we should remove parent->[child hashes] from the disconnected list and the single layer of children from the children.

lib/mempool/mempool.js
```diff
@@ -209,18 +219,27 @@ class Mempool extends EventEmitter {
       const entry = this.getEntry(hash);
 
       if (!entry) {
+        // Confirmed TX was not in mempool
+        // maybe clear other references and conflicts
         this.removeOrphan(hash);
         this.removeDoubleSpends(tx);
```
nodech, line 225: Related to `removeDoubleSpends`:

There needs to be additional check when the transaction that we have not seen is removed by `removeDoubleSpends`. I think `evictEntry` is a good place to clean up `disconnects`, because evicting the original transaction that we are waiting to reconnect will leave them there w/o any way to clean up.

Same as `removeDoubleSpends` it could happen on covenant invalidation, e.g. unconfirmed disconnector transaction, that we are waiting for got invalidated in the mempool (e.g. expire). This also leads to evictEntry.

I believe all evictEntry callers are included here.

nodech: Another thing that we need to track is double spends to the .children transactions. Any of the `children` can get double spent in the block, but since they are not part of the normal transaction map and are not tracked via `getSpent` we will need to track them separately.

nodech (commented):
I think we can also add new check to the `exists(hash)` and `has(hash)` methods, to just skip this tx if it's already in the disconnected list.

lib/mempool/mempool.js
```diff
@@ -1851,6 +1885,75 @@ class Mempool extends EventEmitter {
...
+
+  async reconnectSpenders(parent) {
+    const hashes = this.disconnected.get(parent.hash());
+
+    if (!hashes)
+      return;
+
+    for (const hash of hashes) {
+      const tx = this.children.get(hash);
+      assert(tx);
```
nodech, line 1941: Note: this assert will no longer hold true if we track double spends for children.

lib/mempool/mempool.js
```diff
@@ -193,13 +210,6 @@ class Mempool extends EventEmitter {
    */
 
   async _addBlock(block, txs, view) {
-    if (this.map.size === 0
```
nodech, line 196: Why remove this ?

lib/mempool/mempool.js
```diff
@@ -1851,6 +1885,75 @@ class Mempool extends EventEmitter {
...
+
+      // Child is not linked by name.
+      // Covenant rules don't prevent it from
+      // staying in the mempool.
+      if (!rules.hasNames(tx, names))
+        continue;
+
+      // This child has to be removed from the mempool
+      // but it will be valid again once its parent confirms.
+      this.disconnectSpenders(tx);
```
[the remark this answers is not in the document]
pinheadmz: good point about this, I need to go back down this rabbit hole. it's possible i was clearing out ANY children of the disconnected tx, which would apply to like, if a child spent the change output of a name-linked tx

lib/mempool/mempool.js
pinheadmz: See https://github.com/handshake-org/hsd/pull/319 

I don't think there was any reason for this to be here besides early-bcoin-style laziness and maybe even an accidental copy-paste from `_removeBlock()`where it used to make more sense, to just avoid complicated mempool issues.

What the function does is resolve orphans whose missing parents were just confirmed. There is no reason to NOT do that just because the current mempool is empty.

PR #776, opened 2022-11-02
Fix npm warning

NetOpWibby:
Apparently, `--omit=dev` should be used over `--production`.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -33,7 +33,7 @@ keys. You can get keys following [Security](../SECURITY.md) document.
 ```
 git clone --depth 1 --branch latest https://github.com/handshake-org/hsd.git
 cd hsd
-npm install --production
+npm install --omit-dev
 # run full node in foreground with default configuration
 ./bin/hsd
 ```
--- a/docs/install.md
+++ b/docs/install.md
@@ -33,7 +33,7 @@ keys. You can get keys following [Security](../SECURITY.md) document.
 ```
 git clone --depth 1 --branch latest https://github.com/handshake-org/hsd.git
 cd hsd
-npm install --omit-dev
+npm install --omit=dev
 # run full node in foreground with default configuration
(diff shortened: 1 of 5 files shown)


docs/install.md
```diff
@@ -33,7 +33,7 @@ keys. You can get keys following [Security](../SECURITY.md) document.
 ```
 git clone --depth 1 --branch latest https://github.com/handshake-org/hsd.git
 cd hsd
-npm install --production
+npm install --omit-dev
```
rtdev-com, line 36:
```suggestion
npm install --omit=dev
```

rtdev-com (changes requested):
Need to fix a typo

docs/install.md
NetOpWibby: Oh man, nice catch!

rtdev-com (approved):

pinheadmz:
grep the repo for `production`, there are two more places to change I think: `Dockerfile` and `CONTRIBUTING.md` can you find anything else like this that needs to be updated?

NetOpWibby:
`grep -irn "\-\-production"` shows:

```
./Dockerfile:9:RUN npm install --production
./docs/install.md:36:npm install --production
./CONTRIBUTING.md:23:"developer dependency" if installed without the `--production` flag. The
```

NetOpWibby:
@pinheadmz All set!

pinheadmz:
CI was failing on node v12 which is unsupported now anyway. Removing that so this docs-only PR will pass integration tests!

pinheadmz:
Ugh, @chjj's DNS seed server is down. I'm gonna fix this flakey test by targeting `nic.com` a much more reliable name for lookups.

pinheadmz:
Ugh part III: Fine lets just skip the test that requires external DNS lookups, for now.

Issue #774, opened 2022-10-11
wallet batches may exceed policy weight limit


pinheadmz:
the reason is because as we are building the batch, we only account for ONE funding input: 

https://github.com/handshake-org/hsd/blob/master/lib/wallet/wallet.js#L3851-L3866

If the user's wallet needs to spend more than one coin to pay the fee for a batch, this size estimator will be wrong and the final TX will break the rule.

We could for example maybe call coin selector during the batch process and see how many inputs we'll need to fund the batch and take that size into account -- but i think that would mean calling `fund()` too many times during batching.

nglabs42:
I was able to send 10 HNS to myself and then send the sendbatch '[["REVEAL"]]' was able to sent within the next block.

Issue #770, opened 2022-09-28
Coin selector MAX_FEE is causing issues


pinheadmz:
Currently the wallet will hard cap all tx fees at `0.1 HNS`, which is totally arbitrary and inherited from `0.1 BTC` when that made sense as a hard limit. The problem is now with batching, users are creating very large TXs and with a hard-capped `0.1 HNS` fee, the effective rate is too low to relay.

PR #777, opened 2022-11-11
wallet: Set hard-coded max fee to 10.000000 HNS

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/770

TODO:

- [ ] Also fix https://github.com/handshake-org/hsd/issues/774


pinheadmz:
closing this. on second thought, i think the max fee just needs to be completely removed...

PR #778, opened 2022-11-13
Remove Coinselector.MAX_FEE

pinheadmz:
Closes https://github.com/handshake-org/hsd/issues/770

Replaced with new policy checks in wallet for fees too high aka "absurd fee" (> 10000 * minFee) and too low (< minFee).


lib/wallet/wallet.js
```diff
@@ -3956,6 +3954,16 @@ class Wallet extends EventEmitter {
...
       throw new Error('TX exceeds policy weight.');
 
+    const minFee = policy.getMinFee(
+      mtx.getVirtualSize(),
+      this.network.minRelay
+    );
+    const fee = mtx.getFee();
+    if (fee < minFee)
+      throw new Error('Fee is below minimum relay limit.');
+    if (fee > minFee * 10000)
```
nodech, line 3964: I think we need to make absurd fees little bit more configurable and move it to `policy`. It should default to this still, but allow easier future modifications.

nodech: nit: maybe add more spacing around ifs.

test/wallet-coinselection-test.js
```diff
@@ -0,0 +1,160 @@
...
+const assert = require('bsert');
+const {
+  MTX,
+  Network,
+  WalletDB
+} = require('..');
+
+// Use main instead of regtest because (deprecated)
+// CoinSelector.MAX_FEE was network agnostic
+const network = Network.get('main');
```
nodech, line 12: Configurable policy could help here as well.

pinheadmz: I think I'm going to keep this on main. Configurable absurd factor doesnt matter between networks, but we have default fee rate different for each network. 

What I think is important is this proof that the highest possible fee at default rate is 10 HNS:

https://github.com/pinheadmz/hsd/blob/48d93c452020562807b550119f96425b2eafa19c/test/wallet-coinselection-test.js#L91-L99

lib/wallet/wallet.js
pinheadmz: thanks, added in fa3729d7ae00ca3381affb0118336a21ca8173a8

rithvikvibhu (approved):
LGTM. Tried sending transactions with a few different rates and even the cli errored properly with `< min relay` and `> absurd` rates with `hsw-cli send --rate`.

PR #779, opened 2022-11-14
hostlist: only return MANUAL addrs in getLocal() for all networks

pinheadmz:
Main goal here is to fix tests that fail locally if the platform has a public ipv6 address. That address may get returned by `getLocal()` even if no `public-host` was set.

### DIFF
--- a/lib/net/hostlist.js
+++ b/lib/net/hostlist.js
@@ -1033,11 +1033,9 @@ class HostList {
 
     if (!src) {
       for (const dest of this.local.values()) {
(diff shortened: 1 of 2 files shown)


pinheadmz:
This is the test that fails on master for me with ipv6 enabled:

https://github.com/handshake-org/hsd/blob/9cf8cb83812841628de79454a20a7c2560120633/test/node-http-test.js#L76

nodech (approved):

PR #780, opened 2022-11-16
txdb lock balance accounting fixes.

pinheadmz:
The assertions were only there in the first place because of weird covenant accounting: for some covenants like REVEAL, we dig up the corresponding BID from the database and we MUST HAVE THAT DATA or we can't compute the change to the wallet's locked balance.

The new approach is: we process all inputs and all outputs of every TX, meaning that the BID in the input of the transaction is applied to our balance mere moments before we process the REVEAL in the output of the same transaction. This simplifies balance locking really well (we eliminate DB lookups and the functions are no longer `async/await`).

The assertion errors we've been getting almost daily from users occurs when the BID is not in the database yet because we don't recognize our own address because it's a wallet recovery and we haven't generated the address yet.

What this PR does is "not care": if we see a REVEAL before we see the BID, then we have a *negative locked balance*. Yep that really sucks but it can be fixed with a rescan or two... and no one's node has to crash.

There is one other drawback to this, which is why I'm pausing here and opening as draft:

TODO:

- [ ] `class Balance` in txdb needs to allow negative values. Remove `assert >= 0` and switch all `writeU64` to `writeI64` which interestingly, because it's javascript, I think means we don't need a migration ... ?!

### DIFF
--- a/lib/wallet/account.js
+++ b/lib/wallet/account.js
@@ -309,26 +309,28 @@ class Account extends bio.Struct {
 
   /**
    * Create a new receiving address (increments receiveDepth).
-   * @returns {WalletKey}
+   * @param {Batch} b
+   * @returns {Promise<WalletKey>}
    */
 
-  createReceive() {
-    return this.createKey(0);
+  createReceive(b) {
+    return this.createKey(b, 0);
   }
 
   /**
    * Create a new change address (increments changeDepth).
-   * @returns {WalletKey}
+   * @param {Batch} b
+   * @returns {Promise<WalletKey>}
    */
 
-  createChange() {
-    return this.createKey(1);
+  createChange(b) {
+    return this.createKey(b, 1);
   }
 
   /**
    * Create a new address (increments depth).
    * @param {Boolean} change
-   * @returns {Promise} - Returns {@link WalletKey}.
+   * @returns {Promise<WalletKey>}
    */
 
   async createKey(b, branch) {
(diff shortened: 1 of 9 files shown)


lib/wallet/layout.js
```diff
@@ -72,6 +72,12 @@ exports.wdb = {
  *   H[account][height][hash] -> dummy (tx by height + account)
  *   C[account][hash][index] -> dummy (coin by account)
  *   b[height] -> block record
+ *   A[nameHash] -> Name State
+ *   U[hash] -> Nme Undo
```
pinheadmz, line 76: typo

lib/wallet/account.js
```diff
@@ -511,7 +513,7 @@ class Account extends bio.Struct {
 
   /**
    * Allocate new lookahead addresses if necessary.
-   * @param {Number} receiveDepth
+   * @param {Number} recawait eiveDepth
```
pinheadmz, line 516: what? lol

lib/wallet/txdb.js
```diff
@@ -3003,24 +3036,26 @@ class TXDB {
...
     const coin = await this.getSpentCoin(spent, prevout);
 
     if (!coin)
-      return;
+      return false;
 
     coin.height = height;
 
     b.put(layout.d.encode(spent.hash, spent.index), coin.encode());
+    return true;
```
pinheadmz, line 3058: are these return values used anywhere?

test/wallet-locked-balance-test.js
```diff
@@ -19,17 +19,20 @@ const {
...
-const WALLET_N = 10;
+const WALLET_N = 5;
 const GRIND_NAME_LEN = 10;
 const hardFee = 1e4;
 
 // TODO: Add insert w/ block only tests(no unconfirmed step).
 // TODO: Add claim test.
 // TODO: (Re)Move and merge wallet-test lock balance checks.
+// TODO: Add double spend and double spend/erase of chained txs.
+//        - Test when we have spent tracking(layout.s) but not spendCoin(layoud.d)
```
pinheadmz, line 40: typo: "layout"

lib/wallet/txdb.js
```diff
@@ -24,6 +24,10 @@ const NameUndo = require('../covenants/undo');
 const {TXRecord} = records;
 const {types} = rules;
 
+/**
+ * @typedef {import('./walletdb')} WalletDB
+ */
+
```
pinheadmz, line 30: CI didn't like this.

```

ERROR: Unable to parse a tag's type expression for source file /home/runner/work/hsd/hsd/lib/wallet/txdb.js in line 27 with tag title "typedef" and text "{import('./walletdb')} WalletDB": Invalid type expression "import('./walletdb')": Expected "!", "=", "?", "[]", "|", or end of input but "(" found.
Error: Process completed with exit code 1.
```

PR #782, opened 2022-12-16
Txdb tests and updates

nodech:
This PR attempts to cover balance tracking with tests and fix some edge cases.
This is based on the initial idea of #780 and tests from it.

But expands scope to cover all balances instead of lock/unlocked.
To do full review on the txdb balance states and cover all missing cases, I have created psuedo code like document that includes all the cases this PR covers and future updates that may happen to the txdb methods. All cases that are covered in this PR are marked as `checkbox`es in the document: https://gist.github.com/nodech/c943e7c50b300b0505cc9dfcae243d48

Suggestions to where to move this document or if it's useful to other developers are welcome here or on the document.

Next tasks (Not part of the PR), maybe create issues:
 - Add 'missed credit' events

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## Unreleased
 
-### Wallet API:
+**When upgrading to this version of hsd, you must pass `--wallet-migrate=3` when
+you run it for the first time.**
 
+### Wallet Changes:
+
+- Add migration that recalculates txdb balances to fix any inconsistencies.
 - HTTP Changes:
   - All transaction creating endpoints now accept `hardFee` for specifying the
     exact fee.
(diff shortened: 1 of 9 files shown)


nodech:
Updated the document with the stuff that has been verified/implemented: https://gist.github.com/nodech/c943e7c50b300b0505cc9dfcae243d48

nodech:
Rebased on master to fix bdb segfaults.

nodech:
> P.S. Every attempt at reading test/wallet-balance-test.js, I get lost in some place or another 😄 must've been a pain to write and test.

Yeah, I get lost when I get back to it for couple minutes as well. But otherwise it would have been huge file. Maybe I will try to simplify at some point. But for now it is what it is.

PR #788, opened 2023-01-05
Mempool reorg covenants

pinheadmz:
Minimal alternative to https://github.com/handshake-org/hsd/pull/773

fixes the bugs without adding additional complexity.

### DIFF
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -78,20 +78,29 @@ class Mempool extends EventEmitter {
     this.tip = this.network.genesis.hash;
     this.nextState = this.chain.state;
 
-    this.waiting = new BufferMap();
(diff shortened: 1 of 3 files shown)


lib/mempool/mempool.js
```diff
@@ -193,13 +202,6 @@ class Mempool extends EventEmitter {
    */
 
   async _addBlock(block, txs, view) {
-    if (this.map.size === 0
```
nodech, line 196: Why was this removed? (same as in another PR)

pinheadmz: https://github.com/handshake-org/hsd/pull/773#discussion_r1062779385

nodech (approved):

PR #796, opened 2023-01-16
Import `hs-client` into the project.

nodech:
`hsd` and `hs-client` are tightly coupled. When major or minor changes occur [hs-client][hs-client] needs to reflect them right away. The versioning scheme should be exactly the same as well, which has not been the case up til now (we used patches there). If versions don't match, users will have trouble figuring out which `hs-client` version to use with which `hsd`.

The most troublesome part is the major breaking changes, if something breaks hs-client it becomes troublesome to manage `hs-client`. Having client alongside with `hsd` make this really simple.

Standalone version of `hs-client` is also necessary, for those who want minimal functionality w/o all heavy dependencies in their projects. So this PR introduces script in `scripts/gen-hsclient.js` which can aid with pushing and publishing of the `hs-client` both to [github][hs-client] and npm. 

To sum it up:
 - Pros
   - Easier to work with, especially for breaking API changes.
   - Easier and better management version of the hs-client.
 - Cons
   - Contributors need to be aware of the hs-client files and not introduce dependencies that make standalone 'hs-client' heavy (see scripts/gen-hsclient.js).
   - Extra step in the release process (which should have already been happening if we want better versioning of hs-client)

You can review usage update here: https://github.com/nodech/hsd/blob/6bd7cc0a825facf9dec7894888950939d7b427df/docs/release-files.md#hs-client

[hs-client]: https://github.com/handshake-org/hs-client


rithvikvibhu:
I think this makes a lot of sense and each api change won't need to track a sibling PR on another repo. I've tried this branch and the generated hs-client and other than the slightly awkward process to dev hs-client (discussed in https://t.me/hns_tech/74474) everything looks good to me.

Q: What do you think about the exposed `bin`s of both hsd and hs-client now? Currently, hsd has `hs*-cli` and hs-client has `hs*-cli, hs*-rpc`. I get that hs-client will still be a separate package, but maybe hsd can add `hs*-rpc` as well? So users won't need to install hs-client for the rpc shortcuts. I bring this up because hsd and hs-client both can't be installed globally without `--force` overwriting the two conflicting bin files.

Non-blocking suggestions:
- `HS_CLIENT_DIR` but as an arg to the script
- an option to the script arg to only generate content, without any git work

scripts/gen-hsclient.js
```diff
@@ -0,0 +1,248 @@
...
+  ],
+  main: './lib/client/index.js',
+  bin: {
+    'hsd-cli': './bin/hsd-cli',
+    'hsd-rpc': './bin/hsd-rpc',
+    'hsw-cli': './bin/hsw-cli',
+    'hsw-rpc': './bin/hsw-rpc'
+  },
+  engines: {
+    node: '>=8.0.0'
```
rithvikvibhu, line 37: does this need to be documented somewhere in a release process or something to keep track and update?
I see hsd's package says `>=14.0.0` and this one is lower probably because it's just a client. Anyway, just more of an fyi

scripts/gen-hsclient.js
```diff
@@ -0,0 +1,248 @@
...
+
+const DEPENDENCIES = [
+  'bcfg',
+  'bcurl',
+  'bsert'
+];
+
+const COPY_FILES = [
+  '.npmignore',
+  '.gitignore',
```
rithvikvibhu, line 58: hsd's gitignore is a superset of hs-client's (current) gitignore so not a problem, but you think it needs a separate hardcoded def here instead of copying hsd?

rithvikvibhu (approved):
This is awesome, can finally maintain both packages together. The new script args work great. Left some non-blocking comments.

docs/release-files.md
```diff
@@ -119,10 +122,38 @@ and create PR with the relevant updates to the `download/index.html` and
...
+  Since hsd v5 `hs-client` is part of the `hsd`. Original [hs-client repo][hsclient] is now used to
+publish generated content. `hs-client` version will now be strictly tied to
+the `hsd` version. It is then generated from `hsd` code to release separately on
+`git` and `npm`. Most of the process is done by the introduced helper script
+`scripts/gen-hsclient.js`. It can help you setup `hs-client` that just needs
+publishing on `git` and `npm`. It also gives instructions how to do both.
+  After `hsd` has been released we can also release `hs-client` from the same
+commit/tag, just run: `./scripts/gen-hsclient.js` which will generate `hs-client`
+package with `git` setup in `tmp` directory. You can alternatively pass
+`HS_CLIENT_DIR` env variable for custom place. If generating git failed for some
```
rithvikvibhu, line 135: I guess HS_CLIENT_DIR won't work now, has to be an argument

scripts/gen-hsclient.js
nodech: It's good enough for the purposes. It's not meant to be developed separately. Even if we add `symlink` later, superset is enough.

docs/release-files.md
nodech: True, forgot to update this doc. Good catch

PR #797, opened 2023-01-17
Wallet http endpoint for updating account lookahead

nodech:
This is based on #796, so if it does not make it I can rebase on current master, but #796 makes it much easier.

Summary:
  - Add modify account endpoint
  - Change create account endpoint
  - Remove wallet check-lookahead flag.

Note, this update changes existing API call for `createAccount`, which used to be `PUT /wallet/:id/account/:account`. Until now it was okay, even though it's not RESTish. But this PR adds ability to modify account, so these need to change:
  - Create account becomes: `POST /wallet/:id/account/:account`
  - `PUT /wallet/:id/account/:account` becomes MODIFY/UPDATE the existing account entry.

Currently, only thing that can be changed about account is `lookahead`.

This also deprecates `check-lookahead` which was used previously to max out lookahead for all the wallets. Now this can be done via this endpoint and new accounts already use 200 by default.

To make review easier, you can check the commits since [`2dc2c48`](https://github.com/handshake-org/hsd/pull/796/commits/2dc2c488d95adae93ed379cc37fcf5f4e6df5f51): https://github.com/handshake-org/hsd/compare/2dc2c48...1c5fd1a

NOTE: Users increasing lookahead MAY want to rescan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,17 +6,21 @@
   Validation and request paremeter errors will no longer return (and log) `500`
 status code, instead will return `400`.
 
-### Wallet configuration
-`hsd.conf` can now be used to define wallet options, when wallet is running as a plugin.
-Configurations with `wallet-` prefix will be passed to the wallet. `hsd.conf` wont be used
-if the wallet is running in standalone mode.
+### Wallet Changes
+#### Configuration
+  `hsd.conf` can now be used to define wallet options, when wallet is running as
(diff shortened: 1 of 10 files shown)


rithvikvibhu (approved):
This has been used by 2 Bob users; worked fine and fixed for one, and still unclear on the other (different problem now, maybe unrelated).

Either way, changing lookahead seems to work. ACK 1c5fd1ae54b72ff64903d0642dbbcd777b152be7

PR #802, opened 2023-01-20
wallet-http: add output paths to JSON

pinheadmz:
We already add derivation path information to inputs in tx json from http module (if `{paths: true}` is set) but we could also add those details to the outputs. For inputs we use the coinview but for outputs we can just add the data directly to JSON in the http module, no need to modify tx or mtx.

The use case (for example [palmreader](https://github.com/pinheadmz/palmreader/blob/master/plugin.js#L570-L574)) is for external signers to know the total wallet balance effect a TX will have before signing.


rithvikvibhu:
Bob could've used this a few months ago :smiling_face_with_tear:. Awesome it's being added to hsd!

https://github.com/kyokan/bob-wallet/pull/569/files#diff-adb4f30f2b6f3e461a271d2058bd5679fcd9de5040820f6c3d1c8f7bce05936cR1925-R1937

```js
// Can be removed when hsd starts adding path to output.getJSON()
```

nodech:
Do we want to skip paths if the transaction was signed and broadcast by the wallet ?

That information may not useful when user broadcasts the transaction (it's already gone), but for consistency I was wondering if we want to also return output paths when `broadcast` is true (default), because those will still return input paths.

Also, maybe leave a space before `if (options.paths)` to improve readability?

If we want those outs we could also move out to `async getTXJSON(tx, network, options)` instead as well.

test/wallet-http-test.js
```diff
@@ -188,6 +188,50 @@ describe('Wallet HTTP', function() {
       assert.ok(typeof path.change === 'boolean');
       assert.ok(typeof path.derivation === 'string');
     }
+
+    // cbAddress is a self-send
+    // so all output paths including change should be known
+    for (let i = 0; i < tx.outputs.length; i++) {
+      const path = tx.outputs[i].path;
+
+      assert.ok(typeof path.name === 'string');
```
rithvikvibhu, line 197: Any reason this test uses `assert.ok`? I thought hsd prefers `assert.strictEqual`

rithvikvibhu (commented):
>also return output paths when broadcast is true (default), because those will still return input paths.

yeah that makes sense and since `path` is false by default, it won't increase response size unless needed.

test/wallet-http-test.js
nodech: For bools, it could go either `assert(..)` or `assert.ok(`. `assert.ok` preferred in tests.

PR #804, opened 2023-02-24
Use `nurkel` for the tree instead of `urkel`.

nodech:
[`nurkel`](https://github.com/nodech/nurkel) is a node bindings for the [`liburkel`](https://github.com/chjj/liburkel). 

It has not been tested in action so far, but I have run some tests with it, so if you are willing to experiment/test/run you are welcome.

There's also liburkel-urkel-nurkel integration suite here: https://github.com/nodech/liburkel-test, unfortunately not well documented. 

The benchmarks are pretty promising for the actual sync.

First the most important - How much time difference does it make while syncing the node. These are the sync times until the block height (NOTE: Data is a single run):
total

Graph below shows PER BLOCK TIME differences between the two, both are recorded using single run so they should be less "bumpy" on multiple runs:

perblock

This will be DRAFT until it receives some feedback and has been tested out in the wild. I will try to rebase it on master as often as possible.


handshake-enthusiast:
Nice! Does https://github.com/nodech/liburkel-test generates the images above?

I could at least test on Mac with SPV.

nodech:
> Nice! Does https://github.com/nodech/liburkel-test generates the images above?
> 
> I could at least test on Mac with SPV.

No, liburkel-test generates random tree (and some specific edge cases) and compares that tree files match byte by byte.

Unfortunately, SPV does not have a tree so it wont test the tree.

As for the graphs, they were generated from the log files of the hsd. Specifically `[I:2023-02-23T15:33:07Z] (chain) Block 0000000000000004226d2a1178bd745c43dc7e4ff286e45e6a21194db04e4050 (144964) added to chain (size=286599 txs=133 time=307.788625).`

These were then parsed accumulated in separate file for gnuplot: https://gist.github.com/nodech/da78e501b29cd05afc620648a721d97c

rithvikvibhu:
This is very promising and I had no trouble getting Bob to sync with nurkel, awesome work @nodech :D Even switching between urkel and nurkel with the same tree data worked fine.

Some stats from my machine up to block 150k (with checkpoint at 100k):
image

Bob saw a 2x speed sync that reduced time from 20 hours to 10 hours :exploding_head: 

### Some observations
- The orange line is probably an outlier / due to network effects
- We're not sure what that small steep slope in the middle is
- Indexers being enabled or not didn't make any noticeable difference

As for compacting 161k blocks (approx with ± 200 blocks diff):
|              | time    |
|--------------|---------|
| urkel (Bob)  | 10m 34s |
| nurkel (Bob) | 6m 29s  |

I'd like to run these tests again in a proper environment and syncing from a local node to rule out network weirdness and on a dedicated machine. But ^ gives a rough idea on what we can expect.

handshake-enthusiast:
Great results!

>Unfortunately, SPV does not have a tree so it wont test the tree.

As far as I get https://github.com/nodech/liburkel-test wouldn't work in SPV mode. Can I still run `hsd` in SPV mode and compare sync times using https://gist.github.com/nodech/da78e501b29cd05afc620648a721d97c?

nodech:
> As far as I get https://github.com/nodech/liburkel-test wouldn't work in SPV mode.

https://github.com/nodech/liburkel-test - is not related to HSD at all. It is specific to the urkel/nurkel/liburkel integration test.

 > Can I still run `hsd` in SPV mode and compare sync times using https://gist.github.com/nodech/da78e501b29cd05afc620648a721d97c?

SPV mode in hsd does not have tree (as having 24 GB tree would not make sense with SPV), so whether SPV node has this update or not, there will be no difference.

rithvikvibhu:
Found a bug:

```sh
❯ hsd-cli rpc getnameproof theshake
proof.toJSON is not a function
```

Think nurkel's `Proof` doesn't have `toJSON`

PR #806, opened 2023-03-07
allow plugins to inherit configurations from the hsd.conf

nodech:
After https://github.com/bcoin-org/bcfg/pull/8 (https://github.com/bcoin-org/bcfg/pull/4) we can allow plugins to inherit options from files as well.

Wallet plugin(normal hsd with wallet in it) will inherit configuration from the `hsd.conf` with `wallet-` prefix.

e.g. `wallet-http-port: 12049` will be passed down to the wallet as `http-port`.

### Downsides
Standalone wallet node wont use `hsd.conf` at all. And it should NOT. `hsw` does not need `--wallet` prefix for anything either and will only read `hsw.conf`. So if anyone is jumping from one to another may get confused why `hsd.conf` is not loaded.

Even though https://github.com/bcoin-org/bcfg/pull/8 allows other plugins to decide for themselves what they want to inherit and how, I am not 100% sure if we want wallet to inherit the options given the downside.


rithvikvibhu (approved):
This is nice, was able to set wallet settings from hsd.conf, hsw.conf, cmd args to hsd and preference stayed in same order (latter getting priority over the ones before)

PR #807, opened 2023-03-09
http: update validator status codes.

nodech:
Node and wallet validation failures using `bval` were returning `500` status code and logging them as errors in debug log. Now all bval errors in http will return `400` status code instead and no longer log in the debug. Also updated usage of `assert`s that would also result with `500` error in further parameter validations, so they now return `400`.

You can check `bval` fix here: https://github.com/bcoin-org/bval/pull/8


rithvikvibhu (approved):
These json errors are also much neater! And confirmed it doesn't spam hsd log

Before:
```
500 - {
  "error": {
    "type": "AssertionError",
    "code": "ERR_ASSERTION",
    "message": "Must pass valid name."
  }
}
```

Now:
```
400 - {
  "error": {
    "message": "Must pass valid name."
  }
}
```

PR #813, opened 2023-05-24
Mempool invalidate claims when period ends.

nodech:
Claims if they get ignored in the `claimPeriod - 1` block and still are part the mempool - next block will be invalid as it will include claims (as they are in mempool).

This invalidates claims in the mempool.

### DIFF
--- a/lib/mempool/mempool.js
+++ b/lib/mempool/mempool.js
@@ -295,7 +295,7 @@ class Mempool extends EventEmitter {
     // when the block one before the height that disables
     // GooSig is added to the mempool to prevent the
     // mining of invalid blocks.
-    if (block.height + 1 === this.network.goosigStop) {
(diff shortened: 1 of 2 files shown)


lib/mempool/mempool.js
```diff
@@ -284,6 +285,12 @@ class Mempool extends EventEmitter {
         this.untrackClaim(entry);
     }
 
+    // At the chain.tip.height of claimPeriod - 1, we no longer
+    // accept new claims as they will be invalid in the next block.
+    // We also need to clean them up from the mempool.
+    if (nextHeight === this.network.names.claimPeriod)
```
rithvikvibhu, line 291: tiny nit, the next code block (goosigStop) directly uses `block.height + 1`, that can also use nextHeight right?

rithvikvibhu (approved):

PR #816, opened 2023-05-25
wallet: refactor createTX to be more flexible

nodech:
Remove fundlock force hack for the send.

It now will repeat the same signature as others like it. (open, bid, ..)


rithvikvibhu (approved):
:rocket:

PR #818, opened 2023-05-30
wallet-http: reveals all now can be created without broadcasting.

nodech:
Noticed on previous pr.

Now revealAll will be similar to redeemAll.


rithvikvibhu (approved):
Can create reveal all without broadcast now :+1:
I think the other change for readability? Doesn't affect anything.

PR #823, opened 2023-06-21
Don't reserve Alexa Top100 names, only reserve ICANN TLDs and "custom" list

paste:
### An update to existing SOFT FORK #819.

By looking at the full Alexa list of 100K names, it becomes clear there are many names which are just as popular or arguably more deserving than our _outdated_ [list of top 100](https://gist.github.com/nodech/a5a308e2b01456d1ea0e34f4be4bd609#file-top100-json). Because our list of top 100 is neither comprehensive nor maintained, it seems arbitrary to reserve them and not others. Why should `pinterest` be reserved but not `craigslist`? Additionally, two very generic terms `mail` and `ok` would potentially be reserved forever, requiring a hard fork to be released. Rather than trying to pick and choose which names should make the cut and relying on additional forks to re-enable claims, I think it's better to just let the claim period end as originally intended. We made a good faith effort to allow claims for 4 years, now time to move on.

This PR removes ALL Alexa names from the reserved list, which means they will ALL become available to OPEN around Jan 1st, 2024. Only ICANN TLDs and the short [list of 27 "custom" names](https://gist.github.com/nodech/da108d035cb3c694144cc564b07659e4) chosen by Handshake founders will be permanently reserved.

**Any discussion or feedback on this change should be resolved before July 1st in order to wrap up the HSD release and begin distribution and signaling.**

**Custom list of 27 names to be reserved (chosen by Handshake founders):**
https://gist.github.com/nodech/da108d035cb3c694144cc564b07659e4

**Alexa Top 100 names (to be released) here:**
https://gist.github.com/nodech/a5a308e2b01456d1ea0e34f4be4bd609#file-top100-json

**Alexa Top 100k names (to be released) here:**
https://gist.github.com/paste/4228c764c139a447de30de62afb9a393

### DIFF
--- a/lib/blockchain/chain.js
+++ b/lib/blockchain/chain.js
@@ -705,7 +705,7 @@ class Chain extends AsyncEmitter {
     if (await this.isActive(prev, deployments.hardening))
       state.nameFlags |= rules.nameFlags.VERIFY_COVENANTS_HARDENED;
 
-    // Disable ICANN, TOP100 and CUSTOM TLDs from getting auctioned.
+    // Disable ICANN and CUSTOM TLDs from getting auctioned.
     if (await this.isActive(prev, deployments.icannlockup))
       state.nameFlags |= rules.nameFlags.VERIFY_COVENANTS_LOCKUP;
 
--- a/lib/covenants/rules.js
+++ b/lib/covenants/rules.js
@@ -421,7 +421,7 @@ rules.isLockedUp = function isLockedUp(nameHash, height, network) {
   if (!item)
     return false;
 
-  if (item.root || item.top100 || item.custom)
+  if (item.root || item.custom)
     return true;
 
   return false;
--- a/test/chain-icann-lockup-test.js
+++ b/test/chain-icann-lockup-test.js
@@ -44,7 +44,7 @@ const ACTUAL_RENEWAL_WINDOW = network.names.renewalWindow;
  * Test will run failure and success paths and make sure both give
  * results soft-fork expects:
  *  - on failure: names can be auctioned.
- *  - on success: root, top100, custom and zero become
+ *  - on success: root, custom, and zero become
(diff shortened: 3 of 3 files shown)


Falci:
Why don't we use another bit, keep both flags and let the miners decide?

pinheadmz:
> Why don't we use another bit, keep both flags and let the miners decide?

I don't think we have the free time for this added complexity and I don't think the miners are well-informed or community-active enough to make and impose this decision on consensus rules. BIP9 is not a voting system - it is a readiness-signalling system. We* decide the rules and the miners tell us on chain that they are running the new software.

nickorlabs:
TL;DR - While I have mixed feelings on this, I could easily get behind only including ICANN and Custom protection in the soft fork!


While I have mixed feeling on releasing all of these names, I completely agree with the statement "it becomes clear there are many names which are just as popular or arguably more deserving than our outdated" and think it really is valid for the entire Alexa Top 100,000.  With this we will continue to be backwards compatible with ICANN.  Additionally, the poll I published it seems to me the community's second choice would be to allow expected results for the Alexa Top 100,000, which I think could easily include the Top 100.
image
I am sure it would only take a few people voting there for expected results also to be the overall winner for the choices.  
I would easily get behind this if this is the path we go.  
I also have a tweet asking for the community to help reach out to the companies and communities that have not claimed their names to make a last ditch effort for them to claim before block 210240.  @rithvikvibhu's idea!  Regardless which way we go, some if not all names will be released so we should do anything and everything to prepare people!

nickorlabs:
Had a Twitter Space and discuss this and the poll, we currently have 10 votes and it seems the feeling is the names in the current top 100 are outdated and do not represent the current top 100 sites.

image

I have posted the link to the poll again today on Twitter; maybe we will get some additional votes before the end of the day.

PR #825, opened 2023-07-13
wallet: fix makeBatch to generate addresses early

rithvikvibhu:
Bid actions in `makeBatch` had their addresses overridden after nonce generation, making it impossible to recover the wallet with importnonce.

With this change, the address is derived and passed to make* so the correct address is used to generate nonces.

The test covers both `sendbid` and `sendbatch`.

### DIFF
--- a/lib/wallet/wallet.js
+++ b/lib/wallet/wallet.js
@@ -1760,14 +1760,16 @@ class Wallet extends EventEmitter {
    * @param {Number} lockup
    * @param {Number|String} acct
    * @param {MTX?} mtx
-   * @returns {MTX}
+   * @param {Address?} addr
+   * @returns {Promise<MTX>}
    */
 
-  async makeBid(name, value, lockup, acct, mtx) {
+  async makeBid(name, value, lockup, acct, mtx, addr) {
     assert(typeof name === 'string');
     assert(Number.isSafeInteger(value) && value >= 0);
     assert(Number.isSafeInteger(lockup) && lockup >= 0);
     assert((acct >>> 0) === acct || typeof acct === 'string');
+    assert(addr == null || addr instanceof Address);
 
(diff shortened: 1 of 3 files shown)


lib/wallet/wallet.js
```diff
@@ -1629,12 +1629,14 @@ class Wallet extends EventEmitter {
    * @param {String} name
    * @param {Number|String} acct
    * @param {MTX?} mtx
+   * @param {Address?} addr
    * @returns {Promise<MTX>}
    */
 
-  async makeOpen(name, acct, mtx) {
+  async makeOpen(name, acct, mtx, addr) {
```
Falci, line 1636: If `addr` is provided then `acct` has no use.

rithvikvibhu: Hm true. But `acct` is mandatory and also matches other `make*` methods signatures. Not sure what can be done

lib/wallet/wallet.js
```diff
@@ -3776,11 +3788,17 @@ class Wallet extends EventEmitter {
           break;
         case 'OPEN':
           assert(action.length === 1, 'Bad arguments for OPEN.');
-          await this.makeOpen(...action, acct, mtx);
+          {
```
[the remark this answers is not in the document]
rithvikvibhu: added braces

nodech:
So, previous code was generating new address for every bid, but because receiveAddress in makeBid does not increment the depth it was using the cleanup phase to repopulate them with different depth/addresses. Also for the BID it's critical for constructing nonce and now we can pass these addrs even before, so we can avoid this issue.

Do we really want to pass ADDR to the OPEN though ? I believe it does not use that address for anything and can easily be outside. Only BID has trouble with it.

rithvikvibhu:
Okay now only bids get new addresses.

PR #827, opened 2023-07-14
Update checkpoints guide in release-process.md

handshake-enthusiast:
https://github.com/handshake-org/hsd/pull/597#issuecomment-1635872519


nodech:
Thanks !

PR #828, opened 2023-07-14
icann lockup soft fork - Add 10k alexa to the soft-fork.

nodech:
The upcoming soft fork will encompass an update of the top 10,000 names from the Alexa list, which will be locked for a period of four years following the expiration of their claim. This provision gives us ample time to execute a proper hard fork midway. However, if this does not occur, the names won't be locked indefinitely.

In contrast, ICANN TLDs are permanently locked, though a hard fork could potentially unlock them or make them reclaimable.

To review the name list and data used in creating the lockup database, visit: https://github.com/nodech/hs-lockup-names-tmp/
It is a fork of https://github.com/handshake-org/hs-names

Thanks to @rithvikvibhu you can also check out the spreadsheets that shows the important lists: https://docs.google.com/spreadsheets/d/1DG0kBpLbLB2zKlJW2F7ANkk8SDbADpUa3sQuEY62_Bw/edit?usp=sharing

The new list no longer designates 'custom' as special, nor does it have a unique 'top 100'. Instead, it only differentiates between ROOT and non-ROOT. Please note that these conditions may change. For instance, we may want to keep 'CUSTOM' permanently locked, or adjust the release schedule (currently four years for non-root).

Closes #823


nickorlabs:
This is FANTASTIC!!!  Thank you guys for doing so much work! I know we talked about this in Discord, but wanted to ask here also, should we discuss a expiration date, just for the off chance the hard fork can not happen as quickly as we'd like, for the Top 10K list?  I like NOT doing this as it creates some pressure to finish the hard fork in my opinion, however wanted to propose this also.

rithvikvibhu:
Currently, this soft fork reserves non-root (non ICANN TLDs) for 4 years (i.e. 8 years from mainnet launch) and the root ones stay locked up forever.

>Please note that these conditions may change. For instance, we may want to keep 'CUSTOM' permanently locked, or adjust the release schedule (currently four years for non-root).

test/chain-icann-lockup-test.js
```diff
@@ -571,7 +595,7 @@ describe('BIP9 - ICANN lockup (integration)', function() {
     let wdb, wallet;
 
     const FROOT = ROOT.slice();
-    const FTOP100 = TOP100.slice();
+    const FTOP100 = TOP10k.slice();
```
handshake-enthusiast, line 598: Would it be better to rename occurrences of `TOP100`  to `TOP10`?  
```suggestion
    const FTOP10 = TOP10.slice();
```

At least it looks like `FTOP100` should be renamed.

lib/covenants/rules.js
```diff
@@ -416,12 +417,17 @@ rules.isLockedUp = function isLockedUp(nameHash, height, network) {
...
 
   if (!item)
     return false;
 
-  if (item.root || item.top100 || item.custom)
+  // ICANN Names are always LOCKED.
+  if (item.root)
+    return true;
+
+  // Alexa names will expire after 4 years.
```
handshake-enthusiast, line 429: Should it be 8 years?    
```suggestion
  // Alexa names will expire after 8 years.
```

At least _lib/protocol/networks.js_ has the following:  

```javascript
  /**
   * The time window which the
   * names will be locked before release or hard fork.
   * @const {Number}
   */

  alexaLockupPeriod: (8 * 365) * main.pow.blocksPerDay,
```

nickorlabs:
> Currently, this soft fork reserves non-root (non ICANN TLDs) for 4 years (i.e. 8 years from mainnet launch) and the root ones stay locked up forever.
> 
> > Please note that these conditions may change. For instance, we may want to keep 'CUSTOM' permanently locked, or adjust the release schedule (currently four years for non-root).

Sorry hah I should have paid better attention my bad!  Thank you for clarifying!

rithvikvibhu (approved):
Verified lockup data matches build from [hs-names-2023@3482d1](https://github.com/handshake-org/hs-names-2023/tree/3482d12e9c680030f1cec729f5e3a7aa454d0f15):
```
a7af1e93afa223b43f0d115814e001fc9312e58160c57e59662362c28aaefe3b  lockup.db
80fabbbebbb2609bbab2c020c95e57a96e284744da054498511aec517c31fb68  lockup.json
```

The same data is also updated in spreadsheet linked in PR desc.

PR #830, opened 2023-07-18
Add http handlers for createclaim/sendclaim

eskimo:
This adds http handlers for the createclaim/sendclaim rpc methods.

This was made for shakestation so I could select which wallet to create/send claims from without having to rpc selectwallet which could potentially break other scripts that may have been running at the same time.

### DIFF
--- a/lib/wallet/http.js
+++ b/lib/wallet/http.js
@@ -1439,6 +1439,38 @@ class HTTP extends Server {
 
       return res.json(200, json);
     });
+
+    // Create Claim
+    this.get('/wallet/:id/createclaim/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+      
+      enforce(name, 'Must pass name.');
+      enforce(rules.verifyName(name), 'Must pass valid name.');
+
+      const claim = await req.wallet.createClaim(name);
+      return res.json(200, {
+        name: claim.name,
+        target: claim.target,
+        value: claim.value,
+        size: claim.size,
+        fee: claim.fee,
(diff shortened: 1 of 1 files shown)


nodech:
I think it would be more appropriate if it followed convention for the name related endpoints.

`POST /wallet/:id/open` - used for creating and sending open
`POST /wallet/:id/bid` - used for creating and sending bid
etc..

Accepting post data: `name` for providing name and `broadcast`  - if we want to send right away.
Also test and hs-client wallet update would make this PR complete.

lib/wallet/http.js
```diff
@@ -1439,6 +1439,38 @@ class HTTP extends Server {
 
       return res.json(200, json);
     });
+
+    // Create Claim
+    this.get('/wallet/:id/createclaim/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
```
Falci, line 1447:
```suggestion
      const name = valid.str('name');

```

lib/wallet/http.js
```diff
@@ -1439,6 +1439,38 @@ class HTTP extends Server {
...
+        address: claim.address.toString(this.network),
+        txt: claim.txt
+      });
+    });
+
+    // Send Claim
+    this.post('/wallet/:id/sendclaim/:name', async (req, res) => {
+      const valid = Validator.fromRequest(req);
+      const name = valid.str('name');
+
```
Falci, line 1467:
```suggestion
      const name = valid.str('name');

```

Falci (changes requested):
Make lint happy

PR #835, opened 2023-08-04
pkg: update bweb.

nodech:
Updated bweb contains two PRs:
  - https://github.com/bcoin-org/bweb/pull/15
This can allow us to create API endpoints for broadcast which will not BROADCAST if the transaction creation took too long. Example can be seen here: https://github.com/namebasehq/hsd/pull/78

Another important change:
  - https://github.com/bcoin-org/bweb/pull/13
 Incorrect HTTP requests will no longer log/return 500, but 400 instead.


nodech:
We can add `abort on close` feature to the broadcast requests behind a flag set to `false`. Sometimes it's necessary for users to know 100% if the transaction was sent but control the limits using timeouts. Right now, when transaction send/batch send request times out, it will still broadcast them - because the function keeps running. Using `req.closed` and potential `abortOnClose` flag - we can have additional option to not broadcast if the http request closed.

Note: as mentioned in linked PR, this is not 100% guaranteed as the tx could already be being broadcast. Those will still be inconsistent, but the window gets narrower.

PR #836, opened 2023-08-05
Update network seeds, checkpoints and deps

nodech:
Updates in preparation for v6.x release.

- Update seeds from `seed.easyhandshake.com` and `https://hnsnodes.htools.work/api/v1/snapshots/latest/reachable`.
- Update checkpoints to 160k block which was ~6 months ago (17/02/2023).
- Update patch updates explicitly.


rithvikvibhu (approved):
- blockhashes for 130k and 160k match
- confirmed that the 20 unchanged and 23 new nodes are online

PR #838, opened 2023-08-08
Update hs-client generator

nodech:
Instead of trying to remove master history from `hs-client`, append to it version by version. It will give better visibility to what happened even though it is auto-generated.

### DIFF
--- a/scripts/gen-hsclient.js
+++ b/scripts/gen-hsclient.js
@@ -62,7 +62,9 @@ const COPY_FILES = [
   'bin/hsw-rpc',
   'lib/client/index.js',
   'lib/client/wallet.js',
-  'lib/client/node.js'
+  'lib/client/node.js',
+  'LICENSE',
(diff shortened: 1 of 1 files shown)


rithvikvibhu (approved):
Looks good! Replaced the origin and it worked nicely: https://github.com/rithvikvibhu/hs-client/commit/bd3600af79ceec2010ef88f6f57be7297f3301dc

One Q: does `fs.writeJSON` sort keys deterministically? I see some files in `package.json` move around a bit, is it random every time?

nodech:
writeJSON uses JSON.stringify inside. And it should depend on the way `hsClientPkg` is constructed. If we reorder the way they are added - it should change as well.

Quick check
```
for i in {1..10}; do ./scripts/gen-hsclient.js --no-git=true /tmp/hs-client-$i; sha256sum /tmp/hs-client-$i/package.json;  done
```

PR #853, opened 2023-09-28
Delete duplicates from seed nodes

handshake-enthusiast:
Duplicates were added in 76a0f75595f957252252b8d0bcc290182f5b985b. They are basically named nodes from the code above (in the file).


nodech:
Good catch, generally before releasing next version that's when we update this list. For next release we will also run the script and include dedupe there as well :)

PR #857, opened 2023-10-13
specify permanently hard-coded seed node locations

handshake-enthusiast:
Could be confirmed with https://ipinfo.io, https://hnsnodes.htools.work/api/v1/snapshots/latest/reachable or https://shakenodes.

I decided to look into the permanent seed node locations during my work on https://github.com/opensystm/handshake-micro-grants/issues/5. 8 are in US and 2 are in DE.

### DIFF
--- a/lib/net/seeds/main.js
+++ b/lib/net/seeds/main.js
@@ -1,25 +1,25 @@
 'use strict';
 
 module.exports = [
-  // pinheadmz
+  // Santa Clara, California, US (pinheadmz)
   'aonetsezqp4m52w4jpfq2gv3dggy2wqfwqtkfjyttgdidbvhgp5as@165.22.151.242',
-  // Falci
+  // Atlanta, Georgia, US (Falci)
   'ajpbuxkwtf7hkwbpw27siwg6dylsly4743rbhy2jssb3rxmircbrs@50.116.35.178',
-  // rithvikvibhu
+  // Ashburn, Virginia, US (rithvikvibhu)
   'aksygghkgmciomeldjf5sc6rs2sgn2m34zfdz4xr7z5vguqvjis4e@129.153.177.220',
-  // buffrr
+  // Chicago, Illinois, US (buffrr)
   'ai2wxnj4magzrhtlfvqnzuzg3vko25vryle5vd2bzi23jvcykbj5q@107.152.33.71',
-  // anon
+  // Mainz, Rheinland-Pfalz, DE (anon)
   'apt4rf2dfyelbivg63u47wykvdjtsl4kxzfdylkaae5s5ydldlnwu@159.69.46.23',
-  // chjj
+  // Fremont, California, US (chjj)
   'ajdzrpoxsusaw4ixq4ttibxxsuh5fkkduc5qszyboidif2z25i362@173.255.209.126',
-  // chjj
+  // Fremont, California, US (chjj)
   'akimcha5bck7s344dmge6k3agtxd2txi6x4qzg3mo26spvf5bjol2@74.207.247.120',
-  // chjj
+  // Morris Plains, New Jersey, US (chjj)
   'aoihqqagbhzz6wxg43itefqvmgda4uwtky362p22kbimcyg5fdp54@172.104.214.189',
-  // chjj
+  // Frankfurt am Main, Hesse, DE (chjj)
   'am2lsmbzzxncaptqjo22jay3mztfwl33bxhkp7icfx7kmi5rvjaic@139.162.183.168',
-  // chjj
+  // Morris Plains, New Jersey, US (chjj)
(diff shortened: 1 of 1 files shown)


rithvikvibhu:
I don't think adding locations guessed based on IP is a good idea, they are not reliable.
For example, just last month, @Nathanwoodburn set up his node and hnsnodes/shakenodes shows as UAE, but it's actually in AU.

handshake-enthusiast:
> I don't think adding locations guessed based on IP is a good idea, they are not reliable.

Indeed, those guessed locations may appear to be wrong. What do you think about specifying the locations in general? We could double check with authors or search in Github history/chats.

handshake-enthusiast:
I suppose at least countries are reliable in this case. Also I believe https://ipinfo.io/ is reliable enough. It's what I used as a source.  

Seems like https://shakenodes uses some unreliable source:  

```
shakenodes
IP: 194.50.5.28
Hostname: mail.woodburn.au
Org: FADEDSERVERS PTY LTD
Country: AE
Agent: /hsd:6.1.0/Nathan.Woodburn/

ipinfo.io
ip: "194.50.5.28",
hostname: "mail.woodburn.au",
city: "Albury",
region: "New South Wales",
country: "AU",
loc: "-36.0748,146.9240",
org: "AS140885 FADEDSERVERS PTY LTD",
postal: "2640",
timezone: "Australia/Sydney",
```

```
shakenodes
IP: 194.50.5.27
Hostname: 194.50.5.27
Org: FADEDSERVERS PTY LTD
Country: AE
Agent: /hsd:6.1.0/Nathan.Woodburn/

ipinfo.io
ip: "194.50.5.27",
hostname: "mail.woodburn.au",
city: "Albury",
region: "New South Wales",
country: "AU",
loc: "-36.0748,146.9240",
org: "AS140885 FADEDSERVERS PTY LTD",
postal: "2640",
timezone: "Australia/Sydney",
```

```
shakenodes
IP: 194.50.5.26
Hostname: 194.50.5.26
Org: FADEDSERVERS PTY LTD
Country: AE
Agent: /hsd:6.1.0/Nathan.Woodburn/

ipinfo.io
ip: "194.50.5.26",
hostname: "mail.woodburn.au",
city: "Albury",
region: "New South Wales",
country: "AU",
loc: "-36.0748,146.9240",
org: "AS140885 FADEDSERVERS PTY LTD",
postal: "2640",
timezone: "Australia/Sydney",
```

Falci:
The location added in the comments are not reliable, not verifiable, and IMO they add no value.

handshake-enthusiast:
> The location added in the comments are not reliable, not verifiable, and IMO they add no value.

Thanks for the feedback.

>not reliable

Can you confirm this entry is incorrect?  
```
   // Atlanta, Georgia, US (Falci)
   'ajpbuxkwtf7hkwbpw27siwg6dylsly4743rbhy2jssb3rxmircbrs@50.116.35.178',
```  

If it's incorrect I'll forward this to https://ipinfo.io. On their main page they state "Powering organizations around the world with reliable IP data".

>not verifiable

If you can reply to the question above doesn't this mean the location is verifiable?  

>no value

May be true if others says so as well (preferably providing at least some reasoning). My idea is to explicitly show how well different regions are represented. This is also covered in the grant proposal.

rithvikvibhu:
> >not verifiable
> 
> If you can reply to the question above doesn't this mean the location is verifiable?  

I don't think so. Verifiable would mean that everyone can verify that the server is indeed in that location. Falci (or anyone) claiming anything isn't verification.

> 
> >no value
> 
> May be true if others says so as well (preferably providing at least some reasoning). My idea is to explicitly show how well different regions are represented. This is also covered in the grant proposal.

More of a personal opinion: this kind of data is better visualized on a map with different stats on a separate website (like hnsnodes, etc.)

handshake-enthusiast:
>I don't think so. Verifiable would mean that everyone can verify that the server is indeed in that location. Falci (or anyone) claiming anything isn't verification.

By this definition the current comments aren't verifiable as well I suppose.

>More of a personal opinion: this kind of data is better visualized on a map with different stats on a separate website (like hnsnodes, etc.)

Definitely, it's not a substitution of a better visualised map like [hnsnodes](https://ipinfo.io/tools/summarize-ips/fe2db903-f8f8-4cb3-b71e-88a72b6157d9) or https://bitnodes.io.

rithvikvibhu:
> By this definition the current comments aren't verifiable as well I suppose.

Physical location cannot be verified, but ownership can be by signing a message since there's a brontide key [1].
But true, the current info - the names - isn't verified. I guess it's there to help reaching out to them if a node is offline / needs to update.

[1] https://gist.github.com/pinheadmz/680809e633467c56a3b3de33855bee38

nodech:
I agree that locations don't give any useful information, also if you are interested you can easily look them up using service you want.

maintainers of the nodes is useful information for pinging them when they go down. Otherwise it would not be part of the "set", but instead listed below where tool collected nodes are. There's no big difference between those two sets other than that.

PR #858, opened 2023-10-13
add a new permanently hard-coded seed node

handshake-enthusiast:
16 GiB and 4 vCPU using the latest generation general purpose Amazon EC2 M7g instance in Mumbai. [Status Page](https://empty-bonus-92feac03.cronitorstatus.com) and Live Status Badge:   Handshake seed node Mumbai

See https://github.com/opensystm/handshake-micro-grants/issues/5 for more details.

### DIFF
--- a/lib/net/seeds/main.js
+++ b/lib/net/seeds/main.js
@@ -21,6 +21,8 @@ module.exports = [
   'am2lsmbzzxncaptqjo22jay3mztfwl33bxhkp7icfx7kmi5rvjaic@139.162.183.168',
   // chjj
   'ap5vuwabzwyz6akhesanada4skhetd2jsvpkwuqxzuaoovn5ez4xg@45.79.134.225',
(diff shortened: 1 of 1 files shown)


handshake-enthusiast:
I haven't researched how each client chooses 8 nodes to connect to, curious whether seed node IP addresses order specified in `lib/net/seeds/main.js` matters.

nodech:
Instead of location it would be awesome to have contact information like others, it will be hard to have it in the first part of the code, if it goes down for some time.

Also it seems that it is down rn, any changes on this ?

handshake-enthusiast:
>Instead of location it would be awesome to have contact information like others, it will be hard to have it in the first part of the code, if it goes down for some time.

Changed.

>Also it seems that it is down rn, any changes on this ?

Yes, I had to setup it under another account. Should be stable now.

nodech (approved):

PR #859, opened 2023-10-16
Wallet: Add open/close and connect/disconnect events.

nodech:
WalletDB `open` does not guarantee that the connection will open, if it's in standalone mode it may fail. So having the connect method inside `open` w/o any guarantees leaves any logic depending on it break. So it's much better to totally remove the concept of connection from open.

This changes several things:
  - on open state cache needs to be recovered first (it was done on connect). If we don't have cache, then we initialize after on `connect` in the syncNode.
  - migration can no longer start a rescan (and it may have failed previously on standalone) - it will instead rollback. 
  - Added new flag `wallet-migration-no-rescan` - that tells to ignore rescan request from the migration.

Restructure the life-cycle methods for the walletdb.

 - open - no longer calls connect - now emits `open` event.
 - connect - needs to be called separately - now emits `connect` event.
 - disconnect - now emits `disconnect` event
 - close - will try to close client if it's open - now emits `close` event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ you run it for the first time.**
 ### Wallet Changes:
 
 - Add migration that recalculates txdb balances to fix any inconsistencies.
+- WalletDB Now emits events for: `open`, `close`, `connect`, `disconnect`.
 - HTTP Changes:
   - All transaction creating endpoints now accept `hardFee` for specifying the
     exact fee.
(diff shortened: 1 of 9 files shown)


rithvikvibhu (approved):
Tested Bob with wallet in plugin and remote modes. Didn't have to change anything since `wdb.connect` are called by hsd in both `open`s

Issue #603, opened 2021-05-21
Clean up unused network parameter: revocationDelay


pinheadmz:
This network names parameter has been replaced since mainnet launch, I do not think it is used anywhere in the code and can probably be removed for developer sanity and confusion's sake:

https://github.com/handshake-org/hsd/blob/c861e4fc95f0d023a7da4c90b93dab692b9209d0/lib/protocol/networks.js#L335-L341

pinheadmz:
Another thing we don't use any more is the entire hard-coded ICANN root zone. These files:

```
lib/dns/icann.js
lib/dns/icann-browser.js
lib/dns/tld.db
lib/dns/tld.json
```

I suspect hsd used to be built like hnsd [with a hard-coded ICANN root zone](https://github.com/handshake-org/hnsd/issues/34) but before mainnet launch, the hsd root server switched to actually resolving root zone names using a bns `RootResolver` with hard-coded hints instead.

PR #860, opened 2023-10-16
pkg: clean up unused params.

nodech:
Closes https://github.com/handshake-org/hsd/issues/603


rithvikvibhu (approved):
confirmed none of these are used anywhere :+1:

PR #861, opened 2023-10-18
wallet: don't add coinbase txs to the pending list.

### LINKED ISSUE #791: TXDB should remove "unconfirmed" coinbase TXs
Currently, they sit in pending but if a block is removed its cb tx is gone forever, the wallet should completely remove it

nodech:
Closes #791

### DIFF
--- a/test/wallet-test.js
+++ b/test/wallet-test.js
@@ -382,6 +382,51 @@ describe('Wallet', function() {
     }
   });
 
+  it('should remove unconfirmed coinbase tx', async () => {
+    const wallet = await wdb.create();
+
+    const block = nextBlock(wdb);
+    const normalTX = new MTX();
+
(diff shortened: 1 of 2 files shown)


test/wallet-test.js
```diff
@@ -382,6 +382,51 @@ describe('Wallet', function() {
...
+    cbTX.addInput({
+      prevout: new Outpoint()
+    });
+
+    cbTX.addOutput(await wallet.receiveAddress(), 5000);
+
+    const pendingBefore = await wallet.getPending();
+    assert.strictEqual(pendingBefore.length, 0);
+
+    assert(cbTX.isCoinbase());
```
rithvikvibhu, line 407: nit: can move above pendingBefore I think

rithvikvibhu (approved):

Issue #622, opened 2021-07-29
walletdb: ChainState.startHeight and `scan()`


pinheadmz:
`startHeight` is a strange property of the chainstate in walletDB and we should check how it is used. 

It can affect users when they rescan, because it becomes the default scan-from height:

https://github.com/handshake-org/hsd/blob/98a6491cdbfb173b4834a892b9bd55b6839cadbf/lib/wallet/walletdb.js#L512-L522

The misbehavior of this can be reproduced:

1. start hsd with a fresh data directory in regtest
2. generate 100 blocks
3. stop hsd, delete regtest/wallet directory & restart hsd (this will create a new walletDB "at start height 100")
4. `hsw-cli rescan` does nothing
5. `hsw-cli rescan 0` actually rescans (from genesis block)

I think this behavior may have at one point made sense (why bother rescanning from blocks before the wallet was created?) but anyone who's ever imported a wallet from seed phrase knows you need to rescan the entire chain, not just since the database was initialized. 

Another part of this mystery is the purpose of `markState()`:

https://github.com/handshake-org/hsd/blob/98a6491cdbfb173b4834a892b9bd55b6839cadbf/lib/wallet/walletdb.js#L1870-L1880

nodech:
chain startHeight, startHash both mark the point at which the first known transaction to the walletdb occured. This is not specific to a wallet, but instead earliest for all wallets. This allows easy to rescan height when something some transaction were missed somewhere in history.

`marked` just tracks whether we have ever received a tx, that is checked if we want to move forward the startHeight.

On reorganization if we move past the startHeight in the history, it wont set to 0/ZERO_HASH, instead it will get unmarked and height/hash will follow. This is okay as well with the design goal for the startHeight/startHash, it will mark whenever new transaction is found, will get marked and startHash/startHeight will get frozen at that point.

Better name for this probably would be, firstTransactionBlockHeight and firstTransactionBlockHash. After pagination port it should also possible to use height based transaction indexes to recover this data by checking minimum heights of all wallets, but becomes costlier for walletDBs with a lot of wallets.

This is not helpful for new wallets, because it's unknown where did that wallet receive tx. Is it fresh or imported ?

If @rithvik has some data what's the most used we could drop rescan default to be 0 instead of startHeight/startHash. Instead this will only serve as the information. New rescan will recover to the proper height. Adding per wallet information for this using pagination indexes is also good. If you are not using backup and want to export single wallet, this information is not useful.

[...]

PR #870, opened 2023-11-02
Wallet: add tests for the chain state and fix markState

nodech:
This is part of review for this: https://github.com/handshake-org/hsd/issues/622

Only adds test cases to the codebase related to chain state. Does not fix/change anything.

Covers discovered bug which is currently skipped. There's bug related to corruption but it's currently skipped as it's not solved.

More discussion on the issue.

### DIFF
--- a/lib/wallet/walletdb.js
+++ b/lib/wallet/walletdb.js
@@ -2142,7 +2142,7 @@ class WalletDB extends EventEmitter {
   /**
    * Get a wallet block meta.
    * @param {Hash} hash
-   * @returns {Promise}
+   * @returns {Promise<BlockMeta?>}
    */
 
   async getBlock(height) {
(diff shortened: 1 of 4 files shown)


nodech:
Latest commit as mentioned fixes two issues:
  - On close/crash of walletdb on add block wdb state would have
    incorrect state for the startHeight/startHash and mark. WalletDB
    could end up having incorrect startHeight/startHash for the rest
    of the walletDB until rescan before height. Check the
    wallet-chainstate-test case for corruption.
  - On the specific height where the walletDB received first
    transaction, state.height would also be incorrect (only for one
    block).

rithvikvibhu (approved):
Love the test! Feels like it repeats a lot, but very easy to read

Issue #831, opened 2023-08-01
Fix missing transaction with filter updates


nodech:
https://github.com/bcoin-org/bcoin/commit/c6859b347bd6989c1eb87fc58e05b6fcc4b91c8b
https://github.com/handshake-org/hsd/pull/519

Update the filter logic based on above two.

PR #883, opened 2023-12-27
Wallet Sync Updates

nodech:
### Node HTTP
  - Add WS call `get options` to get `chain` options: `prune` and `spv`. This
  allows wallet to detect capabilities of the client it connected to. E.g.
  `prune` can be used to check the maximum rescan depth available from the node.
  This is only called on wallet -> node connect.

### Deprecate `--no-wallet` for `spvnode`
  SPVNode will no longer accept `no-wallet` option. Currently, SPVNode
does not support standalone version. Filter synchronization to the pool
on SPVNode happens from wallet plugin nodeclient. HTTP Client currently
does not support that use case. In order to make that happen, node http
will have to add SPV Checks on Filter setting calls and even then it
won't be able to support more than 1 client.
  Using wallet plugin for `spvnode` is the only correct use.


nodech:
Note on #831. The PR and the commit tries to attempt synchronizing filter when running in SPV mode. Unfortunately, that's not going to work. The SPV add filter/set filter on `pool` does not have a way to get "acknowledgement" when ALL the peers have updated the filter, so you can't effectively wait for it from the SPV Node side.
  SPV Wallet initial sync wont get affected by it, as it will still miss the txs in the blocks that are streaming. So only constant reset is the only solution for SPV wallet for now. 
  

  Fullnode wallets can use rescan feature introduced in this PR to properly resolve the transactions, this does not change/fix add blocks on normal sync.

PR #885, opened 2024-01-16
wallet: set icannlockup to true for wallets.

nodech:
After the ICANNLOCKUP has been locked in, walletdb can safely enable this feature and make sure it does not accidentally try to OPEN locked up names.

Wallets using `v6` of the `hsd`, can now also enable `icannlockup` for the wallet.


rithvikvibhu (approved):
Bob 2.1.0 has this set to true and it's working as expected.

PR #888, opened 2024-03-04
Wallet TX Count and time indexing

nodech:
This is a port of the https://github.com/bcoin-org/bcoin/pull/605. It includes fixes mentioned in the PR.

  > This resolves CPU and memory exhaustion issues when requesting transaction history for a wallet with many transactions. Transactions are now also able to be queried by time based on median-time-past (MTP) based on chain data.

Backport:
  - https://github.com/bcoin-org/bcoin/pull/605

Issues:
  - bcoin: https://github.com/bcoin-org/bcoin/issues/559
  - hsd reference issue: #867

Related:
 - Closes #659 - zap will now use time and count indexing for unconfirmed txs.
 - Closes #867

Tasks:
 - [x] WDB Migration from v3 to v4.

### Minor changes
  - Adds `get median time` to the node http and hs-client node.
  - Adds `get entries` to the node http and hs-client node.

## Wallet changes
  - Add `getMedianTime` - to get median time past for the height, where height is already stored in the db.
  - Add `getMedianTimeTip` - to get median time past for the yet to be committed block.
  - Add block time by height cache.
  - Wallet now has additional methods for quering history:
    - `listUnconfirmed(acc, { limit, reverse })` - Get first or last `limit` unconfirmed transactions.
    - `listUnconfirmedAfter(acc, { hash, limit, reverse })` - Get first or last `limit` unconfirmed transactions after/before tx with hash: `hash`.
    - `listUnconfirmedFrom(acc, { hash, limit, reverse })` - Get first or last `limit` unconfirmed transactions after/before tx with hash `hash`, inclusive.
    - `listUnconfirmedByTime(acc, { time, limit, reverse })` - Get first or last `limit` unconfirmed transactions after/before `time`, inclusive.
    - `listHistory(acc, { limit, reverse })` - Get first or last `limit` unconfirmed/confirmed transactions.
    - `listHistoryAfter(acc, { hash, limit, reverse })` - Get first or last `limit` unconfirmed/confirmed transactions after/before tx with hash `hash`.
    - `listHistoryFrom(acc, { hash, limit, reverse })` - Get first or last `limit` confirmed/unconfirmed transactions after/before tx with hash `hash`, inclusive.
    - `listUnconfirmedByTime(acc, { time, limit, reverse })` - Get first or last `limit` confirmed/unconfirmed transactions after/before `time`, inclusive.
    - NOTE: Default is ascending order, from the oldest.

 Median time past is used by TX Pagination for time indexes. See details below.

### Wallet HTTP
[...]

PR #889, opened 2024-03-18
WalletDB Migration to v3

nodech:
Migration to WalletDB v3 is different from previous migrations. It does not try to reindex the old data into new layout, instead it drops most of it - allowing us to do heavier modifications. This migration removes all txdb mappings (except for blinds) and cleans up block entries received from the chain. This should trigger rescan from the genesis block.

  Migration is separated from the actual txdb/walletdb changes in order to allow multiple PRs to modify the database and w/o it getting in the way. Migration needs to get merged last, before release of v7. This allows us to invalidate all commits between wdb v3 modifications and the migration. After migration there should no longer be any wdb modifications for wdb v3.

  This PR currently tracks modifications:
   - #888

### DIFF
--- a/lib/wallet/migrations.js
+++ b/lib/wallet/migrations.js
@@ -74,7 +74,7 @@ class MigrateMigrations extends AbstractMigration {
 
 /**
  * Run change address migration.
- * Applies to WalletDB v0
+ * Applies to WalletDB v1
  */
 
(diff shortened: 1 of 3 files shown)


nodech:
Closing this. Proper migration is implemented in #888.

PR #890, opened 2024-04-21
Fix typo in Docker example

sethforprivacy:
This typo breaks the Docker command, as `hsd` is expecting an `=`.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -113,7 +113,7 @@ $ docker run --name hsd \
     --publish 12037:12037 \
     --volume $HOME/.hsd:/root/.hsd \
     hsd:$VERSION-$COMMIT \
-    --http-host 0.0.0.0 \
+    --http-host=0.0.0.0 \
     --api-key=foo
 ```
 


Falci:
I have a server running it with `--http-host '::'` and it works fine.

sethforprivacy:
Running the image [here](https://github.com/handshake-org/hsd/pkgs/container/hsd/145861724?tag=latest), `hsd` crashed on startup claiming that the http-host arg didn't exist due to the space instead of =.

As soon as I made it match [every other config value is supposed to be passed](https://hsd-dev.org/guides/config.html) in it started right up without issues.

Falci:
Tested again with the following code:
``` 
docker run --name hsd \
    --publish 12037:12037 \
    --volume /tmp/hsd:/root/.hsd \
    ghcr.io/handshake-org/hsd:latest \
    --http-host 0.0.0.0 \
    --api-key=foo
```
And it worked with or without the `=`.

Also, checked [bcfg](https://github.com/bcoin-org/bcfg/blob/2f49d5a84dbed5653bfe7fe90da32273f3a53056/lib/config.js#L976) and it seems to support both space or `=`.

There's no harm in merging this PR, but I don't see how this could be root of the problem you faced, unless there's a bigger issue with `bcfg`.

sethforprivacy:
Ah, perhaps it's a difference in how Docker Compose's `command` arg parses these?

Will share the format I used shortly that had issues.

rithvikvibhu:
I think similar to Dockerfile CMD, the `command`, an array, should be split and separated by spaces: https://docs.docker.com/reference/dockerfile/#cmd

So if you want to use spaces in the compose file, then it should look like
```yml
commands:
    - "--http-host"
    - "0.0.0.0"
    - "--api-key"
    - "foo"
```

Or of course, like the previous comment, using `=` in a single line will work since it's still 1 argument passed to hsd/bcfg.

This behaviour applies to any executable/program receiving arguments, not just hsd/bcfg.

Since the current example in INSTALL.md is with docker CLI (and not Compose), I don't think it's wrong; space is perfectly acceptable.

However, adding a new Compose example below this one would be helpful. It can use `=` which brings the keys and values on the same line making it easier to read.

sethforprivacy:
Good catch, @rithvikvibhu, has been so long since I used `docker run` instead of Compose I had forgotten that difference.

Will open a PR adding an example instead!

sethforprivacy:
Related PR: https://github.com/handshake-org/hsd/pull/891

PR #895, opened 2024-05-31
test: change ownership resolver from google to cloudflare.

nodech:
google DNS resolvers have limits on the size of the queries. cloudflare.com TXT sometimes is over that limit. Switch to cloudflare servers for now and make it configurable for tests.


Anunayj (approved):
I even tried flushing Google DNS Cache from [here](https://developers.google.com/speed/public-dns/cache). Did not seem to work, query size is not a issue either (tested with large TXT records on my domain), probably some internal issue at Google DNS, likely better to switch to Cloudflare.

PR #900, opened 2024-07-03
Update migrations

nodech:
Update Migrations
=================

Old migrations depend on the methods and states that may be broken by the future updates to the codebase. This PR attempts to fix old migrations to reduce depenencies to the codebase. For example, Change address migration is currently broken, if user were to migrate from earlier non-fixed versions (e.g. v2.1.0)

Migrations are now run against generated data from the original versions of hsd. Generator and generated codes can be found in `test/data/migrations`, for both the wallet and the chain.

This PR attempts to go over all migrations. Review and fix test cases and migrations:

  - WDB
    - [x] MigrateMigrations
    - [x] MigrateChangeAddress
      - This can be further detached from the actual code, but currently
      still depends on wallet/walletkey, wallet/path and
      wallet/records.MapRecord.
    - [x] MigrateAccountLookahead
    - [x] MigrateTXDBBalances
  - ChainDB
    - [x] MigrateMigrations
    - [x] MigrateChainState
    - [x] MigrateBlockStore
    - [x] MigrateTreeState


nodech:
Tested migration from v2.1.3 to this branch for `spv`, `full` and `pruned`. All migrated w/o issues.

PR #901, opened 2024-08-26
MTX - Resolve coins from coinview as well during coinselection.

nodech:
This is another simpler alternative to the: https://github.com/handshake-org/hsd/pull/666

It still gets rid of the inputs for the coinselection, but will make sure to check if coins are in coinview first and Coins list second.  This also fixes another issue, where existing inputs in the tx could get in conflict with passed inputs in the options.

Closes: #666
Closes: #639

### DIFF
--- a/lib/primitives/mtx.js
+++ b/lib/primitives/mtx.js
@@ -1409,12 +1409,13 @@ class CoinSelector {
   /**
    * Create a coin selector.
    * @constructor
-   * @param {TX} tx
+   * @param {MTX} tx
    * @param {Object?} options
    */
 
   constructor(tx, options) {
     this.tx = tx.clone();
+    this.view = tx.view;
     this.coins = [];
     this.outputValue = 0;
     this.index = 0;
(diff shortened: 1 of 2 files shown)


Anunayj:
This is so much cleaner and I think it's probably ideal to keep it as simple as possible. There's just a few things to keep in mind as of right now:
1. fund() will steal clear witness items, and those will need to be added back.
2. It will also clear `nSequence` for all inputs! I am unsure how big of a deal this is, but please beware when using setLocktime or using Relative timelocks before calling fund!

Anunayj (approved):
This implementation is significantly cleaner than the previous, but we would still have be cautious about potential interactions between nLocktime, nSequence, and fund() in the future. Specifically, NEVER call fund after setting nLocktime or nSequence.

- nLocktime is likely to be less of an issue, as it will be active as long as one of the inputs is < 0xFFFFFFFF, and It throws a a exception when called with no inputs. However there can still be issues when the other inputs are ANYONECANPAY and Locktime being active was extremely important.

- nSequence being reset could cause a more significant problem if someone is unaware of this and was expecting Relative Locktimes to be active.

We might want to consider a future revision to address these concerns, as they aren't necessarily within the scope of this PR. There also needs to be a discussion on what behavior should be expected from addCoin() when nLocktime is set.

PR #927, opened 2025-04-19
chain: add airstop soft fork

rithvikvibhu:


AndreiChyrych:
«This is a very simple PR to disable airdrops with a soft fork»

is it possible to reintroduce airdrop with a new soft fork?

Falci:
> is it possible to reintroduce airdrop with a new soft fork?

No, that would require a hard fork.

j9011303:
I support this proposal. Ending the airdrop makes sense at this stage of the project.
It can help reduce unnecessary token supply and move the project forward.

smcki012:
My Father allows this. You can proceed.

agaamin:
Lets do this. 
Burn the damn spare tokens.

smcki012:
No idiots in the handshake kitchen. Just do it already you weirdos.

HNSbroker:
Isn't there a way to actually burn the coins and permanently remove them from the supply, instead of just stopping the claims?

Falci:
> Isn't there a way to actually burn the coins and permanently remove them from the supply, instead of just stopping the claims?

These coins don't actually exist. The airdrop creates coins out of thin air when they're claimed. By stopping the airdrop, we prevent these coins from ever being created, removing them from the supply.

faltrum:
Great Moment and great decision!. Congrats community involved in this step. Next steps:

1. Ensure miners awareness 
2. Update the BIP9 dates
3. Merge
4. Wait the merge to be activated, or not.

Thanks @rithvikvibhu and @Falci for clarifications.

lib/blockchain/chain.js
```diff
@@ -661,6 +661,16 @@ class Chain extends AsyncEmitter {
             100);
         }
 
+        // Disable airdrop claims if
+        // - airstop is activated, and
+        // - this is an airdrop, not a faucet claim
+        if (state.hasAirstop && !proof.isAddress()) {
```
Falci, line 667: Honestly, the branch name is more descriptive than the version bit's name. 
But I understand this has a bit of marketing/promotion/momentum attached to it as is.

rhymincymon:
I think this is the right way forward.

The supply looks as follows [without](https://shakeshift.com/stats) and [with](https://shakeshift.com/stats?airstop=1) the soft fork activated.

airdrop airstop


After the last subsidized block is mined, a total of 961,262,671.16 HNS would have been issued.

lib/blockchain/chain.js
rhymincymon: There shouldn't be an exception for faucet claims, imo.

The faucet has been emptied anyway, so it doesn't make a difference. But it would be clear that no further coins are going to be issued.

Falci: IMO, the PR should be small and precise. The goal was to remove airdrop claims and this is all it should do.

rhymincymon: There are recipients who received their airdrop via the faucet (their keys have been removed from the tree).

Plus I don't really see someone wanting to start a new independent chain with the faucet being distributed to the same recipients all over again but airdrops being disabled.

I think would make more sense for this fork to disable everything.

rithvikvibhu: You're right, I had forgotten about the bit where early airdrop recipients were moved to faucet and out of tree.

I would've liked to separate out and include only faucet claims from the airdrop and leave the sponsor/creator faucet as-is, but it doesn't look like they are marked anywhere. So included all faucet claims.

lib/protocol/networks.js
```diff
@@ -480,6 +480,16 @@ main.deployments = {
     required: false,
     force: false
   },
+  airstop: {
+    name: 'airstop',
+    bit: 2,
+    startTime: 1751328000, // July 1st, 2025
+    timeout: 1759276800, // October 1st, 2025
```
nodech, line 487: Given how the windows work, October 1st will be in the middle of the window. This will actually make that last window non-existent. This will essentially make 2025 Sep 23 the last day of voting.

Windows ends:
```
1 - Tue, 15 Jul 2025 00:00:00 GMT
2 - Tue, 29 Jul 2025 00:00:00 GMT
3 - Tue, 12 Aug 2025 00:00:00 GMT
4 - Tue, 26 Aug 2025 00:00:00 GMT
5 - Tue, 09 Sep 2025 00:00:00 GMT
6 - Tue, 23 Sep 2025 00:00:00 GMT
7 - Tue, 07 Oct 2025 00:00:00 GMT
8 - Tue, 21 Oct 2025 00:00:00 GMT
```

If we want to include October 1st, maybe do October 7 + 10 minutes (make it a day Oct 8).

 bip9 stats will prob incorrectly say that auction is still in `started` state in the last auction, but timeout check just skips the last period.

rithvikvibhu: makes sense, updated to 8th Oct 2025

nodech:
> After the soft fork activates, miners may still accidentally use airdrops broadcasted in their blocks. This could potentially halt the network as those blocks are no longer valid. I believe miner code AND mempool code needs addressing.
> 
> Mempool on activation could drop all airdrops that are logged in it, to ensure miners don't include them in the blocks. And mempool also needs to not accept/propagate airdrops. Also in addition relevant miner code could be modified so it does not include it either way. I believe both changes are necessary.

Okay, this is done.

lib/mempool/mempool.js
```diff
@@ -274,6 +277,7 @@ class Mempool extends EventEmitter {
     // for a now expired name. Another
     // example is a stale BID for a name
     // which has now reached the REVEAL state.
+    const prevState = await this.nextState;
```
pinheadmz, line 280: Is this.nextState a promise?

lib/mempool/mempool.js
```diff
@@ -1326,6 +1321,9 @@ class Mempool extends EventEmitter {
     if (!proof.isSane())
       throw new VerifyError(proof, 'invalid', 'bad-airdrop-proof', 100);
 
+    if (this.nextState.hasAirstop)
+      throw new VerifyError(proof, 'invalid', 'bad-airdrop-disabled', 0);
```
pinheadmz, line 1325: I think you can actually enforce this policy rule by default as part of the release. This is a bit different than Bitcoin soft forks that add a feature but I think you can see in Taproot that mempool policy was implemented without any activation:

https://github.com/bitcoin/bitcoin/commit/e9a021d7e6a454d610a45cb9b3995f0d96a5fbb6

This at least rules out any race condition or off-by-one issue with the miners. Anyone who really wants to get a claim in still can - it just be hard to relay as the next release is deployed.

lib/mempool/mempool.js
nodech: No it's not. Nice catch.

lib/mempool/mempool.js
nodech: After soft-fork activates, I believe this one needs to be in place. It's will no longer be an option to have bad mempool (Can't create a block if there are airdrops in it)

Policy on top could definitely help (Some could preemptively disallow those), but enabling by default would mean that you will have to contact miners to get airdrops in?

I am not 100% sure that's what you meant ?

If I understood btc commit properly, that PR adds new checks for newly introduced thing, does not disallow old thing ?

test/chain-airstop-test.js
```diff
@@ -186,69 +202,119 @@ describe('BIP-9 - Airstop (integration)', function () {
...
-        tryClaimingAirdropProofs(node, [faucetproof]),
-        {
-          code: 'invalid',
-          reason: 'bad-airdrop-disabled'
-        }
-      );
+      let err;
+
+      try {
+        await tryClaimingAirdropProofs(nodeCtx, [airdropProof, faucetProof]);
```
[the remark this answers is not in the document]
nodech: Nope, it's meant to be how it was, testing them separately.

PR #937, opened 2025-07-03
feat: add GitHub Actions workflow to tweet on release

Falci:


Nathanwoodburn:
Looks good. Do you have a link to the test tweets?

Also is there a better message to have instead of below?
```
🚀 New release! hsd v8.0.0 is out now.
Check it out: https://github.com/handshake-org/hsd/releases/tag/v8.0.0
```

It looks good just wondering if we can pull some of the major changes from the release notes. Although that might be more effort than it is worth

PR #945, opened 2025-09-05
feat: Add chaindb option for compaction directory

udogri:
This commit introduces a new option, , to the ChainDB configuration. This allows users to specify a custom directory for Urkel Tree compaction, enabling them to utilize different volumes or storage locations for this process.

Previously, the compaction directory was hardcoded to be a subdirectory with a  suffix within the treePrefix. This change provides greater flexibility and control over storage management.

The following changes were made:

- Modified the ChainDB constructor to check for the  option.
- If provided, the  option is set to the value of .
- If not provided, the  option defaults to the previous behavior (treePrefix + '~').

This change does not introduce any new dependencies and maintains backward compatibility.

Testing:

Due to limitations in the testing environment, full integration tests could not be performed. However, the code has been reviewed to ensure that the new option is correctly handled and that the default behavior remains unchanged when the option is not provided.


lib/blockchain/chaindb.js
```diff
@@ -63,6 +63,15 @@ class ChainDB {
...
     this.blocks = options.blocks;
 
+    if (this.options.compactionDirectory == null) {
+      this.options.compactionDirectory = this.options.treePrefix + '~';
+    }
+
+    // Allow users to specify a custom compaction directory
+    if (options.customCompactionDirectory != null) {
+      this.options.compactionDirectory = options.customCompactionDirectory;
+    }
```
rithvikvibhu, line 73: What's the difference between `compactionDirectory` and `customCompactionDirectory`?

udogri: `customCompactionDirectory` is the  user-provided option, while `compactionDirectory` is what the system  actually uses. If the user provides a custom directory, we use that; otherwise, it defaults to `treePrefix + '~'`.

nodech (changes requested):
this.options in the chaindb is passed from the chain.js. The type of it is ChainOptions. That's the one responsible for normalizing and parsing all the configurations. chaindb does not check options, it should get finalized in chain.js ChainOptions.

Chain also needs to somehow accept this options, that happens in fullnode.js and spvnode.js. Given spvnode does not have a tree, it only needs to be applied to fullnode.

You can check compactTreeOnInit option for example. (relevant to the tree compaction as well)

udogri:
Okay so compactTreeOnInit option will give me an idea of the changes that I need to make

Issue #757, opened 2022-07-02
Include `size_on_disk` on `getblockchaininfo`


Falci:

pinheadmz:
🥪

Falci:
> 🥪

Did you mean 🍉?

pinheadmz:
No:
https://discord.com/channels/822591034202521641/825772593180704768/992125032060944564
> Free cheese sandwich to whoever opens issue #100!

Falci:
Which directories/files should we consider? All them?

```
hsd_data % du -hsx 2>/dev/null *
9.2G    blocks
 11G    chain
 14M    debug.log
 20M    debug_2022-02-04_10-44-02-910.log
 20M    debug_2022-02-10_09-28-29-262.log
 20M    debug_2022-02-19_18-19-25-959.log
 20M    debug_2022-03-04_07-07-09-799.log
 20M    debug_2022-03-31_08-43-33-499.log
 20M    debug_2022-04-17_12-01-03-976.log
 20M    debug_2022-04-18_06-58-51-976.log
 20M    debug_2022-04-18_14-34-50-939.log
 20M    debug_2022-05-03_20-39-59-716.log
 20M    debug_2022-06-11_22-25-42-401.log
 96K    hosts.json
4.0K    key
224K    regtest
 27G    tree
 12M    wallet
4.0K    wallet.log
```

rithvikvibhu:
I think `size_on_disk` in `getblockchaininfo` should semantically be the same as in bitcoin, so `blocks` + `chain` (or `spvnode` when in spv mode).
All full nodes will have around the same total. Including tree makes it a bit dynamic, since it can be compacted at any time and as many times as needed.

We could have another object in `hsd-cli info` or `getblockchaininfo` (not sure why we have two, whichever is a better fit):
```js
sizes: {
  blockchain: spv ? size(spvchain) : (size(blocks)+size(chain)),
  tree: size(tree),
  total: size(.), // includes everything: wallet, logs, etc.
},
```

pinheadmz:
The `chain` folder also currently includes the addr and tx indexes. Those will get moved out to their own folder hopefully with indexer refactor in hsd v5 (to pull from bcoin)

Also now that we have `treeState` object in chainDB I think we can add a `gettreeinfo` which should cover the size of the Urkel tree on disk.

Nathanwoodburn:
I've just started on this with https://github.com/handshake-org/hsd/pull/946
Let me know if there is anything I've missed

PR #946, opened 2025-10-08
Add size_on_disk to getblockchaininfo

Nathanwoodburn:
This adds the size_on_disk value to the rpc call `getblockchaininfo` (see https://github.com/handshake-org/hsd/issues/757)

The output is like so

image


Nathanwoodburn:
Do we want the `sizes` left in or do we only want the `size_on_disk`

